PR #265, opened 2020-12-24
Allow users to configure tasks location

DerekYu177:
Right now, when you install a new task, it's located by default under `app/tasks` and `{test|spec}/tasks`. In legacy systems, this might not be a convenient location, and I think it would be worth it to expose this location for configurability, since the typical location would be `app/jobs/migration` instead. 

EDIT: Sorry, I meant `app/jobs/maintenance` instead of `app/jobs/migration`

### DIFF
--- a/lib/generators/maintenance_tasks/task_generator.rb
+++ b/lib/generators/maintenance_tasks/task_generator.rb
@@ -14,7 +14,7 @@ module MaintenanceTasks
     # Creates the Task file.
     def create_task_file
       template_file = File.join(
-        "app/tasks/#{tasks_module_file_path}",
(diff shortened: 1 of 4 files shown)


volmer:
I don't see a compelling reason to support a configurable path for Tasks. Just as any other generator, the `maintenance_tasks:task` generator will create Tasks under `app/tasks`, which is where Tasks are supposed to be found in a Rails app that uses our gem. It wouldn't make sense to place Tasks under `app/jobs` either, since Tasks and Jobs are different things.

rafaelfranca:
Ok. So let's close this PR. Thank you so much for the contribution.

PR #284, opened 2021-01-06
Push to rubygems

etiennebarrie:
That should let shipit deploy to rubygems by default.


volmer (approved):

adrianna-chang-shopify (approved):
LGTM! I'll get the PR up to bump the version and create the GH release, and then we can 🚢 this!

PR #287, opened 2021-01-06
Revert "Merge pull request #216 from Shopify/deleted-tasks"

volmer:
This reverts pull request #216.

The conversation about deleted tasks is still ongoing. While we don't have a clear vision about discoverability and what to do about orphaned runs, or even if we care about this at all, we should not ship optimizations regarding them.

### DIFF
--- a/app/controllers/maintenance_tasks/runs_controller.rb
+++ b/app/controllers/maintenance_tasks/runs_controller.rb
@@ -6,7 +6,7 @@ module MaintenanceTasks
   #
   # @api private
   class RunsController < ApplicationController
-    before_action :set_run, except: :index
+    before_action :set_run, :set_task, except: :index
 
     # Shows a full list of Runs.
     def index
@@ -21,7 +21,7 @@ module MaintenanceTasks
     # Updates a Run status to paused.
     def pause
       @run.pausing!
-      redirect_to(task_path(@run.task_name))
+      redirect_to(task_path(@task))
     rescue ActiveRecord::RecordInvalid => error
       redirect_to(task_path(@run.task_name), alert: error.message)
     end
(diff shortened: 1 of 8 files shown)


etiennebarrie (changes requested):
It is not an optimization. A running "deleted" task may be trashing your production and this button allows to cancel it.

rafaelfranca:
Let me see if I understood the problem correctly.

If you have a task that was deployed, started being executed and later reverted, would not the execution of the task stop automatically because the job would fail to fetch the class because it doesn't exist anymore?

etiennebarrie:
Yes that's a good point, when the job is interrupted because of the deploy, if it's picked up by a new worker, the execution will stop since the code isn't there. BTW we have a bug about this, I'll open a PR.

But that's only part of the problem, a deploy may take a while, the interrupted job may be picked up by an old worker again during the rollout, or the web workers may be restarting before the job workers. Whatever the case, you can end up with a web server which thinks the task is deleted when it's not yet so this shortens the time to recovery in those situations.

Put another way, if you can see the cancel button for a deleted task, it may be an edge case, but it means we can easily cancel the task, and today we offer the choice, but this PR wants to remove it.

It is an edge case, but for me there was a discussion of compromise when shipping #216:
- yes it's an edge case, the value is relatively small overall
- but it can be hugely beneficial to be able to cancel a task when the need arise
- and it's a very small change, with nothing substantial in it, no changed assumptions, no consequences for other potential features, it's mostly changing the view to not disable a button

So it's this calculation that led me to ship that PR.

Even if a service disruption happens, and someone reverts, while someone else thinks maybe they can just cancel the task, and it turns out they can't because the button is disabled, they will never come here and ask us to change because they will usually blame their bug/process, it wasn't our fault really, and yet we could have easily helped them in their service disruption by letting them cancel an actually running task.

rafaelfranca:
Ok. I'm convinced that this has value, specially because of the rollout phase that, yes, can take a while in some deployment. Thank you for explaining.

PR #288, opened 2021-01-06
Revert "Merge pull request #274 from Shopify/runs-index"

volmer:
This reverts pull request #274. As of now there is no evidence of any benefit of allowing such search feature and it is simply polluting the UI. Runs can all be retrieved on their respective Task pages.


etiennebarrie (commented):
Search may be debatable but the list is important for auditing IMO.

Simply going to the different task pages is not enough because deleted tasks are not listed anywhere, to be able to go there you'd need to dive into git to find all the tasks that ever existed in your codebase.

The alternative is either going to the console or the database, which most likely is not possible for applications that most need auditing. It would be another discussion if something like basecamp/console1984 was public. I think we're all fine and comfortable reading through a gem's code and using the console, but I think we want to provide most users a black box that just works, and we already had feature requests about auditing so I think it's pretty clear the list of runs at least is needed.

Once you have the list, it felt to us like search was useful to avoid having to go through all the pages, but I'm willing to entertain the idea of removing the search only, going through all the pages almost sound reasonable compared to having to go through git to just get some basic information about task runs.

adrianna-chang-shopify:
I'm of the opinion that search is a must if we provide a list of runs for auditing purposes. If you have 1000+ tasks, clicking through pages and pages to try and find your task run is going to be painful.

rafaelfranca:
Why people want to see old runs of tasks that they chose to delete? What information is on those old run that are pertinent to people?

etiennebarrie:
> Why people want to see old runs of tasks that they chose to delete?

It's not only runs of tasks that were deleted, it's all the runs.

For example, today someone asked in the channel about passing shop ids to a Task, which we could do with CSV upload but was worried about auditability by comparison with a job being enqueued with shop ids in a migration. If we keep runs around, we will be able to keep the CSV files used as parameters (or any other way we come up with parameters for runs), and have a good story around auditing. I totally admit is a bit speculative as we don't have params or CSV at the moment, but it's basically the same argument we can have to keep the runs today, except the information we have today is pretty limited.

> What information is on those old run that are pertinent to people?

- at the simplest level, the fact a Task ran, which in itself can be useful (e.g. tracing down performance regressions to a running task)
- the time the run was enqueued at
- the number of iterations
- the time spent running (as opposed to paused or interrupted)
- the time it finished running
- the time it actually started running (but we don't show this to users, only when it was enqueued)

rafaelfranca:
Got it. But do we need to have auditing and view of those auditing in our library? Would not the logs be sufficient for that? It is for jobs since we don’t have a view to check all the jobs that have run.

adrianna-chang-shopify:
Okay, I think I'm convinced that this is not something our library needs to do. Re: logs, the argument before was that logs are not retained for long enough to be valuable in auditing Tasks. But it's true that it's a framework for running Tasks, not an auditing system, so I think we can forego this.

The ability to audit CSV tasks was also on our radar (as Étienne mentioned above, we heard hesitation from some users who were concerned about the audibility of these types of tasks), but I suppose there are good counter-arguments to this:
- A one-off CSV Task that cares about documenting which params the Task ran with should just do a regular Task so that the params are hard-coded into the collection and there is a paper trail of this Task
- As long as the Task is not deleted, these Runs will still be auditable. If a PR goes up to delete the Task, they can always download the set of CSVs for all Runs of the CSV Task if they care about keeping these around.

rafaelfranca:
I think today those users don't have audibility since they use regular migrations, right? So it is not like they are losing something.

adrianna-chang-shopify (approved):

PR #289, opened 2021-01-06
Revert "Merge pull request #248 from Shopify/bugsnag-integration"

volmer:
This reverts PR #248. I don't think we should make assumptions about the usage of Bugsnag. As long as we make integrating Bugsnag possible, and document it in the README, it should suffice.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,11 @@ The generator creates and runs a migration to add the necessary table to your
 database. It also mounts Maintenance Tasks in your `config/routes.rb`. By
 default the web UI can be accessed in the new `/maintenance_tasks` path.
 
+In case you use an exception reporting service (e.g. Bugsnag) you might want to
+define an error handler. See
+[Customizing the error handler](#customizing-the-error-handler) for more
+information.
+
 ## Usage
 
 ### Creating a Task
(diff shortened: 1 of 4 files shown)


rafaelfranca (approved):
Agree this doesn't fit this library. This is a Shopify default that can live in any integration library we have. It could be in our application generator, or in our cloud-support library. So better to remove from this library.

adrianna-chang-shopify (approved):
Yep, I'm okay to revert 👍

etiennebarrie (commented):
I think we should implement one of the solutions proposed by rafael for shopifolk before we ship this, or at the very least add some more documentation in the installation section. 

Otherwise, I'm willing to bet a good number of projects installing the gem will end up not being configured properly.

rafaelfranca:
Let's expand the documentation. But I don't think we need to implement the generator before merging this since we are not even including this gem yet in the generators. If an app configure this wrongly, given our documentation explain it well, it is the price to pay for being early adopters.

volmer:
Okay I expanded the README a bit more about error handlers. Let me know if you need any more info mentioned in there!

README.md
```diff
@@ -224,15 +224,15 @@ be placed in a `maintenance_tasks.rb` initializer.
...
-If your application uses Bugsnag to monitor errors, the gem will automatically
-notify Bugsnag of any errors raised while a Task is performing.
-
-If you want to integrate with another exception monitoring service or customize
-error handling, a callback can be defined:
+If you want to perform operations with errors raised during a Task execution,
+such as reporting them to an exception monitoring service like Bugsnag,
+you can define an error handler. Simply create a proc that takes an error as
+argument. Maintenance Tasks will then call the error handler proc whenever an
+error is raised.
```
etiennebarrie, line 231: IMO the original text from the revert was clearer, we may be able to improve on it but this is not.

```suggestion
If you want to integrate with an exception monitoring service (e.g. Bugsnag),
you can define an error handler:
```

etiennebarrie (commented):
It's still missing from the installation step.

adrianna-chang-shopify (approved):

README.md
```diff
@@ -40,6 +40,11 @@ The generator creates and runs a migration to add the necessary table to your
 database. It also mounts Maintenance Tasks in your `config/routes.rb`. By
 default the web UI can be accessed in the new `/maintenance_tasks` path.
 
+In case you use an exception reporting service (e.g. Bugsnag) you might want to
+define an error handler. See
+[Customizing the error handler](#customizing-the-error-handler) for more
+information.
```
etiennebarrie, line 46:
```suggestion
In case you use an exception reporting service (e.g. Bugsnag) you might want to
define an error handler. See [Customizing the error
handler](#customizing-the-error-handler) for more information.
```

volmer: I'd rather not break the link in the end.

PR #290, opened 2021-01-07
Return the Task Job as a string in the config

volmer:
Similar to what was done in #255 for `task_module`, this change returns a String for the job class name, which can be constantized where necessary.

### DIFF
--- a/test/lib/maintenance_tasks_test.rb
+++ b/test/lib/maintenance_tasks_test.rb
@@ -15,14 +15,6 @@ class MaintenanceTasksTest < ActiveSupport::TestCase
     MaintenanceTasks.tasks_module = previous_task_module
   end
 
-  test '.job can be set' do
-    original_job = MaintenanceTasks.job.name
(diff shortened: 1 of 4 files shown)


test/lib/maintenance_tasks_test.rb
```diff
@@ -2,27 +2,6 @@
...
-  end
-
-  test '.tasks_module can be set' do
-    previous_task_module = MaintenanceTasks.tasks_module
-
-    MaintenanceTasks.tasks_module = 'Task'
-    assert_equal('Task', MaintenanceTasks.tasks_module)
-  ensure
-    MaintenanceTasks.tasks_module = previous_task_module
-  end
```
adrianna-chang-shopify, line 16: 👍 to getting rid of these tests, but can we commit this separately?

adrianna-chang-shopify (approved):

etiennebarrie (approved):

PR #293, opened 2021-01-08
Handle error handling for deleted tasks

etiennebarrie:
If a Task is enqueued and is then deleted before being performed, or if a Task is interrupted and deleted, it results in a crash.

To repro:

```sh-session
$ bundle add sidekiq
$ $EDITOR test/dummy/config/application.rb
# add this line
# config.active_job.queue_adapter = :sidekiq
```

Start a server, visit http://localhost:3000/maintenance_tasks/tasks/Maintenance::UpdatePostsTask, start a run then:

```
$ rm test/dummy/app/tasks/maintenance/update_posts_task.rb
$ (cd test/dummy; bundle exec sidekiq)
# crash
```

This is fixed by the first commit.

```
$ git checkout task-deleted-during-run^
$ (cd test/dummy; bundle exec sidekiq)
# this should pick up the retry, otherwise clean up and redo the steps above
```

```sh-session
$ git restore -- test/dummy/app/tasks/maintenance/update_posts_task.rb
$ (cd test/dummy; bundle exec sidekiq)
# enqueue the job, once it starts, hit Ctrl-C to stop sidekiq and interrupt the task
$ rm test/dummy/app/tasks/maintenance/update_posts_task.rb
$ (cd test/dummy; bundle exec sidekiq)
# crash
```

This is fixed by the second commit:

```sh-session
$ git checkout task-deleted-during-run^
$ (cd test/dummy; bundle exec sidekiq)
# this should pick up the retry, otherwise clean up and redo the steps above
```

### DIFF
--- a/app/jobs/maintenance_tasks/task_job.rb
+++ b/app/jobs/maintenance_tasks/task_job.rb
@@ -90,7 +90,7 @@ module MaintenanceTasks
     end
 
     def on_error(error)
-      @ticker.persist
+      @ticker.persist if defined?(@ticker)
       @run.persist_error(error)
       MaintenanceTasks.error_handler.call(error)
     end
--- a/app/validators/maintenance_tasks/run_status_validator.rb
+++ b/app/validators/maintenance_tasks/run_status_validator.rb
@@ -57,7 +57,9 @@ module MaintenanceTasks
       #   it is interrupted.
       # interrupted -> cancelling occurs when the task is cancelled by the user
       #   while it is interrupted.
-      'interrupted' => ['running', 'pausing', 'cancelling'],
+      # interrupted -> errored occurs when the task is deleted while it is
+      #   interrupted.
+      'interrupted' => ['running', 'pausing', 'cancelling', 'errored'],
     }
 
     # Validate whether a transition from one Run status
--- a/app/validators/maintenance_tasks/run_status_validator.rb
+++ b/app/validators/maintenance_tasks/run_status_validator.rb
@@ -11,7 +11,8 @@ module MaintenanceTasks
       # enqueued -> pausing occurs when the task is paused before starting.
       # enqueued -> cancelling occurs when the task is cancelled
       #   before starting.
-      # enqueued -> errored occurs when the task job fails to be enqueued.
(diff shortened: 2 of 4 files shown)


app/jobs/maintenance_tasks/task_job.rb
```diff
@@ -90,7 +90,7 @@ def after_perform
     end
 
     def on_error(error)
-      @ticker.persist
+      @ticker.persist if defined?(@ticker)
```
rafaelfranca, line 93: Why the variable would not be defined? Because `before_perform` did run?

etiennebarrie: `before_perform` runs but has an exception before `@ticker` is set, on line 57:

https://github.com/Shopify/maintenance_tasks/blob/42fd4ee3cec2286f4a1f53f31f51c6402101e587/app/jobs/maintenance_tasks/task_job.rb#L55-L62

etiennebarrie: At first I just used `@ticker&.persist` but it didn't pass the test because we use verbose mode. We could also add a `@ticker = nil` at the beginning of `before_perform` too but I felt that would feel weird.

etiennebarrie:
Worth noting that a similar exception could happen if there's a job for a run that is then removed, except the exception would be during deserialization, but then in `on_error` we would fail because we access `@run`. This is not something that can be done without accessing a console or removing a row in the DB though.

This PR solves an issue that can happen with a simple rollback before the job starts running or while the job is still running (it will be interrupted, then would fail upon resuming).

adrianna-chang-shopify (approved):
Nice find Étienne! Changes LGTM, all was as expected on the tophat 👍

One small request, could we tweak:
https://github.com/Shopify/maintenance_tasks/blob/2a3f4bda7b5caadae7f17044f0c834d2e0791fca/app/validators/maintenance_tasks/run_status_validator.rb#L14

to add
```
 # enqueued -> errored occurs when the task job fails to be enqueued,
or if the Task is deleted before is starts running.
```

rafaelfranca (approved):

PR #295, opened 2021-01-11
Remove stylesheet_link_tag patch

etiennebarrie:
With #294, and assuming our users use either 6.0 or 6.1.1 but not 6.1.0, we can remove this patch that was merged upstream in https://github.com/rails/rails/pull/40841 (included in 6.1.1 with https://github.com/rails/rails/commit/80077d3675b9baceb0fee2a9bef8fcce9cf7080c)

I'm not sure it warrants adding that to the gemspec given our stance that dependencies should always be up-to-date.


volmer (approved):

adrianna-chang-shopify (approved):
🙏

PR #297, opened 2021-01-12
Add ActiveStorage to the dummy app

adrianna-chang-shopify:
Closes: https://github.com/Shopify/maintenance_tasks/issues/200

CSV functionality has been requested by a number of users at this point, so I think it makes sense to start building out this feature.

This PR sets up ActiveStorage in the dummy app by adding the necessary tables (from https://edgeguides.rubyonrails.org/active_storage_overview.html#setup, these were generated by doing `cd test/dummy; rails active_storage:install; rails db:migrate`).

We've settled on using the ActiveStorage library Rails provides because it seems reasonable for an uploaded CSV to be attached to a `Run` record, and ActiveStorage makes it simple for users to set up their apps with a cloud storage solution to store their uploaded CSVs.

### DIFF
--- a/test/dummy/db/migrate/20210112141723_create_active_storage_tables.active_storage.rb
+++ b/test/dummy/db/migrate/20210112141723_create_active_storage_tables.active_storage.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+# This migration comes from active_storage (originally 20170806125915)
+class CreateActiveStorageTables < ActiveRecord::Migration[5.2]
+  def change
+    create_table(:active_storage_blobs) do |t|
+      t.string(:key,          null: false)
+      t.string(:filename,     null: false)
+      t.string(:content_type)
+      t.text(:metadata)
+      t.string(:service_name, null: false)
+      t.bigint(:byte_size,    null: false)
+      t.string(:checksum,     null: false)
+      t.datetime(:created_at, null: false)
+
+      t.index([:key], unique: true)
+    end
+
+    create_table(:active_storage_attachments) do |t|
+      t.string(:name, null: false)
+      t.references(:record,   null: false, polymorphic: true, index: false)
+      t.references(:blob,     null: false)
+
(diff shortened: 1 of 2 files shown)


etiennebarrie (approved):
Shouldn't we add it to the gemspec as well? The dummy app/Gemfile depends on rails but the gem not on activestorage. Or we can wait until the gem really needs it. Or maybe we want the gem to not depend on it so that it doesn't force it to users, if they don't use active storage they don't get the CSV feature?

But for now we'll need this anyway. 👍

adrianna-chang-shopify:
That's an interesting point, @etiennebarrie. I had assumed that we wouldn't want to force it on users who don't need CSV processing. Plus, even if we make ActiveStorage a dependency of the gem, if the users haven't done the setup steps in their own app then the gem will break for them if they try to run a task that takes a CSV file. So I feel like it's reasonable to expect users to require the framework if they don't have it (presumably most people will just have the full rails framework anyways), and to additionally note that they need to do the ActiveStorage setup if they intend to use CSVs.

But, maybe it's also okay to just require it as a dep of the gem 😁  @rafaelfranca any opinion here?

etiennebarrie:
Yeah I think you're right and we shouldn't depend on it. We may want to try to not breaking if users don't have active storage properly setup, it should just not offer CSV/file uploads in that case. But maybe it's a pain and we can just expect users to have active storage properly setup before writing a CSV Task.

adrianna-chang-shopify:
> We may want to try to not breaking if users don't have active storage properly setup, it should just not offer CSV/file uploads in that case. 

I think that makes a lot of sense! Shouldn't be too difficult either - we could create a `CSVTaskGenerator` that checks for the `ActiveStorage` constant (it could additionally ensure that the migrations have been run), and fails with a warning if the host app is not set up properly. 

Beyond that, outlining it in the Readme should suffice. I'll ship this for now, we can always revisit adding it to the gemspec later on if we change our minds 😄

PR #300, opened 2021-01-14
Attach CSVs to runs

adrianna-chang-shopify:
Smallish PR, as ActiveStorage handles all of the complexity associated with attaching an uploaded file to an ActiveRecord.
This PR attaches a `csv_file` to our Run model using the `has_one_attached` macro (this will always be a 1:1 relationship).

The `Runner` class is now responsible for attaching an incoming `csv_file` to newly created runs in its `#run` method. It will be up to the `TasksController` / CLI to pass the CSV from the UI or command line to the runner.

### DIFF
--- a/test/models/maintenance_tasks/runner_test.rb
+++ b/test/models/maintenance_tasks/runner_test.rb
@@ -81,5 +81,15 @@ module MaintenanceTasks
         )
       end
     end
+
+    test '#run attaches CSV file to Run if one is provided' do
+      csv_file = file_fixture('sample.csv')
+      uploaded_csv = Rack::Test::UploadedFile.new(csv_file, 'text/csv')
+      @runner.run(name: @name, csv_file: uploaded_csv)
+
+      run = Run.last
+      assert_predicate run.csv_file, :attached?
+      assert_equal File.read(csv_file), run.csv_file.download
+    end
   end
 end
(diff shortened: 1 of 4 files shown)


test/models/maintenance_tasks/runner_test.rb
```diff
@@ -81,5 +81,17 @@ class RunnerTest < ActiveSupport::TestCase
         )
       end
     end
+
+    test '#run attaches CSV file to Run if one is provided' do
+      csv_file = File.open(
```
etiennebarrie, line 86: Let's try to use `file_fixture`.

test/models/maintenance_tasks/runner_test.rb
```diff
@@ -81,5 +81,17 @@ class RunnerTest < ActiveSupport::TestCase
...
+        Engine.root.join('test', 'fixtures', 'maintenance_tasks', 'sample.csv')
+      )
+      uploaded_csv = Rack::Test::UploadedFile.new(csv_file, 'text/csv')
+      @runner.run(name: @name, csv_file: uploaded_csv)
+
+      run = Run.last
+      assert run.csv_file.attached?
```
etiennebarrie, line 93: assert_predicate?

etiennebarrie (approved):

PR #305, opened 2021-01-19
Finish Rails 6.1 upgrade

etiennebarrie:
This runs the `rails app:update` task in the dummy app, and then goes through the update process.

Also I changed the behavior to raise on deprecations for tests.


test/dummy/config/environments/development.rb
```diff
@@ -42,6 +44,12 @@
   # Print deprecation notices to the Rails logger.
   config.active_support.deprecation = :log
 
+  # Raise exceptions for disallowed deprecations.
+  config.active_support.disallowed_deprecation = :raise
```
adrianna-chang-shopify, line 48: Nice 👍

adrianna-chang-shopify (approved):
🚢

test/dummy/config/environments/development.rb
etiennebarrie: Oh that's actually just the default for the new feature for disallowed deprecations, and we don't have any disallowed deprecations.
But I now realize I haven't changed the comment above the change to raise exceptions for deprecations in the test environment. Will fix that.

Issue #304, opened 2021-01-18
Avoid missing configuration for ActiveJob queue adapter


etiennebarrie:
Let's add a line to the README about making sure that active_job.queue_adapter is configured in the application configuration/environment.

etiennebarrie:
I'm thinking we may also want to add a notice during the installation as well, checking if they have something else than async configured.

Pros:
* It's right there when they install the gem.
* It doesn't show if it's not needed.
  * But if it's also in the README, users will check anyway, so that may be useless.
Cons:
* It will be a false positive if people only use Sidekiq in production but not in development.
* It requires booting the app to know the configuration (e.g. a `rails runner`)

PR #306, opened 2021-01-20
Add note about ActiveJob to README

adrianna-chang-shopify:
Closes: https://github.com/Shopify/maintenance_tasks/issues/304

We should document in the installation step of the README that the queue adapter for ActiveJob needs to be configured to a persistent backend (ie. not the default async backend) in order for tasks continue running through infra changes / code deploys / etc...

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@ A Rails engine for queuing and managing maintenance tasks.
 
 * [Demo](#demo)
 * [Installation](#installation)
+  * [Active Job Dependency](#active-job-dependency)
 * [Usage](#usage)
   * [Creating a Task](#creating-a-task)
     * [Considerations when writing Tasks](#considerations-when-writing-tasks)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -52,6 +52,16 @@ In case you use an exception reporting service (e.g. Bugsnag) you might want to
 define an error handler. See [Customizing the error
 handler](#customizing-the-error-handler) for more information.
 
+**ActiveJob Dependency**
```
etiennebarrie, line 55: I think elsewhere in this document we've split the camel case words, I think we should do the same here.

Also my markdown linter tells me this should be a header instead of emphasis.

```suggestion
#### Active Job Dependency
```

README.md
```diff
@@ -52,6 +52,16 @@ In case you use an exception reporting service (e.g. Bugsnag) you might want to
...
+
+The Maintenance Tasks framework relies on ActiveJob behind the scenes to run
+Tasks. The default queuing backend for ActiveJob is
+[asynchronous](https://api.rubyonrails.org/classes/ActiveJob/QueueAdapters/AsyncAdapter.html).
```
etiennebarrie, line 59: Can we pull out URLs from the text and use markdown references like I did in other places in the README?

README.md
```diff
@@ -52,6 +52,16 @@ In case you use an exception reporting service (e.g. Bugsnag) you might want to
...
+It is **strongly recommended** to change this to a persistent backend so that
+Task progress is not lost during code or infrastructure changes. For more
+information on configuring a queuing backend, take a look at the [ActiveJob
+documentation](https://guides.rubyonrails.org/active_job_basics.html#setting-the-backend).
```
etiennebarrie, line 63: I feel like it's missing something along the lines of "you may be already using Sidekiq but may not have set the queue adapter". Or maybe we could give them a thing to grep or a line to run in their console (i.e. `active_job.queue_adapter`).

But I may be wrong, telling them we use ActiveJob and pointing them to the docs on how to configure the queue adapter should be good!

etiennebarrie (approved):

README.md
adrianna-chang-shopify: Yes, although it still goes over the 80 character limit 😢

README.md
adrianna-chang-shopify: I think the `Setting the Backend` section in the docs is pretty straightforward in terms of telling users they need to set it as a config option. If folks are still getting confused about using Sidekiq but not using ActiveJob, we can add an extra note later, but I'd rather keep it lean for now!

Issue #207, opened 2020-11-30
Provide support for enumerable collections beyond Array and ActiveRecord::Relation in TaskJob


adrianna-chang-shopify:
Will be partially solved when CSV functionality is implemented, after which we will decide whether further collections should be supported.

sambostock:
👋 I was going to open a new issue, but this seemed like a more appropriate place to chime in.

----

**TL;DR: Lack of support for custom enumerators prevents certain use cases, such as processing external resources.**

----

I was investigating adopting this gem for our app (`Shopify/athena`), and support being limited to `ActiveRecord::Relation` & `Array` (and CSVs, once that is added) would block us from adopting it for many of our use cases.

Athena manages many resources in Twilio. One of the things we do is run a monthly job which iterates over certain resources which are over a month old and deletes them. For this, we use `Shopify/job-iteration`, and a custom enumerator which "streams" in all the results for our search, and iterates over them until they're all gone. In some other jobs which don't delete the resources, we use their `start_time` as an increasing cursor.

This works because `JobIteration` supports custom `Enumerator`s. Given the current implementation, we would have to pre-load all the resources (or IDs) into an Array before starting to process them, or some other workaround.

**It would be great if `MaintenanceTasks` supported custom enumerators, similarly to `JobIteration`!**

sambostock:
As per the discussion in https://github.com/Shopify/maintenance_tasks/pull/307#discussion_r561420323, I'm going to expand on some of our use cases, with respect to processing API resources in Twilio.

Most of these tasks run on a schedule, though we have had one-off tasks come and go.

- Synchronization tasks: We have tasks which iterate through resources in Twilio, typically using a timestamp field (`created_at`, `updated_at`, `event_date`, etc.) as a cursor, which we want to upsert into our SQL "mirror" database
- Deletion tasks: We delete various resources as they exit our retention period. Some of these are "mirrored" in our database, so we enumerate over the records, deleting the resources and their record, one by one. For resources which are not mirrored, we must iterate via the API, again with some time based cursor, but this time with extra parameters to narrow the time range.
- One-off "backfill" tasks:
  - After adding code to automatically move call recordings from Twilio into GCS once ready, we had to iterate through existing recordings and move them over.
  - After adding some code to populate a worker attribute after worker creation, we had to backfill the attribute for all existing workers.

In these examples, we want to iterate over a set of resources

- not available as an `ActiveRecord::Relation`
[...]

PR #307, opened 2021-01-20
Add support for dynamic collections

sambostock:
## Context

Currently, the framework supports two/three types of collections:

- Instances of `Array`
  _(delegating to `enumerator_builder.build_array_enumerator(collection, cursor: cursor)`)_
- Instances of `ActiveRecord::Relation`
  _(delegating to `enumerator_builder.active_record_on_records(collection, cursor: cursor)`)_
- CSVs (WIP as far as I can tell)
  _(probably using `JobIteration::CSVEnumerator`)_
  
This presents problems for use cases such as:

- Tasks which need to process external resources (requiring custom enumeration)
- Tasks which need to process active records in batches

## Proposed Solution

**This PR adds support for the `collection` method to return an object responding to `.call(cursor:)`, which when called should return a `JobIteration` style enumerator (successively yielding a pair containing an item to process, and a cursor identifying it).**

This alternative was chosen, as it is non-breaking, and provides a "we'll handle the simple cases, you handle the advanced ones" approach for consumers. **Of course, feedback is welcome!**

#### Return an `Enumerator` from `Task#collection` in special cases

While we could check if the result of `collection` is an `Enumerator` already, and forward it to just return it from `TaskJob#build_enumerator`, we have no way of injecting the `cursor`, meaning the enumerator would have no way to resume from a certain spot.

We could do something like optionally pass `cursor:` to `collection`, but that would require checking its parameters and would result in a weird API.

#### Require `collection` to always return an `Enumerator`

This would be a breaking change to the API, and would complicate the simple use cases (relations & arrays), since currently they don't need to know anything about cursors.

#### Check another method first

We could first check if `CustomTask#enumerator` exists, and if so, call it with `cursor:`, falling back to the current `collection` call otherwise, but this would mean the methods defined by `Task`s would differ from each other. That said, perhaps this makes sense if we introduce the concept of a `StaticTask` and a `DynamicTask`, and move the enumerator creation into the task itself. 🤔 

## For Reviewers

I have split the PR into three commits with detailed commit messages. In short though:

- **Add `Maintenance::DynamicTask` dummy task for tests**: stands alone to isolate unrelated changes
- **Add support for dynamic collection**: this is where the feature is actually added
- **Add dynamic collection parameter introspection**: this adds some more rigorous checking of the arguments accepted by the "callable", so we can produce a better error message if it doesn't accept a `cursor:` keyword argument. I'm not sure this is required, so I left it on its own so it can easily be dropped depending on reviewer feedback 👌 

----
Related to #207


app/jobs/maintenance_tasks/task_job.rb
```diff
@@ -94,5 +99,31 @@ def on_error(error)
...
+      else
+        collection.method(:call)
+      end.parameters
+
+      return true if parameters.any? do |parameter|
+        type = parameter.first
+        (type == :key || type == :keyreq) && parameter.last == :cursor
+      end
+
+      return true if parameters.any? do |parameter|
```
rafaelfranca, line 121: Why we are iterating over the same collection twice? The content of those loops are very similar, so we can appply boolean logic to avoid this kind of duplication. That way we also remove those early returns and the `false` in the end of the method.

sambostock: I had originally kept them separate because I figured the conditional might get confusing. I've combined them, and opted not to extract any helper methods for now. Let me know what you think.

rafaelfranca (approved):
Implementation LGTM and the feature as well, but let's wait @adrianna-chang-shopify's review

app/jobs/maintenance_tasks/task_job.rb
```diff
@@ -35,8 +35,13 @@ def build_enumerator(_run, cursor:)
       when Array
         enumerator_builder.build_array_enumerator(collection, cursor: cursor)
       else
-        raise ArgumentError, "#{@task.class.name}#collection must be either "\
-          'an Active Record Relation or an Array.'
+        if dynamic_collection?(collection)
+          collection.call(cursor: cursor)
```
sambostock, line 39: It just occurred to me while exploring alternatives that this wouldn't address the case where we want to `enumerator_builder.active_record_on_batches`, since we don't have a reference to `enumerator_builder`.

I wonder if instead it would make sense to set `@cursor` on the job, and then inject the entire job in, so you can something like

```ruby
def collection
  lambda do |job:|
    job.enumerator_builder.active_record_on_batches(
      Product.all,
      cursor: job.cursor,
      batch_size: 100,
    )
  end
end
```

This might even be useful for future use-cases like run-specific params 🤔 

```ruby
def collection
  lambda do |job:|
    job.enumerator_builder.active_record_on_batches(
      Product.where(shop_id: job.arguments.second[:shop_id]),
      cursor: job.cursor,
      batch_size: 100,
    )
  end
end
```

rafaelfranca: This is leaking too much of the implementation to the users in my option. People should not even know a cursor exists, or an enumerator builder. What people want is to be able to create batches on in the collection, so maybe we should use a declarative approach in the task like:

```
run_in_barches_of 100

def collection
  Product.all
end
```

sambostock: One of the questions I struggled with when designing this was

> How to accommodate complex use cases without adding complexity to the simple ones?

I see what you mean, but I'd argue that in the complex use case the author already has to be aware of the cursor (and why it's needed) and it's not that much more complexity to be aware of the enumerator builder. We could always pass in some sort of simpler object than the entire job, like

```ruby
CollectionContext = Struct.new(:cursor, :enumerator_builder, keyword_init: true)
```

which would easily allow future expansion, such as forwarding run-specific parameters. Alternatively, we just make a breaking change if the time comes. 🤷‍♂️ 

The `run_in_batches_of` idea is intriguing though.

----

Another thing I noticed in the [implementation of `JobIteration::EnumeratorBuilder`](https://github.com/Shopify/job-iteration/blob/ae9f11e316c99eaf0c5a478c1fa0a659fcd7730a/lib/job-iteration/enumerator_builder.rb#L11-L27) is that there may be plans to force enumerators to be wrapped in a `JobIteration::EnumeratorBuilder::Wrapper`:

```ruby
    # These wrappers ensure we have a custom type that we can assert on in
    # Iteration. It's useful that the `wrapper` passed to EnumeratorBuilder in
    # `enumerator_builder` is _always_ the type that is returned from
    # `build_enumerator`. This prevents people from implementing custom
    # Enumerators without wrapping them in
    # `enumerator_builder.wrap(custom_enum)`. We don't do this yet for backwards
    # compatibility with raw calls to EnumeratorBuilder. Think of these wrappers
    # the way you should a middleware.
    class Wrapper < Enumerator
      def self.wrap(_builder, enum)
        # ...
```

If this comes to fruition, it may force the custom enumerator use case to be aware of the `enumerator_builder` so it can call `.wrap` on it. 🤔 Or maybe we just auto-wrap if needed.

rafaelfranca: One of the design philosophy of this library is make complex use cases easy to do. We don't aim to give all control and power to the users so they can build complex code in their apps. The code written to use this app should be easy to read and complex things should be made simple with less code possible.

In other words, we should focus our energy to implement something that is easy to use, not in allowing people to have access to all the nobs and implementation detail possible.

I think the fact that anyone need to now about this cursor is a huge design flaw in the job_iterator library. People usually don't do anything with this cursor, they don't define anything different, they just pass around. If people always pass around the same thing, why do we make them pass around anyway?

I don't want this library to make the same mistakes. The API needs to follow more conventions, use less configurations, make complex cases easier.

So maybe instead of accepting this PR as is I should be asking, is there anything we can make easier  in this library so you don't need to write a enumerator for your task? The CSV use case was one of them.

sambostock: Thanks for the context and thorough review @rafaelfranca 👍

> is there anything we can make easier in this library so you don't need to write an enumerator for your task?

I have added some more concrete example use cases to https://github.com/Shopify/maintenance_tasks/issues/207#issuecomment-764189242. Ultimately though, they're similar to those described in [`JobIteration`'s _Custom Enumerator Guide_](https://github.com/Shopify/job-iteration/blob/master/guides/custom-enumerator.md): the application needs to iterate over a resource type the iteration framework should not know about.

I spent some time thinking about this (maybe too much), and think I figured out why it's so tricky: `JobIteration`'s `build_enumerator` requires knowledge about _how to build the enumerator_ **and** _how far along it is in the iteration_. If instead, we could build an object that knows how to iterate, but defer telling it how far it is in the iteration, then we could encapsulate the `cursor`.

It's really rough, but I've tried to illustrate my idea in [this Gist](https://gist.github.com/sambostock/3323f401fd3061f1f8e620fa92b44da1), the key being basically this change:

```diff
 def interruptible_perform(*arguments)
    # ...
-   enumerator = build_enumerator(*arguments, cursor: cursor_position)
+   enumerator = build_enumerator(*arguments).call(cursor: cursor_position)
    # ...
 end
```

**If we did that, only maintainers and custom enumerator authors would need to know about `cursor`, not consumers using existing enumerators.**

Of course... that would likely be breaking changes to both `JobIteration` and `MaintenanceTasks`... 😞

adrianna-chang-shopify (commented):
Hey @sambostock, thanks for the extremely descriptive and thoughtful PR ❤️ 

Yeah, unfortunately we're constrained by JobIteration's API and the requirement to supply a cursor when building the enumerator. We would definitely like to preserve our existing API and opt to support the simplest of use cases as Rafael mentioned.

I think this is the best solution for the reasons you mentioned - gives users the option of supplying a dynamic collection without breaking any existing APIs, and still keeps the API relatively decoupled from JobIteration. Thanks for providing examples - I think you make a very good point that the use cases described in https://github.com/Shopify/maintenance_tasks/issues/207#issuecomment-764189242 cannot be solved by the library without allowing a collection that knows about a cursor to be defined.

I'm not a huge fan of the parameter introspection, as it's adding a good deal of complexity for the sake of providing a nicer error message. Maybe we could opt to just call ` collection.call(cursor: cursor)` if `collection.respond_to?(:call)`, and then if it raises an `ArgumentError`, we could rescue and reraise it with a better error message? (`"Dynamic collection must respond to .call(cursor:)."`? Or maybe just drop it altogether, like you mentioned 😄 

Could we also add a section to the README for this feature? I would document it at the bottom in its own section so we can leave the focus on the simple use cases like Rafael mentioned.

Other than that, code LGTM, and again, appreciate the amount of thought you've put into this ❤️

etiennebarrie:
The name is not working for me, what is more dynamic about this collection than others? It's just a task based on enumerator/cursor rather than the simple case of collection. Actually now that I think of it, that's what bothers me about this, we shouldn't make it about a collection being dynamic, but more about a task that's based on an enumerator instead of a collection.

So from the alternatives mentioned, I think "Check another method first" is the one that makes the most sense to me.

The current solution feels kind of hijacking the collection method, to actually return a callable instead of a collection, and then we have a some amount of code to handle that case: does it respond to call? is it directly a method/proc or just an object? and what are his parameters?

I don't think we should be doing all of this. Using another message would solve all of this: does the task respond to `enumerator`, if it does we're good, we just send it the cursor, and pass that back to job iteration.

---

There's another argument that Rafael makes as to whether we want this entirely or not. Most of the time users shouldn't have to deal with a cursor, and we can solve specific features like batches by providing an API. But I think it makes sense for other things like APIs where you have a `since`, `after` or `before`  cursor, that can't work without passing a cursor to the task.

sambostock:
Thanks for the reviews @adrianna-chang-shopify & @etiennebarrie!

----

> Could we also add a section to the README for this feature? I would document it at the bottom in its own section so we can leave the focus on the simple use cases like Rafael mentioned.

💯 I've added it to my "Before Merging" checklist, but I'll hold off on writing it until we settle on the implementation/API.

----

> I'm not a huge fan of the parameter introspection,

Yeah, I'm not either. 😅 However, the reason I didn't want to `rescue ArgumentError` and re-raise with a custom error message is that erroneously replacing an unrelated `ArgumentError` with a misleading error message would be even worse than an unhelpful one.

Removing the check or switching to the `def enumerator(cursor:)` would address the complexity increase, so I'll do one of those two. 👍

---

> does the task respond to `enumerator`, if it does we're good, we just send it the cursor, and pass that back to job iteration.

This approach does come with one trade off: the API is slightly blurred. I am now expected to implement exactly one of two different API methods. How do we make this clear to the consumer and provide useful feedback if they implement it incorrectly?

For example, we document the `collection` method

```ruby
# Placeholder method to raise in case a subclass fails to implement the
# expected instance method.
#
def collection
  raise NotImplementedError,
    "#{self.class.name} must implement `collection`."
end
```

if we document `Task#enumerator` in the same way it breaks `task.respond_to?(:enumerator). If we don't document it, then _we haven't documented it_. We also don't really have a way to make them mutually exclusive, though I don't know if that's an issue.

One possible alternative would be to move the `enumerator` building from `TaskJob` into this new `enumerator` method

```ruby
class TaskJob
  attr_reader :cursor
  def build_enumerator(_run, cursor:)
    @cursor = cursor || @run.cursor
    @task.enumerator(job: self) # delegate, instead of building enumerator here
  end
end

class Task
  # Documentation for special case consumers would go here
  def enumerator(job:)
    collection = self.collection
    
    case collection
    when ActiveRecord::Relation
      job.enumerator_builder.active_record_on_records(collection, cursor: job.cursor)
    when Array
      job.enumerator_builder.build_array_enumerator(collection, cursor: job.cursor)
    else
        'an Active Record Relation, or an Array'
    end
  end
  
  # collection method unchanged
end

# Backwards compatible with existing API
class MySimpleTask < Task
  def collection() = Model.all
  def count() = Model.count
  def process(model) = puts "Processed #{model}"
end

# Supports customization
class MyCustomTask < Task
  def enumerator(job:) = SomeCustomEnumerator.new('whatever', cursor: job.cursor)
  def process(item) = puts "Processed custom enumerated #{item}"
end
```

This is slightly different to the discussion started in https://github.com/Shopify/maintenance_tasks/pull/307#discussion_r561364421, but @rafaelfranca's comments about revealing implementation details still may still apply. That said, it
- simplifies the method calls (no need for `.respond_to?`)
- is entirely backwards compatible
- reveals no complexity to consumers with simple use cases

----

> Most of the time users shouldn't have to deal with a cursor, and we can solve specific features like batches by providing an API.

Yeah, I agree with you and @rafaelfranca that we can provide batch functionality via a simple API, instead of requiring the consumer to build the custom enumerator. 👍 I'll consider that out of scope for this PR.

sambostock (commented):
I've changed the implementation to use a `Task#enumerator` method instead, which be default returns the appropriate `Enumerator` for the `Task#collection`, but can be overridden by consumers.

I'm not sure which approach is better: this `enumerator` one, or the original  `collection` returning a `callable`. Both have tradeoffs.

app/jobs/maintenance_tasks/task_job.rb
```diff
@@ -25,21 +25,17 @@ def retry_on(*, **)
...
-        enumerator_builder.build_array_enumerator(collection, cursor: cursor)
-      when CSV
-        JobIteration::CsvEnumerator.new(collection).rows(cursor: cursor)
-      else
-        raise ArgumentError, "#{@task.class.name}#collection must be either "\
-          'an Active Record Relation, Array, or CSV.'
-      end
+      @task.enumerator(context: EnumeratorContext.new(
+        cursor: cursor || @run.cursor,
+        enumerator_builder: enumerator_builder,
```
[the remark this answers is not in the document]
etiennebarrie: I think we could do without the EnumeratorContext. If you look at JobIteration, the only place where it uses the job passed to EnumeratorBuilder is in the ThrottleEnumerator, in order to retry jobs.
We plan on implementing throttling but we can work around the EnumeratorBuilder, in any case it's a problem for later.
That means we could manage by passing only the cursor to `Task#enumerator`, and creating a EnumeratorBuilder without a job in Task. We used to do that BTW, see #303. (Though I don't think we should expose the enumerator builder to Task, just instantiate it in `enumerator` where we need it).

app/tasks/maintenance_tasks/task.rb
```diff
@@ -66,6 +75,50 @@ def load_constants
...
+    #
+    #       def process(resource)
+    #         puts "Processed #{resource}"
+    #       end
+    #     end
+    #
+    # @param context [EnumeratorContext] context for the enumerator, including
+    # a cursor
+    # @return the enumerator.
+    def enumerator(context:)
```
etiennebarrie, line 100: Yeah I think that's the right spot to put it, it's really neater like this.

adrianna-chang-shopify (commented):
Yeah, there are definitely tradeoffs in all directions with this one, so thanks for sticking it out!

> I'm not sure which approach is better: this enumerator one, or the original collection returning a callable. Both have tradeoffs.

IMHO the new solution is a nicer API. It's a bit awkward to require users to have their `#collection` method return a callable that ends up building an enumerator, when instead we could just have them build the enumerator directly (as with the new approach).

There's one thing that's still a bit problematic to me about the API. Even if we delegate building the enumerator to Task, and allow special case consumers to override this method, there is still some ambiguity in the API as they are expected to implement this new `enumerator` method, but not collection. I guess there's no harm in having a collection defined for Tasks that defined their own enumerator, as it will get completely ignored, but I'm still a bit worried this will cause confusion. I don't think we've really addressed the fact that users still need to implement one of two methods that are essentially mutually exclusive.  This is not a blocker at all, and good documentation is hopefully sufficient for communicating which API the consumer is expected to follow, I guess I'm just reasoning about the fact that our API is still ambiguous even with the new approach and with shifting the enumerator building code into Task :sweat_smile: (Although I do still agree that documenting it as a method subclasses can override, and not having to rely on `respond_to?` is a favourable approach)

On the subject of naming, the new approach is a lot more conducive to a more descriptive name - maybe something like `CustomEnumeratorTask`? That fits with `JobIteration`'s language.

Anyways, I'll let @etiennebarrie chime in with his thoughts, but I am completely willing to have this new approach ship! I'll hold out on putting the green check until we confirm the approach, and the tests/docs are updated. Thanks again for working on this!

etiennebarrie (commented):
I really like `Task#enumerator`, I think we should probably avoid the `EnumeratorContext` since we don't really have a use for it yet.

sambostock:
@adrianna-chang-shopify another alternative I considered was to simply have two different classes one could inherit from, which have different APIs: one has `enumerator` "not implemented", and the other implements it and introduces "collection". This might be awkward though, since we'd be inserting a class into the existing hierarchy, and might have to do weird stuff to accommodate `DescendantsTracker`. Nonetheless, we could do something like this

```ruby
class EnumeratorTask
  def enumerator(context:) = raise NotImplementedError
  def count = nil
  def process(item) = raise NotImplementedError
end

class CollectionTask < EnumeratorTask
  def enumerator(context:)
    # ...build enumerator around collection
  end
  
  def collection = raise NotImplementedError
end
Task = CollectionTask
```

Alternatively, we could stick to the two single methods, and implement `Task.method_added` to track the addition of one method and raise if the other is added too (similarly to how `JobIteration` prevents `perform` from being redefined).

app/jobs/maintenance_tasks/task_job.rb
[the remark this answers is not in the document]
adrianna-chang-shopify: Nice catch Étienne 👏  Yeah let's use `JobIteration.enumerator_builder` and we can revisit what needs to be done fpr throttling in the future.

adrianna-chang-shopify:
Actually @sambostock, what do you think of having custom enumerator functionality coming in the form of a Module? This is what we're doing for Tasks that need to process CSVs (we opted not to use inheritance for the exact reason of avoiding extra complexity with `DescendantsTracker`).

(See https://github.com/Shopify/maintenance_tasks/blob/7131371ad6e7cf0604ad54f27b6a01b7914dea56/app/tasks/maintenance_tasks/csv_task.rb)

Something like this:
```ruby
module EnumeratorTask
  def enumerator(cursor:) = raise NotImplementedError # Assuming we're able to drop use of context and just use cursor
  def collection = nil
  def process(item) = raise NotImplementedError
end

class MyCustomTask < Task
 include EnumeratorTask
  def enumerator(cursor:) = SomeCustomEnumerator.new('whatever', cursor: cursor)
  def process(item) = puts "Processed custom enumerated #{item}"
end
```

sambostock:
@adrianna-chang-shopify how do you see that being used compared to a standard `Task` within the framework? Would we keep the logic in `TaskJob`?

```ruby
def build_enumerator(_run, cursor:)
  cursor ||= @run.cursor
  return @task.enumerator(cursor: cursor) if @task.is_a?(EnumeratorTask)

  collection = @task.collection
  case collection
  # ...
end
```

I worry that this feels a little like `respond_to?(:enumerator)`, but with extra steps... 😅

adrianna-chang-shopify:
@sambostock I am very much in favour of your suggestion to move `enumerator` (and the logic to build enumerators using JobIteration's API) to Task!

So my proposition of using a mixin wouldn't really change anything with your most recently proposed solution, it would just be a way to define the API more clearly for Enumerator Tasks, and a way to ensure that these Tasks are adhering to defining `enumerator` properly.

```ruby
# task_job.rb (No changes)
def build_enumerator(_run, cursor:)
  @task.enumerator(cursor: cursor)
end

# task.rb (No changes)
def enumerator(cursor:)
    collection = self.collection

    case collection
    when ActiveRecord::Relation
      JobIteration.enumerator_builder.active_record_on_records(
        collection,
        cursor: cursor,
      )
    when Array
      JobIteration.enumerator_builder.build_array_enumerator(
        collection,
        cursor: cursor,
      )
    when CSV
      JobIteration::CsvEnumerator.new(collection).rows(cursor: cursor)
    else
        'an Active Record Relation, an Array, or a CSV.'
    end
end

# enumerator_task.rb
module EnumeratorTask
  def enumerator(cursor:)
   # Tasks that include this Module must define their own #enumerator
    raise NotImplementedError, "#{self.class.name} must implement `enumerator(cursor:)`." 
  end
  def collection
    nil
  end
end

# my_custom_task.rb
class MyCustomTask < Task
 include EnumeratorTask
  def enumerator(cursor:)
    SomeCustomEnumerator.new('whatever', cursor: cursor)
  end

  def process(item)
    puts "Processed custom enumerated #{item}"
  end
end
```

And then:
```ruby
#another_custom_task.rb
class AnotherCustomTask < Task
 include EnumeratorTask
  def process(item)
    puts "Processed custom enumerated #{item}"
  end
end

AnotherCustomTask.new.enumerator(cursor: 1)
> NotImplementedError: MaintenanceTasks::MyCustomTask must implement `enumerator(cursor:)`.
```

So, any Tasks that includes `EnumeratorTask` would _need_ to provide their own definition for `enumerator`, but any other type of Task that doesn't include the mixin would just delegate `enumerator` to the method in `Task`.

What do you think?

PR #308, opened 2021-01-21
Allow uploading CSVs to CSV supporting Tasks through the UI

adrianna-chang-shopify:
Closes: https://github.com/Shopify/maintenance_tasks/issues/202

Allows files to be uploaded for CSV Tasks in the UI. I've modified any of the action partials that allow a Task to be run to use the [form_helper](https://api.rubyonrails.org/v6.1.1/classes/ActionView/Helpers/FormHelper.html#method-i-form_with) so that we can allow files to be submitted via a file input upload tag. The views call a new method `TaskData#csv_task?` to determine whether to render the file upload element.

I've also modified the TasksController show action to only set `@refresh` if there is an active run - this avoids unnecessarily refreshing when nothing is happening, which is particularly important for CSV Tasks, since the form might get refreshed before it is submitted with the current refresh behaviour (wiping the uploaded file).

We may want to look into allowing CSV tasks to run from the command line, but this is out of scope for this PR.

🎩 

**You can tophat the success case using `Maintenance::ImportPostsTask` and uploading `sample.csv` from the fixtures**
Screen Shot 2021-01-21 at 3 26 32 PM

**Failing to upload a CSV:**
Screen Shot 2021-01-21 at 3 34 54 PM

**Uploading a non-CSV file:**
Screen Shot 2021-01-21 at 3 37 00 PM

(Most scenarios where the CSV is invalid / not a CSV / does not conform to what is expected by `CSvTask#process` will simply be rescued and cause the Run to become `errored`, and don't need special treatment)

### DIFF
--- a/app/controllers/maintenance_tasks/tasks_controller.rb
+++ b/app/controllers/maintenance_tasks/tasks_controller.rb
@@ -19,12 +19,16 @@ module MaintenanceTasks
     # Shows running and completed instances of the Task.
     def show
       @task = TaskData.find(params.fetch(:id))
+      set_refresh if @task.last_run.active?
       @pagy, @previous_runs = pagy(@task.previous_runs)
     end
 
     # Runs a given Task and redirects to the Task page.
     def run
-      task = Runner.run(name: params.fetch(:id))
+      task = Runner.run(
+        name: params.fetch(:id),
+        csv_file: params[:csv_file]
+      )
       redirect_to(task_path(task))
     rescue ActiveRecord::RecordInvalid => error
       redirect_to(task_path(error.record.task_name), alert: error.message)
--- a/app/controllers/maintenance_tasks/tasks_controller.rb
+++ b/app/controllers/maintenance_tasks/tasks_controller.rb
@@ -7,7 +7,7 @@ module MaintenanceTasks
   #
   # @api private
   class TasksController < ApplicationController
-    before_action :set_refresh, only: [:index, :show]
+    before_action :set_refresh, only: [:index]
 
     # Renders the maintenance_tasks/tasks page, displaying
     # available tasks to users, grouped by category.
@@ -19,7 +19,7 @@ module MaintenanceTasks
     # Shows running and completed instances of the Task.
     def show
       @task = TaskData.find(params.fetch(:id))
-      set_refresh if @task.last_run.active?
+      set_refresh if @task.last_run&.active?
       @pagy, @previous_runs = pagy(@task.previous_runs)
     end
 
--- a/app/views/maintenance_tasks/tasks/actions/_succeeded.html.erb
+++ b/app/views/maintenance_tasks/tasks/actions/_succeeded.html.erb
@@ -1,1 +1,7 @@
-<%= button_to 'Run', run_task_path(task), method: :put, class: 'button is-success', disabled: task.deleted? %>
+<%= form_with url: run_task_path(task), method: :put do |form| %>
+  <%= render 'maintenance_tasks/tasks/upload_csv', form: form if task.csv_task? %>
+
+  <div class="block">
+    <%= form.submit 'Run', class: "button is-success", disabled: task.deleted? %>
+  </div>
+<% end %>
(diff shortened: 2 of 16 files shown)


app/views/maintenance_tasks/tasks/actions/_succeeded.html.erb
```diff
@@ -1 +1,7 @@
-<%= button_to 'Run', run_task_path(task), method: :put, class: 'button is-success', disabled: task.deleted? %>
+<%= form_with url: run_task_path(task), method: :put do |form| %>
+  <%= render 'maintenance_tasks/tasks/upload_csv', form: form if task.csv_task? %>
+
+  <div class="block">
+    <%= form.submit 'Run', class: "button is-success", disabled: task.deleted? %>
+  </div>
+<% end %>
```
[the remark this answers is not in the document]
adrianna-chang-shopify: Yes, definitely in favour of getting rid of this duplication 👍

app/controllers/maintenance_tasks/tasks_controller.rb
```diff
@@ -19,12 +19,17 @@ def index
     # Shows running and completed instances of the Task.
     def show
       @task = TaskData.find(params.fetch(:id))
+      @last_run = @task.last_run
+      set_refresh if @last_run&.active?
```
etiennebarrie, line 23: We already have memoization in `TaskData#last_run`, I don't think we need a instance variable for the last run.

app/views/maintenance_tasks/tasks/show.html.erb
```diff
@@ -4,10 +4,26 @@
...
+<%= render 'maintenance_tasks/runs/info', run: @last_run, with_status: false if @last_run %>
 
 <div class="buttons">
-  <%= render "maintenance_tasks/tasks/actions/#{@task.status}", task: @task %>
+  <% if @last_run.nil? || @last_run.completed? %>
+    <%= render 'maintenance_tasks/tasks/run_task', task: @task %>
+  <% elsif @last_run.cancelling? %>
+    <%= button_to 'Run', run_task_path(@task), method: :put, class: 'button is-success', disabled: true %>
+    <% if @last_run.stuck? %>
+      <%= button_to 'Cancel', cancel_task_run_path(@task, @last_run), method: :put, class: 'button is-danger', disabled: @task.deleted? %>
```
adrianna-chang-shopify, line 15: @etiennebarrie I noticed that this `Cancel` action is still disabled depending on whether the task is deleted or not... Which makes sense because if it's stuck, the job is actually already terminated, it's just the status that hasn't been updated to reflect that, so there's no concern about actually needing to cancel it.

I'm wondering though if it makes sense to have the button pop up at all if the Task is deleted.. Maybe we check that first:
```ruby
 <% if !@task.deleted? && @last_run.stuck? %>
   <%= button_to 'Cancel', cancel_task_run_path(@task, @last_run), method: :put, class: 'button is-danger' %>
<% end %>
```

(This is a pretty specific situation and it's quite possible I'm overthinking it, I was just bothered by the lack of symmetry between the `Cancel` buttons and disliked that this was the only one that was conditionally disabled 😛 )

WDYT?

app/views/maintenance_tasks/tasks/show.html.erb
```diff
@@ -4,10 +4,26 @@
   <%= @task %> <%= status_tag(@task.status) %>
 </h1>
 
-<%= render 'maintenance_tasks/runs/info', run: @task.last_run, with_status: false if @task.last_run %>
+<%= render 'maintenance_tasks/runs/info', run: @last_run, with_status: false if @last_run %>
 
 <div class="buttons">
-  <%= render "maintenance_tasks/tasks/actions/#{@task.status}", task: @task %>
+  <% if @last_run.nil? || @last_run.completed? %>
+    <%= render 'maintenance_tasks/tasks/run_task', task: @task %>
```
etiennebarrie, line 11: I would also inline this one personally.

etiennebarrie (approved):

app/views/maintenance_tasks/tasks/show.html.erb
etiennebarrie: I would definitely ignore for this PR. But I think we should definitely be able to do it, it was simply forgotten from #216.

adrianna-chang-shopify: I'm okay to ignore, wasn't sure if it was forgotten or intentional to disable this. It's such an odd edge case too, I'm not sure it's even worth addressing, was just curious what you thought 😁

PR #311, opened 2021-01-22
Set Task as errored if the Task doesn't define collection or process

### LINKED ISSUE #301: Missing collection method results in crash
Since we raise NotImplementedError, it's not rescued and surfaced to the user.

We may want to add a test in the test case template, not sure.

etiennebarrie:
Before this commit, we raised a `NotImplementedError` exception, which doesn't inherit from `StandardError` so is not caught by our `rescue_from`.

Even if we try to change our rescue_from to rescue from `NotImplementedError`, `ActiveJob::Execution#perform_now` doesn't rescue `Exception` like `ActionController::Rescue#process` does, so it doesn't catch the exception.

https://github.com/rails/rails/blob/v6.1.1/activejob/lib/active_job/execution.rb#L50
https://github.com/rails/rails/blob/v6.1.1/actionpack/lib/action_controller/metal/rescue.rb#L23

Close #301

### DIFF
--- a/app/tasks/maintenance_tasks/task.rb
+++ b/app/tasks/maintenance_tasks/task.rb
@@ -72,8 +72,7 @@ module MaintenanceTasks
     # @raise [NotImplementedError] with a message advising subclasses to
     #   implement an override for this method.
     def collection
-      raise NotImplementedError,
-        "#{self.class.name} must implement `collection`."
+      raise NoMethodError, "#{self.class.name} must implement `collection`."
     end
 
     # Placeholder method to raise in case a subclass fails to implement the
@@ -84,8 +83,7 @@ module MaintenanceTasks
     # @raise [NotImplementedError] with a message advising subclasses to
     #   implement an override for this method.
     def process(_item)
-      raise NotImplementedError,
-        "#{self.class.name} must implement `process`."
+      raise NoMethodError, "#{self.class.name} must implement `process`."
     end
 
     # Total count of iterations to be performed.
--- a/test/jobs/maintenance_tasks/task_job_test.rb
+++ b/test/jobs/maintenance_tasks/task_job_test.rb
@@ -233,6 +233,26 @@ module MaintenanceTasks
       assert_equal expected_message, @run.error_message
     end
 
+    test '.perform_now sets the Run as errored when the Task collection is not defined' do
+      collection_method = Maintenance::TestTask.instance_method(:collection)
+      Maintenance::TestTask.remove_method(:collection)
+      TaskJob.perform_now(@run)
+      @run.reload
+      assert_predicate(@run, :errored?)
+    ensure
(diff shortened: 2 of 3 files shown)


app/tasks/maintenance_tasks/task.rb
```diff
@@ -84,8 +83,7 @@ def collection
     # @raise [NotImplementedError] with a message advising subclasses to
     #   implement an override for this method.
     def process(_item)
-      raise NotImplementedError,
-        "#{self.class.name} must implement `process`."
+      raise "#{self.class.name} must implement `process`."
```
rafaelfranca, line 86: This is not a `RuntimeError`, so it seems wrong to raise that class to an error that is really a programming error. Would you raise an `RuntimeError` for a `SyntaxError`?

If we know that `NotImplementedError` can happen in our syste, why we don't rescue it in the `Job#process`?

rafaelfranca (approved):
I have my problems with raising a RuntimeError for a programming error and I think we don't need to do that but I'm ok with this as well.

adrianna-chang-shopify (approved):
LGTM!
@rafaelfranca we can change this back to a `NotImplementedError` once https://github.com/rails/rails/pull/41214 is released and we upgrade to the version of activejob with that patch if you'd prefer?

etiennebarrie:
> If we know that `NotImplementedError` can happen in our syste, why we don't rescue it in the `Job#process`?

I tried wrapping the the calls to `collection` and `process` instead: 9d8db9d1c73d19c3927c46bad574e338d2c5a683

That could work too, though the problem I see is that if someone uses NotImplementedError in their own code, and it gets called via Task#collection or Task#process, we would catch those too. So I also handled that in 584cbb2bb10be9a9adcbd7728f3abd04f99a8a90, the whole alternative implementation is in https://github.com/Shopify/maintenance_tasks/compare/missing-collection-wrap.

I think ultimately `NotImplementedError` is not well suited in this case, quoting the docs, it seems a bit like a stretch:

> Raised when a feature is not implemented on the current platform. For example, methods depending on the `fsync` or `fork` system calls may raise this exception if the underlying operating system or Ruby runtime does not support them.

It is very useful for things like `retry_on` because we raise at file loading time, so it does sound like a `ScriptError` with an adequate name in this case.

I agree `RuntimeError` is not a great fit. I thought about `ArgumentError`, in the sense that the `run` argument of `TaskJob` does no implement what it needs to, but then I realized when you pass an object that does not implement what it needs to you get a `NoMethodError`:

```ruby
puts(BasicObject.new)
# NoMethodError (undefined method `to_s' for #<BasicObject:0x00007f99e530ad00>)

Hash[BasicObject.new, true]
# NoMethodError (undefined method `hash' for #<BasicObject:0x00007f99e54e0f30>)

```

So I changed it to `NoMethodError`. That's what would happen in the first place if we didn't define it, which we only need as a placeholder for the documentation.

PR #313, opened 2021-01-25
Use superclass for CSV method

adrianna-chang-shopify:
Change the implementation of `CsvTask` to be a class from which application Tasks can inherit from, instead of a mixin.

It was unclear to me whether we're okay keeping the use of `respond_to?(csv_content=)` in `TaskJob` when setting `csv_content` on a Task:
https://github.com/Shopify/maintenance_tasks/blob/d5835468ea104dbe7c58fe4932c207229eaacb35/app/jobs/maintenance_tasks/task_job.rb#L59-L62

Or whether we wanted to introduce a class-level method like `csv_task?` instead. Let me know!

### DIFF
--- a/app/tasks/maintenance_tasks/task.rb
+++ b/app/tasks/maintenance_tasks/task.rb
@@ -27,7 +27,7 @@ module MaintenanceTasks
       # @return [Array<Class>] the list of classes.
       def available_tasks
         load_constants
-        descendants
+        descendants.without(CsvTask)
       end
 
       # Processes one item.
--- a/README.md
+++ b/README.md
@@ -123,16 +123,13 @@ Generate a CSV Task by running:
 $ rails generate maintenance_tasks:task import_posts --csv
 ```
 
-The generated task is a subclass of `MaintenanceTasks::Task` that includes a
-`MaintenanceTasks::CsvTask` module and implements:
+The generated task is a subclass of `MaintenanceTasks::CsvTask` that implements:
 * `process`: do the work of your maintenance task on a `CSV::Row`
 
 ```ruby
 # app/tasks/maintenance/import_posts_task.rb
 module Maintenance
-  class ImportPostsTask < MaintenanceTasks::Task
-    include MaintenanceTasks::CsvTask
-
+  class ImportPostsTask < MaintenanceTasks::CsvTask
     def process(row)
       Post.create!(title: row["title"], content: row["content"])
     end
--- a/app/tasks/maintenance_tasks/csv_task.rb
+++ b/app/tasks/maintenance_tasks/csv_task.rb
@@ -3,8 +3,9 @@
 require 'csv'
 
 module MaintenanceTasks
-  # A mixin that provides a Task with the ability to process CSV files.
-  module CsvTask
+  # Base class that is inherited by the host application's Task classes for
+  # processing CSV files.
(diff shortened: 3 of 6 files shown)


app/tasks/maintenance_tasks/task.rb
```diff
@@ -27,7 +39,7 @@ def named(name)
       # @return [Array<Class>] the list of classes.
       def available_tasks
         load_constants
-        descendants
+        descendants.reject(&:abstract_class?)
```
etiennebarrie, line 42: I think we could do without `abstract_class?` by doing this:
```suggestion
        descendants.without(CsvTask)
```

etiennebarrie (commented):
> Or whether we wanted to introduce a class-level method like csv_task? instead.

What I don't like about things like `csv_task?` is that now we have to define it in Task which should not care about CsvTask at all. I think `respond_to?` is fine.

adrianna-chang-shopify:
> What I don't like about things like csv_task? is that now we have to define it in Task which should not care about CsvTask at all

Yeah, that was my reasoning as well - I don't think the parent class should have to know anything about its subclasses and the specialized behaviour they provide. I'm keen on sticking to our use of `respond_to?`

app/tasks/maintenance_tasks/task.rb
adrianna-chang-shopify: I's suspecting that using a superclass might be the best option for implementing custom iteration Tasks (see https://github.com/Shopify/maintenance_tasks/pull/310#issuecomment-766843596), which prompted me to keep `abstract_class?` in here. But we can always go with the simplest option for now, and only implement `abstract_class`  once we actually have another class that needs to be abstract 😁

sambostock:
If I'm not mistaken, `.respond_to?(:csv_content=)`/`.csv_task?` are used to trigger loading of the CSV contents in `TaskJob`, right?

```ruby
if @task.respond_to?(:csv_content=)
  @task.csv_content = @run.csv_file.download
end
```

If that's it, would it make sense to define a generic hook/callback method on `Task`, which `CsvTask` overrides. That way you can invert things and make it so the `TaskJob` doesn't know anything about CSVs

```ruby
# task.rb
class Task
  def before_perform(_run)
  end
end

# csv_task.rb
class CsvTask < Task
  def before_perform(run)
    self.csv_content = run.csv_file.download
    super
  end
end
```

This could alternatively be implemented as an ActiveSupport::Callback.

etiennebarrie:
> would it make sense to define a generic hook/callback method on `Task`, which `CsvTask` overrides

It could make sense, but then we would give too much API to Task writers, which don't want to. We provide CsvTask and TaskJob, so I think even if not ideal, it's fine to have some knowledge between them.
Whereas for all the tasks that the users write, we don't want to give them control over the Run object, it's an implementation detail they shouldn't have to worry about.

etiennebarrie (approved):

adrianna-chang-shopify:
Providing a callback on Task is pretty neat! But yeah, as Étienne mentioned, the tradeoff is making Task aware of Run, which we want to avoid. It's not the prettiest to use `respond_to?`, I agree, but since `TaskJob` already knows how to talk to `Run` and the `Task` instance it's creating, I believe it's the favourable to having Tasks be aware of implementation details.

PR #315, opened 2021-01-27
Move non-models from app/models to lib

sambostock:
These helper objects are not `ActiveRecord` models, and (IMO) make more sense in the `lib` directory.

@adrianna-chang-shopify and I had discussed this, and I figured the easiest thing was to put up the PR and see what the reception is. Glad to close this if it doesn't make sense.


lib/maintenance_tasks.rb
```diff
@@ -9,6 +9,11 @@
 require 'pagy'
 require 'pagy/extras/bulma'
 
+require 'maintenance_tasks/progress'
+require 'maintenance_tasks/runner'
+require 'maintenance_tasks/task_data'
+require 'maintenance_tasks/ticker'
```
sambostock, line 15: I'm guessing these were in `app/models` to avoid having to explicitly `require` them?

rafaelfranca:
The models folder is not only for active record objects. Is for domain model classes, and the classes you are moving out are exactly that

sambostock:
Fair enough 👍 

Thanks for the lightning fast review!

sambostock:
For posterity, @rafaelfranca and I continued discussing this over Slack, and this [screencast by DHH](https://www.youtube.com/watch?v=hkmrfjex7jI) which discusses the topic of the `app/models` directory came up.

Issue #314, opened 2021-01-26
Stop making all constants private


adrianna-chang-shopify:
Rafael mentioned that we should not prevent people from extending our classes if they are willing to accept the risks involved. 

A recent example involved a team needing to patch the gem's `ApplicationController` in order to specify a `before_action` callback that would provide user authorization.

We should remove the `private_constant` declarations on our classes and the corresponding [test](https://github.com/Shopify/maintenance_tasks/blob/main/test/lib/maintenance_tasks_test.rb).

etiennebarrie:
We don't really prevent people from extending the classes though, they can still do it, it's just more clear they're doing something that is not supported and that can break easily.

I'm all for supporting specific uses cases, for example regarding the authorization, I think if hooking ApplicationController is what's needed, we could definitely make it public, and maybe point out to people in the README what they can do.

Or maybe it's too high level, if making decisions based on `controller_name` and `action_name` isn't ideal (that's what's going to happen if you hook into ApplicationController), we could implement some kinds of hooks specific to what's needed in that use case (authorization on starting/pausing/cancelling a run). Because even if you prevent starting a run with a controller hook, the button would still be there, which isn't great.

adrianna-chang-shopify:
> I'm all for supporting specific uses cases

Agree 100% that if there is an obvious use case then users should be able to use a documented, public API to implement what they need. Auth is a good example, we had talked about building that right into the gem, but exposing ApplicationController and providing documentation on how to patch it to set filters for authentication is good enough.

> We don't really prevent people from extending the classes though, they can still do it, it's just more clear they're doing something that is not supported and that can break easily.

Right, but they have to use `const_get` which is a pain. @rafaelfranca had mentioned
> It is always better to allow people to extend classes if they are willing to accept the risks

I was wondering if we could rely entirely on `@api private` to denote parts of the library that shouldn't be extended as it's risky, but interestingly YARD actually says [this](https://rubydoc.info/gems/yard/file/docs/Tags.md#private) about the private tag:
> This tag should only be used to mark objects private when Ruby visibility rules cannot do so. In Ruby 1.9.3, you can use private_constant to declare constants (like classes or modules) as private, and should be used instead of @api private

So 🤷‍♀️  Maybe we just need to be more open to making things in the gem public if users come forward with potential use cases where they'd like to extend

etiennebarrie:
[...]

PR #317, opened 2021-01-28
Use a class method to declare a CSV task

etiennebarrie:
In #302 we went with a new subclass, which we refactored to a module to be included in a Task.

We got feedback in https://github.com/Shopify/maintenance_tasks/pull/302#discussion_r562911483 but we can't really change the behavior only based on data from the Run (a CSV file being present), since we need to know something about the class to show the input which results in a CSV file being updated. The class itself needs to somehow declare that it supports CSV.

Rafael pointed out that we shouldn't use `include` as an API for changing behavior to the classes, we discussed that on Slack but we missed the mark with #313 when we went instead with a superclass, as explained in https://github.com/Shopify/maintenance_tasks/pull/310#issuecomment-768529654.

So instead this PR uses a class method on Task, doesn't require the Task authors to include a module or inherit from a different class. They simply call `csv_collection`. I'm open to feedback about the name, but the idea with this one is that it defines a `collection` like a Task author would do for you, so there some parallel between defining a method `collection` and calling `csv_collection` on the class.

Using a class method opens up features like being able to choose whether the CSV should have headers, changing the column separator, using converters, etc. as arguments to the method call that could be passed in to [CSV.new](https://ruby-doc.org/stdlib-2.7.2/libdoc/csv/rdoc/CSV.html#method-c-new).

Internally this still includes a module into the class rather than adding conditionals to `collection` and `count` and adding the `csv_content` attribute to the receiver.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,13 +123,16 @@ Generate a CSV Task by running:
 $ rails generate maintenance_tasks:task import_posts --csv
 ```
 
-The generated task is a subclass of `MaintenanceTasks::CsvTask` that implements:
+The generated task is a subclass of `MaintenanceTasks::Task` that implements:
+
 * `process`: do the work of your maintenance task on a `CSV::Row`
 
 ```ruby
 # app/tasks/maintenance/import_posts_task.rb
 module Maintenance
-  class ImportPostsTask < MaintenanceTasks::CsvTask
+  class ImportPostsTask < MaintenanceTasks::Task
+    csv_collection
+
     def process(row)
       Post.create!(title: row["title"], content: row["content"])
     end
--- a/app/models/maintenance_tasks/csv_collection.rb
+++ b/app/models/maintenance_tasks/csv_collection.rb
@@ -3,9 +3,11 @@
 require 'csv'
 
 module MaintenanceTasks
-  # Base class that is inherited by the host application's Task classes for
+  # Module that is included into Task classes by Task.csv_collection for
   # processing CSV files.
-  class CsvTask < Task
+  #
+  # @api private
+  module CsvCollection
     # The contents of a CSV file to be processed by a Task.
     #
     # @return [String] the content of the CSV file to process.
@@ -28,4 +30,5 @@ module MaintenanceTasks
       csv_content.count("\n") - 1
     end
   end
+  private_constant :CsvCollection
 end
--- a/app/models/maintenance_tasks/task_data.rb
+++ b/app/models/maintenance_tasks/task_data.rb
@@ -129,7 +129,7 @@ module MaintenanceTasks
 
     # @return [Boolean] whether the Task inherits from CsvTask.
     def csv_task?
-      !deleted? && Task.named(name) < CsvTask
+      !deleted? && Task.named(name) < CsvCollection
     end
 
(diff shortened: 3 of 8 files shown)


app/tasks/maintenance_tasks/csv_task.rb
```diff
@@ -3,9 +3,9 @@
 require 'csv'
 
 module MaintenanceTasks
-  # Base class that is inherited by the host application's Task classes for
+  # Module that is included into Task classes by Task#csv_collection for
   # processing CSV files.
-  class CsvTask < Task
+  module CsvTask
```
sambostock, line 8: Would a name like `CsvHandling` be more appropriate, since on its own this module isn't a `Task`?

sambostock, line 8: Should this be annotated as a private API (on top of `private_constant`)?
```suggestion
  # @api private
  module CsvTask
```

adrianna-chang-shopify (approved):

app/tasks/maintenance_tasks/csv_task.rb
adrianna-chang-shopify: Yeah, or `CsvProcessing` to make use of the language we've already established with `Task#process`?

etiennebarrie: `CsvProcessing` is weird because it's actually not doing the processing, but everything else 😄 

What about `CsvCollection` just like the `csv_collection` method?

adrianna-chang-shopify: I'm not sure this module is more deserving of being denoted as private API compared to some of our other internals. However, we are planning on actually [dropping the private_constant declarations](https://github.com/Shopify/maintenance_tasks/issues/314) so that we no longer prevent users from extending our classes if they choose to. Consequently, we may want to switch to using YARD's `@api private` tag in the future anyways. But I think I'd leave this for now 😄

adrianna-chang-shopify: LOL true. `CsvCollection` actually came to my mind as well, I like that 👍

etiennebarrie: I think all the private constants should be `@api private`, I fixed this one, there's a couple of others missing, I'll add them to a PR turning on documentation generation I think.

etiennebarrie:
One last thing I'm not sure about is whether to make the method private or not. On one hand it doesn't make sense to be called from outside the class, on the other hand things like `attr_reader`/`belongs_to`/`before_action` are not private and it's fine.

app/tasks/maintenance_tasks/csv_task.rb
adrianna-chang-shopify: Yeah sorry, I didn't even realize most of our constants were already denoted as `@api private` 🤦‍♀️

adrianna-chang-shopify:
> One last thing I'm not sure about is whether to make the method private or not. 

I would be inclined to keep it public, given that most class-level macros are given public visibility in Rails.

rafaelfranca (approved):

PR #318, opened 2021-01-29
Allow CSV files for previous runs to be downloaded via the UI

adrianna-chang-shopify:
This PR leverages [ Active Storage's download link helper](https://guides.rubyonrails.org/active_storage_overview.html#linking-to-files) to provide download links for CSV attachments associated with previous runs for CSV Tasks. Note that we have to use `Rails.application.routes.url_helpers.rails_blob_path` instead of `rails_blob_path` because the download URL is associated with the host app, not the engine (and additionally, we don't specify Active Storage as a dependency of the engine, so this helper isn't even available to us).

Since we're using `run.csv_file.attached?` to check whether to show the download link, I've added preloading for the CSV file attachments when we generate the `ActiveRecord::Relation` for Runs, as @etiennebarrie suggested. (And we can use `with_attached_csv_file`, since [this scope](https://github.com/rails/rails/blob/f815e773412f1064492507d76a9a011e6dde6042/activestorage/lib/active_storage/attached/macros.rb#L44) ships out-of-the-box with ActiveStorage)

To 🎩 
- Go to `http://localhost:3000/maintenance_tasks/tasks/Maintenance::ImportPostsTask/`
- Upload the `sample.csv` from `test/fixtures/files`
- Click the `Download CSV` link that appears when the Task finishes

### DIFF
--- a/app/controllers/maintenance_tasks/runs_controller.rb
+++ b/app/controllers/maintenance_tasks/runs_controller.rb
@@ -10,7 +10,7 @@ module MaintenanceTasks
 
     # Shows a full list of Runs.
     def index
-      query = Run.all.order(id: :desc)
+      query = Run.with_attached_csv_file.all.order(id: :desc)
       if params[:task_name].present?
         task_name = Run.sanitize_sql_like(params[:task_name])
         query = query.where('task_name LIKE ?', "%#{task_name}%")
--- a/app/helpers/maintenance_tasks/task_helper.rb
+++ b/app/helpers/maintenance_tasks/task_helper.rb
@@ -105,6 +105,14 @@ module MaintenanceTasks
       end
       safe_join(tokens)
     end
+
+    # Returns a download link for a Run's CSV attachment
+    def csv_file_download_path(run)
+      Rails.application.routes.url_helpers.rails_blob_path(
+        run.csv_file,
+        only_path: true
+      )
+    end
   end
   private_constant :TaskHelper
 end
--- a/app/models/maintenance_tasks/task_data.rb
+++ b/app/models/maintenance_tasks/task_data.rb
@@ -135,7 +135,7 @@ module MaintenanceTasks
     private
 
     def runs
-      Run.where(task_name: name).order(created_at: :desc)
+      Run.where(task_name: name).with_attached_csv_file.order(created_at: :desc)
     end
   end
   private_constant :TaskData
--- a/app/views/maintenance_tasks/runs/_info.html.erb
+++ b/app/views/maintenance_tasks/runs/_info.html.erb
@@ -8,3 +8,9 @@
 <div class="content">
   <%= render "maintenance_tasks/runs/info/#{run.status}", run: run %>
 </div>
+
+<% if run.csv_file.attached? %>
+  <div class="block">
+    <%= link_to('Download CSV', csv_file_download_path(run)) %>
+  </div>
+<% end %>
--- a/test/application_system_test_case.rb
+++ b/test/application_system_test_case.rb
@@ -11,6 +11,10 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
 
   driven_by :selenium, using: :headless_chrome do |options|
     options.add_argument('--disable-dev-shm-usage')
+    options.add_preference(
+      :download,
+      default_directory: 'test/dummy/tmp/downloads'
+    )
   end
 
   setup do
@@ -21,5 +25,6 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
   teardown do
     assert_empty page.driver.browser.manage.logs.get(:browser)
     Maintenance::UpdatePostsTask.fast_task = true
+    FileUtils.rm_rf('test/dummy/tmp/downloads')
   end
 end
--- a/test/helpers/maintenance_tasks/task_helper_test.rb
+++ b/test/helpers/maintenance_tasks/task_helper_test.rb
@@ -84,5 +84,14 @@ module MaintenanceTasks
         '<span class="ruby-int">2</span>', highlight_code('1 2')
       assert_equal "\n", highlight_code("\n")
     end
+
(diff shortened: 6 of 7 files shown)


app/views/maintenance_tasks/runs/_run.html.erb
```diff
@@ -8,4 +8,7 @@
   <% else %>
     <%= render 'maintenance_tasks/runs/info', run: run, with_status: true %>
   <% end %>
+  <% if defined?(task) && task.csv_task? %>
```
adrianna-chang-shopify, line 11: This is not so nice, but is hard to achieve otherwise given that this partial is used to show the list of previous runs, but also to power the Runs index page (in which Runs are not scoped to Tasks). This will hide the download links from the Runs index page (based on https://github.com/Shopify/maintenance_tasks/pull/288 I'm not sure we're intending to keep this anymore?). If we wanted to show this link on the index page as well, I think we could modify it to be:

```ruby
 <% if defined?(task) && task.csv_task? || TaskData.find(run.task_name).csv_task? %>
```

But not sure I like that...

adrianna-chang-shopify, line 11: This is not so nice, but a result of the fact that this partial is shared between the previous runs section on the show page, and the runs index page (in which Runs are not scoped to a Task). It seems like from https://github.com/Shopify/maintenance_tasks/pull/288 we might end up getting rid of this anyways, which would simplify our lives here because we could rely on `task` being defined.

adrianna-chang-shopify: Alternatively, we could map runs to TaskData objects in `RunsController#index`, but I think that defeats the purpose of having the separate page and adds extra complexity I'm not sure we want... But, being able to download CSVs for deleted Tasks was on our list of advantages for keeping the full list of runs, s maybe we do want it.

app/controllers/maintenance_tasks/runs_controller.rb
```diff
@@ -34,6 +34,19 @@ def cancel
       redirect_to(task_path(@run.task_name), alert: error.message)
     end
 
+    # Downloads the CSV attachment associated with the Run.
+    def csv_file
+      respond_to do |format|
+        format.csv do
+          csv_file = @run.csv_file.download
+          task_name = @run.task_name.underscore.gsub('/', '_')
```
etiennebarrie, line 42: Why not keep the task name as is? It looks like Firefox removes the `::` and puts a space, while Safari and Chrome use `__`.

app/controllers/maintenance_tasks/runs_controller.rb
```diff
@@ -34,6 +34,19 @@ def cancel
...
+          timestamp = @run.created_at.strftime('%Y%m%d%H%M%S')
+
+          send_data(csv_file, filename: "#{task_name}_#{timestamp}.csv")
```
etiennebarrie, line 45: Maybe we could put the timestamp first for better ordering? OTOH this sorts by task then date, it's pretty cool too.
Why the underscore? spaces are fine in file names, no?

Also maybe we could just use the name of the file that was sent to us to make things simpler if a bit less useful?

app/views/maintenance_tasks/runs/_run.html.erb
etiennebarrie: Why not check whether the run itself has an attachment instead of looking up the Task? Something like `run.csv_file.attached?`. We may need to preload that relation though.

app/views/maintenance_tasks/runs/_run.html.erb
```diff
@@ -8,4 +8,7 @@
...
+    <%= link_to('Download CSV', csv_file_task_run_path(task, run)) %>
```
etiennebarrie, line 12: It seems like there's a way to directly link to an attachment without writing a controller action: https://guides.rubyonrails.org/active_storage_overview.html#linking-to-files
Is this something that didn't work in our case?

app/views/maintenance_tasks/runs/_run.html.erb
adrianna-chang-shopify: Oh yeah good call, I thought we'd need to download and check if it was nil, but we can definitely use `attached?` instead

app/controllers/maintenance_tasks/runs_controller.rb
adrianna-chang-shopify: Timestamp vs task name first / underscores vs spaces are not things I feel strongly about, so happy to change any of these if you'd like.

I'm torn on using the original file name. It simplifies things for sure, but OTOH it's kind of nice to have the task + date in the file name, especially if you're downloading multiple CSVs for a Task.

app/views/maintenance_tasks/runs/_run.html.erb
adrianna-chang-shopify: Yeah, great question - I mostly just wanted more flexibility around naming the file 😄  But perhaps that's an embellishment we don't need, we can just have it use the original file name...

adrianna-chang-shopify:
Okay, I've changed it to use the ActiveStorage url helper and have gotten rid of the new action in the Runs Controller. My system test is failing now though, will have to investigate what's up next week 🤔  Everything is working when I 🎩  in dev

app/views/maintenance_tasks/tasks/show.html.erb
```diff
@@ -4,7 +4,14 @@
   <%= @task %> <%= status_tag(@task.status) %>
 </h1>
 
-<%= render 'maintenance_tasks/runs/info', run: @task.last_run, with_status: false if @task.last_run %>
+<% if last_run = @task.last_run %>
+  <%= render 'maintenance_tasks/runs/info', run: last_run, with_status: false %>
```
etiennebarrie, line 8: Do you think the link could go inside the `maintenance_tasks/runs/info` partial instead?

etiennebarrie (approved):
It was indeed neat to have a nicer file name. Not sure it's worth the added complexity though.

Issue #309, opened 2021-01-22
Pass task details to error handler


timgladwell:
Enhance the existing `maintenance_tasks` error hook 👇  to add context information about the task that caused the error

https://github.com/Shopify/maintenance_tasks/blob/7131371ad6e7cf0604ad54f27b6a01b7914dea56/app/jobs/maintenance_tasks/task_job.rb#L100

Note: host system can capture additional details about the background job itself with a setup similar to 
```ruby
        Bugsnag.notify(exception) do |notification|
          notification.severity = 'warning'
          notification.add_tab(:sidekiq, job_context)
        end
```
Which produces a [bugsnag report](https://app.bugsnag.com/shopify/shipify/errors/6008a2eb31660f001796913b) similar to 
![image](https://user-images.githubusercontent.com/17774743/105541514-dc580200-5cc5-11eb-999b-9ec2adb2bcb8.png)
so a full-featured background job report may not be necessary, just the maintenance task details.

PR #321, opened 2021-02-01
Supply additional task info as argument when calling error handler

adrianna-chang-shopify:
Closes: https://github.com/Shopify/maintenance_tasks/issues/309

This PR changes the call to the defined error handler to pass along an additional `task_context` argument. This arg is a hash containing some additional information about the Task and the corresponding run.

An example of how users can leverage this info has been added to the `README`: if they integrate with Bugsnag, the simplest thing to do would just be to add a tab containing the details of the Task's context when notifying.

We might want to think a bit more about what information it makes sense to expose. Assuming that most users are using Sidekiq for their queuing backend, a lot of the information pertaining to the job is exposed through the `job_context` (as shown in the original issue) - `job_id`, `cursor`, `job_class`, etc... I thought it might make sense to expose `started_at` and `ended_at` (although `enqueued_at` is available from the job context). Alternatively, we could expose as much information about the Task as possible (include `job_id`, `cursor`, `time_running`, etc...) under the assumption that not everyone will use Sidekiq and have access to all this additional information to augment their error notifier.

cc @timgladwell

### DIFF
--- a/README.md
+++ b/README.md
@@ -319,8 +319,30 @@ you can define an error handler:
 
 ```ruby
 # config/initializers/maintenance_tasks.rb
-MaintenanceTasks.error_handler = ->(error) { Bugsnag.notify(error) }
+MaintenanceTasks.error_handler = ->(error, task_context, _errored_element) do
+  Bugsnag.notify(error) do |notification|
+    notification.add_tab(:task, task_context)
+  end
+end
 ```
+
+The error handler should be a lambda that accepts three arguments:
+
+* `error`: The object containing the exception that was raised.
+* `task_context`: A hash with additional information about the Task and the
+  error:
+  * `task_name`: The name of the Task that errored
+  * `started_at`: The time the Task started
+  * `ended_at`: The time the Task errored
+* `errored_element`: The element, if any, that was being processed when the
+  Task raised an exception. If you would like to pass this object to your
+  exception monitoring service, make sure you **sanitize the object** to avoid
+  leaking sensitive data and **convert it to a format** that is compatible with
+  your bug tracker. For example, Bugsnag only sends the id and class name of
+  ActiveRecord objects in order to protect sensitive data. CSV rows, on the
+  other hand, are converted to strings and passed raw to Bugsnag, so make sure
+  to filter any personal data from these objects before adding them to a
+  report.
 
 #### Customizing the maintenance tasks module
 
--- a/app/jobs/maintenance_tasks/task_job.rb
+++ b/app/jobs/maintenance_tasks/task_job.rb
@@ -49,9 +49,16 @@ module MaintenanceTasks
     # @param _run [Run] the current Run, passed as an argument by Job Iteration.
     def each_iteration(input, _run)
       throw(:abort, :skip_complete_callbacks) if @run.stopping?
-      @task.process(input)
+      task_iteration(input)
       @ticker.tick
       @run.reload_status
+    end
+
+    def task_iteration(input)
+      @task.process(input)
+    rescue => error
+      @errored_element = input
+      raise error
     end
 
     def before_perform
@@ -97,7 +104,14 @@ module MaintenanceTasks
     def on_error(error)
       @ticker.persist if defined?(@ticker)
       @run.persist_error(error)
-      MaintenanceTasks.error_handler.call(error)
+
+      task_context = {
+        task_name: @run.task_name,
+        started_at: @run.started_at,
+        ended_at: @run.ended_at,
+      }
+      errored_element = @errored_element if defined?(@errored_element)
+      MaintenanceTasks.error_handler.call(error, task_context, errored_element)
     end
   end
 end
--- a/lib/maintenance_tasks.rb
+++ b/lib/maintenance_tasks.rb
@@ -32,6 +32,23 @@ module MaintenanceTasks
   #   the ticker during Task iterations.
   mattr_accessor :ticker_delay, default: 1.second
 
+  # Retrieves the callback to be performed when an error occurs in the task.
+  def self.error_handler
+    return @error_handler if defined?(@error_handler)
+    @error_handler = ->(_error, _task_context, _errored_element) {}
+  end
+
   # Defines a callback to be performed when an error occurs in the task.
-  mattr_accessor :error_handler, default: ->(_error) {}
+  def self.error_handler=(error_handler)
+    unless error_handler.arity == 3
+      ActiveSupport::Deprecation.warn(
+        'MaintenanceTasks.error_handler should be a lambda that takes three '\
+         'arguments: error, task_context, and errored_element.'
+      )
+      @error_handler = ->(error, _task_context, _errored_element) do
+        error_handler.call(error)
+      end
+    end
+    @error_handler = error_handler
+  end
 end
--- a/test/jobs/maintenance_tasks/task_job_test.rb
+++ b/test/jobs/maintenance_tasks/task_job_test.rb
@@ -262,19 +262,31 @@ module MaintenanceTasks
     test '.perform_now calls the error handler when there was an Error' do
       error_handler_before = MaintenanceTasks.error_handler
       handled_error = nil
-      MaintenanceTasks.error_handler = ->(error) { handled_error = error }
+      handled_task_context = nil
+      handled_errored_element = nil
+
+      MaintenanceTasks.error_handler = ->(error, task_context, errored_el) do
+        handled_error = error
+        handled_task_context = task_context
+        handled_errored_element = errored_el
+      end
+
       run = Run.create!(task_name: 'Maintenance::ErrorTask')
 
       TaskJob.perform_now(run)
 
       assert_equal(ArgumentError, handled_error.class)
+      assert_equal('Maintenance::ErrorTask', handled_task_context[:task_name])
+      assert_equal(2, handled_errored_element)
     ensure
       MaintenanceTasks.error_handler = error_handler_before
     end
 
     test '.perform_now still persists the error properly if the error handler raises' do
       error_handler_before = MaintenanceTasks.error_handler
-      MaintenanceTasks.error_handler = ->(error) { raise error }
+      MaintenanceTasks.error_handler = ->(error, _task_context, _errored_el) do
+        raise error
+      end
       run = Run.create!(task_name: 'Maintenance::ErrorTask')
 
       assert_raises { TaskJob.perform_now(run) }
--- a/test/lib/maintenance_tasks_test.rb
+++ b/test/lib/maintenance_tasks_test.rb
@@ -19,4 +19,14 @@ class MaintenanceTasksTest < ActiveSupport::TestCase
     end
     assert_equal expected_public_constants.sort, public_constants.sort
   end
+
+  test 'deprecation warning raised when error_handler does not accept three arguments' do
+    error_handler_before = MaintenanceTasks.error_handler
+
+    dep_msg = 'MaintenanceTasks.error_handler should be a lambda that takes '\
+    'three arguments: error, task_context, and errored_element.'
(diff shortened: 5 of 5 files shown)


timgladwell:
Thanks @adrianna-chang-shopify!

Would it also be possible to include some kind of detail about the specific entity being processed when the error occurred? I'm thinking that would make it easier to track down and identify the cause of the error.

adrianna-chang-shopify:
Hey @timgladwell , that should be doable! I'll make the adjustment to this PR

app/jobs/maintenance_tasks/task_job.rb
```diff
@@ -97,7 +97,12 @@ def after_perform
     def on_error(error)
       @ticker.persist if defined?(@ticker)
       @run.persist_error(error)
-      MaintenanceTasks.error_handler.call(error)
+      task_context = {
+        task_name: @run.task_name,
+        started_at: @run.started_at,
+        ended_at: @run.ended_at,
+      }
+      MaintenanceTasks.error_handler.call(task_context, error)
```
etiennebarrie, line 105: We need to keep backwards compatibility for existing users who have already configured their error handler. I think we can add an arity check here.

adrianna-chang-shopify: Thanks, I forgot that would be an issue ❤️

adrianna-chang-shopify:
@etiennebarrie I've made the following changes:
* Modified https://github.com/Shopify/maintenance_tasks/pull/321/commits/6f246d07e468bc941ae9feccf24515d14a4eea24 to change the order of the arguments to `error_handler.call`, and to add the arity check
* Added https://github.com/Shopify/maintenance_tasks/pull/321/commits/8f72aeda327141205c84698824570ef4d2ff86eb to introduce a deprecation warning for the old error handler signature. I think it's worth having people move over to this new expected API to eventually be able to remove that extra arity check in `TaskJob`.
* Added https://github.com/Shopify/maintenance_tasks/pull/321/commits/15b44440481ec08073b8701ce963967b321a815d to accomplish what @timgladwell requested. I opted to set it as an ivar on the `TaskJob` instance so we can pass it as part of the `task_context`, but I've been debating whether this is something we'd actually like to expose in the UI as well (and consequently persist to the `@run`.) I think it could bring a lot of value in terms of helping users figure out why something happened with their Task, but does involve a schema change... WDYT?

etiennebarrie (commented):
Regarding the processed element which caused the error:
* we have the cursor_position, which we could save when we have an error and it would have an up-to-date value of the cursor, pointing to the element (meaning if we rebuild the enumerator, and take the first element, it should be the one that caused the error), so I don't think we need to persist the element, but we may want to properly save the cursor_position upon error.
* the PII question worries me a bit, and I'm thinking as a result it should probably be a third parameter to the error handler so that application owners are more explicitly handling it. But then the processed element type will depend on each task, so it may seem natural to call back the task so that it knows what to do with it, but that adds to the Task API. Maybe we could let app developers deal with it though, and not prescribe anything, e.g.:

```ruby
MaintenanceTasks.error_handler = ->(error, context, element) do
  # they could go full in, calling back the Task
  context[:element] = Task.named(context[:task_name]).sanitize(element)
  # or just generic
  context[:element] = "#{element.class} ##{element.id}" if element.is_a?(ActiveRecord::Base)
  context[:element] = row.to_s.hash if element.is_a?(CSV::Row)

  # Bugsnag.notify ...
end
```

Ultimately, the cursor gives developers the way to dive deeper without taking any risk about sending PII to the bug tracker. The problem is that we've hidden the cursor from the API so well that they may not know what to do with it.

README.md
```diff
@@ -316,7 +316,11 @@ you can define an error handler:
 
 ```ruby
-MaintenanceTasks.error_handler = ->(error) { Bugsnag.notify(error) }
+MaintenanceTasks.error_handler = lambda do |error, task_context|
```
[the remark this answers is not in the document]
adrianna-chang-shopify: Thanks, thought `lambda` was preferred for multilines! 👍

app/jobs/maintenance_tasks/task_job.rb
```diff
@@ -96,8 +101,25 @@ def after_perform
 
     def on_error(error)
       @ticker.persist if defined?(@ticker)
+      element = defined?(@processed_element) ? @processed_element : nil
       @run.persist_error(error)
-      MaintenanceTasks.error_handler.call(error)
+
+      task_context = {
+        task_name: @run.task_name,
+        processed_element: element,
```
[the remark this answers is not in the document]
adrianna-chang-shopify: Yeah, good question! Bugsnag does indeed seem to handle ActiveRecords well enough (I'm seeing `"processed_element":"[OBJECT]: [Class]: Post [ID]: 69"` which isn't the greatest, you at least you have the model + ID). 

I would be inclined to take a stance that is more along the lines of "it's up to the users to handle the data the way they need to". ie. If they need more info about the `processed_element` then what Bugsnag would provide if they sent the AR directly to the bug tracker, then they need to unpack it. Similarly, if they are concerned about scrubbing data for PII, then they need to do that in their error handler prior to sending the notification to Bugsnag.

lib/maintenance_tasks/engine.rb
```diff
@@ -17,6 +17,12 @@ class Engine < ::Rails::Engine
     end
 
     config.after_initialize do
+      if MaintenanceTasks.error_handler.arity.equal?(1)
```
adrianna-chang-shopify, line 20: I've moved the deprecation check into the `after_initialize`  block in the engine. Seems like a sensible place to put it?

README.md
```diff
@@ -316,7 +316,36 @@ you can define an error handler:
...
+**convert it to a format that is compatible with your bug tracker**.
+
+For example:
+```ruby
+# config/initializers/maintenance_tasks.rb
+MaintenanceTasks.error_handler = ->(error, task_context) do
+  # task_context[:errored_element] is
+  # <CSV::Row "name":"John Doe" "address":"123 Main Street">
+  errored_element = task_context.delete(:errored_element).to_h
+  errored_element['address'] = 'REDACTED'
```
etiennebarrie, line 341: Given the example above (the safe default that people can copy paste without spending more time on it), I'm not sure we need to want to have another one.

---

If we want another example, I would go ahead and point out the more Railsy way to do this with a parameter filter, and since we've established [Bugsnag only keeps the id](https://github.com/bugsnag/bugsnag-ruby/blob/f9c539670c448f7f129a3f8be7d412e2e824a357/lib/bugsnag/cleaner.rb#L126-L127).

```ruby
parameter_filter = ActiveSupport::ParameterFilter.new(Rails.application.config.filter_parameters)
MaintenanceTasks.error_handler = ->(error, task_context) do
  errored_element = task_context.delete(:errored_element)
  errored_element = if errored_element.respond_to?(:to_h)
    parameter_filter.filter(errored_element.to_h)
  elsif errored_element.is_a?(ActiveRecord::Base)
    errored_element # bugsnag only sends the id
  end
  task_context[:errored_element] = errored_element
  # ...
end
```

lib/maintenance_tasks/engine.rb
etiennebarrie: I think I would have overridden `error_handler=` to do that, that way we can't miss it.

etiennebarrie (commented):
I know we try to be minimalist for this project, so going with 2 parameters to the handler makes sense, but this is a situation where a block is the perfect tool since it doesn't care about the arity, so my preferred API for this would be:

```ruby
MaintenanceTasks.error_handler { |error| Bugsnag.notify(error) }
MaintenanceTasks.error_handler { |error, context|
  Bugsnag.notify(error) { _1.add_tab(:task, context) } }
MaintenanceTasks.error_handler { |error, context, element|
  Bugsnag.notify(error) { _1.add_tab(:task, context.merge(element: element)) } }
```

That way we give the pieces to our users and they can handler them the way they want. And the documentation could give as an example the second case. That way people copy/pasting would not get into any risk of sending PII.

adrianna-chang-shopify:
Fair enough, I'm okay to go with a block! 👍  I thought maybe it would be less confusing to enforce an error handler with a specific signature to ensure consistency and to make users aware of all the possible info that comes in when an error triggers a callback, but I think documenting the different use cases in the README is sufficient.

I'll get this updated!

lib/maintenance_tasks/engine.rb
adrianna-chang-shopify: AH, sorry misunderstood what you meant initially. I'll update this to raise on the setter

lib/maintenance_tasks.rb
```diff
@@ -32,6 +32,18 @@ module MaintenanceTasks
...
   # Defines a callback to be performed when an error occurs in the task.
-  mattr_accessor :error_handler, default: ->(_error) {}
+  def self.error_handler=(error_handler)
+    unless error_handler.arity.equal?(3)
+      ActiveSupport::Deprecation.warn(
+        'MaintenanceTasks.error_handler be an object that responds to .call '\
+         'and takes three arguments: error, task_context, and errored_element.'
+      )
+    end
+    @@error_handler = error_handler # rubocop:disable Style/ClassVars
```
etiennebarrie, line 47: `MaintenanceTasks` is not going to be inherited, so we could be using standard instance variables for the Module instance. That would require changing the reader but would avoid the rubocop disable here.

app/jobs/maintenance_tasks/task_job.rb
```diff
@@ -49,7 +49,12 @@ def build_enumerator(_run, cursor:)
     # @param _run [Run] the current Run, passed as an argument by Job Iteration.
     def each_iteration(input, _run)
       throw(:abort, :skip_complete_callbacks) if @run.stopping?
-      @task.process(input)
+      begin
```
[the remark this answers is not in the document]
adrianna-chang-shopify: I'll extract a method called `task_iteration` 👍

app/jobs/maintenance_tasks/task_job.rb
```diff
@@ -97,7 +102,23 @@ def after_perform
...
-      MaintenanceTasks.error_handler.call(error)
+
+      task_context = {
+        task_name: @run.task_name,
+        started_at: @run.started_at,
+        ended_at: @run.ended_at,
+      }
+      errored_element = @errored_element if defined?(@errored_element)
+
+      if MaintenanceTasks.error_handler.arity == 1
```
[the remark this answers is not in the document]
adrianna-chang-shopify: Ah, smart!

README.md
```diff
@@ -316,9 +316,30 @@ you can define an error handler:
...
 # config/initializers/maintenance_tasks.rb
-MaintenanceTasks.error_handler = ->(error) { Bugsnag.notify(error) }
+MaintenanceTasks.error_handler = ->(error, task_context, _errored_element) do
+  Bugsnag.notify(error) do |notification|
+    notification.add_tab(:task, task_context)
+  end
+end
 ```
 
+The error handler must be an object that responds to `.call`. It should accept
```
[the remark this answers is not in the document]
adrianna-chang-shopify: Will make the change, but out of curiosity, what kind of object would respond to `.call` but not to `arity`? Lambdas, procs, and blocks all would, as would a new type of object with a `call` method defined...?

README.md
```diff
@@ -319,9 +319,32 @@ you can define an error handler:
...
-MaintenanceTasks.error_handler = ->(error) { Bugsnag.notify(error) }
+MaintenanceTasks.error_handler = ->(error, task_context, _errored_element) do
+  Bugsnag.notify(error) do |notification|
+    notification.add_tab(:task, task_context)
+  end
+end
 ```
 
+The error handler should be a lambda that accepts three arguments,
+in the following order:
```
etiennebarrie, line 330:
```suggestion
The error handler should be a lambda that accepts three arguments:
```

etiennebarrie (approved):
Let's squash the commits, but everything looks good!

PR #322, opened 2021-02-04
Enqueue CSVs from the CLI

adrianna-chang-shopify:
This should be the last missing piece for CSV functionality in the gem. This PR allows a file to be passed via a `--csv` option on the CLI in order to run CSV processing Tasks from the command line.

## Changes
- https://github.com/Shopify/maintenance_tasks/commit/039e3e9889e348ea5dd2671cb860a7848160730d adds a `--csv` option to the CLI. If a file is supplied, we build an attachable (see https://edgeguides.rubyonrails.org/active_storage_overview.html#attaching-file-io-objects) to send to `Runner.run`. The CLI will surface the error if the file doesn't exist / isn't openable / etc.
- https://github.com/Shopify/maintenance_tasks/commit/ab1ee8c943867b6f7265de4efb43fe9e07c14c97 modifies `Runner.run` to verify that the Task processes CSVs before attaching the CSV (since a user can now pass a CSV for a non-CSV Task via the CLI).
     - Why not do this check in the CLI? This would involve performing `Task.named` from the CLI to check if the Task includes the `CsvCollection` module, and I felt like the CLI should not be aware of such constants and the details of how `Task` and `CsvCollection` are related
- https://github.com/Shopify/maintenance_tasks/commit/84074ee1033cb68ccf506e29238e228b72f50cde this commit could potentially be dropped - it makes it so that if a CSV task is run from the Runner without a `csv_file` supplied, we raise. 
     - Why _wouldn't_ we want this? The system will fail gracefully without this code, in the same way that it does if you run a CSV Task from the UI without attaching a file:
Screen Shot 2021-02-04 at 12 15 16 PM
     - Why _would_ we want this? It offers a better UX from the CLI in my opinion, as we fail early without enqueuing and then they know to supply a CSV from the command line.
-  https://github.com/Shopify/maintenance_tasks/commit/8867b515f52cfd50c34d6069298971cdd1256e18 updates the docs, including adding documentation for using `Runner.run` to run a CSV Task (which is possible today, but not documented)

## Tophatting
The following scenarios are relevant to 🎩 
In `test/dummy`:
*  `be maintenance_tasks perform 'Maintenance::ImportPostsTask'` =>
     * produces `error  Maintenance::ImportPostsTask must be supplied with a CSV file to process.`
     * If we drop https://github.com/Shopify/maintenance_tasks/commit/84074ee1033cb68ccf506e29238e228b72f50cde, will enqueue successfully, and error `Cannot parse nil as CSV` will be surfaced in the UI
*  `be maintenance_tasks perform 'Maintenance::ImportPostsTask' --csv 'invalid.csv'` =>
     * produces `error  No such file or directory @ rb_sysopen - invalid.csv`
* `be maintenance_tasks perform 'Maintenance::ImportPostsTask' --csv '../fixtures/files/sample.csv'`
     * Make sure to use inline queue adapter / Sidekiq / etc.. so this runs
     * This should be successful
*  `be maintenance_tasks perform 'Maintenance::UpdatePostsTask'` --csv '../fixtures/files/sample.csv'`
     * Should be successful
     * Should not attach a CSV to the Run for this Task

### DIFF
--- a/app/models/maintenance_tasks/run.rb
+++ b/app/models/maintenance_tasks/run.rb
@@ -33,6 +33,8 @@ module MaintenanceTasks
     validates :task_name, on: :create, inclusion: { in: ->(_) {
       Task.available_tasks.map(&:to_s)
     } }
+    validate :csv_attachment_presence, on: :create
+
     attr_readonly :task_name
 
     serialize :backtrace
@@ -176,6 +178,21 @@ module MaintenanceTasks
     # @return [Boolean] whether the Run is stuck.
     def stuck?
       cancelling? && updated_at <= 5.minutes.ago
+    end
+
+    # Performs validation on the presence of a :csv_file attachment.
+    # A Run for a Task that uses CsvCollection must have an attached :csv_file
+    # to be valid. Conversely, a Run for a Task that doesn't use CsvCollection
+    # should not have an attachment to be valid. The appropriate error is added
+    # if the Run does not meet the above criteria.
+    def csv_attachment_presence
+      if Task.named(task_name) < CsvCollection && !csv_file.attached?
+        errors.add(:csv_file, 'must be attached to CSV Task.')
+      elsif !(Task.named(task_name) < CsvCollection) && csv_file.attached?
+        errors.add(:csv_file, 'should not be attached to non-CSV Task.')
+      end
+    rescue Task::NotFoundError
+      nil
     end
   end
   private_constant :Run
--- a/README.md
+++ b/README.md
@@ -229,10 +229,26 @@ Alternatively, you can run your Task in the command line:
 $ bundle exec maintenance_tasks perform Maintenance::UpdatePostsTask
 ```
 
+To run a Task that processes CSVs from the command line, use the --csv option:
+
+```bash
+$ bundle exec maintenance_tasks perform Maintenance::ImportPostsTask --csv 'path/to/my_csv.csv'
+```
+
 You can also run a Task in Ruby by sending `run` with a Task name to Runner:
 
 ```ruby
 MaintenanceTasks::Runner.run(name: 'Maintenance::UpdatePostsTask')
+```
+
+To run a Task that processes CSVs using the Runner, provide a Hash containing an
+open IO object and a filename to `run`:
+
+```ruby
+MaintenanceTasks::Runner.run(
+  name: 'Maintenance::ImportPostsTask'
+  csv_file: { io: File.open('path/to/my_csv.csv'), filename: 'my_csv.csv' }
+)
 ```
 
 ### Monitoring your Task's status
--- a/lib/maintenance_tasks/cli.rb
+++ b/lib/maintenance_tasks/cli.rb
@@ -24,16 +24,31 @@ module MaintenanceTasks
       #{MaintenanceTasks::Task.available_tasks.join("\n\n")}
     LONGDESC
 
+    # Specify the CSV file to process for CSV Tasks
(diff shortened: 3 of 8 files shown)


app/models/maintenance_tasks/runner.rb
```diff
@@ -39,8 +40,12 @@ def initialize(run)
     def run(name:, csv_file: nil)
       run = Run.active.find_by(task_name: name) || Run.new(task_name: name)
 
-      if Task.named(name) < CsvCollection && csv_file
-        run.csv_file.attach(csv_file)
+      if Task.named(name) < CsvCollection
+        if csv_file
+          run.csv_file.attach(csv_file)
+        else
+          raise "#{name} must be supplied with a CSV file to process."
```
etiennebarrie, line 47: I think we should make it a ValidationError from a Run missing the CSV for a CSV task.

app/models/maintenance_tasks/runner.rb
```diff
@@ -39,8 +40,12 @@ def initialize(run)
     def run(name:, csv_file: nil)
       run = Run.active.find_by(task_name: name) || Run.new(task_name: name)
 
-      if Task.named(name) < CsvCollection && csv_file
-        run.csv_file.attach(csv_file)
+      if Task.named(name) < CsvCollection
```
etiennebarrie, line 43: We could also add a validation for a CSV given to a non-CSV task.

etiennebarrie (commented):
We could also support `--csv -` for stdin, and also just loading from the stdin if it's not a TTY when using pipes like `curl https://example.com/posts.csv | bundle exec maintenance_tasks perform Maintenance::ImportPostsTask` but than can be done in another PR.

app/models/maintenance_tasks/run.rb
```diff
@@ -177,6 +179,20 @@ def cancel
...
       cancelling? && updated_at <= 5.minutes.ago
     end
+
+    # Performs validation on the presence of a :csv_file attachment.
+    # A Run for a Task that uses CsvCollection must have an attached :csv_file
+    # to be valid. Conversely, a Run for a Task that doesn't use CsvCollection
+    # should not have an attachment to be valid. The appropriate error is added
+    # if the Run does not meet the above criteria.
+    def csv_attachment_presence
+      return if errors[:task_name].any?
```
adrianna-chang-shopify, line 189: If we've already failed validation because the Task name did not exist, calling `Task.named` will blow up, so we should just return early.

etiennebarrie: I'd rather rescue `Task::NotFoundError` than rely on the other validation having run before.

app/models/maintenance_tasks/run.rb
```diff
@@ -177,6 +179,20 @@ def cancel
...
+      if Task.named(task_name) < CsvCollection && !csv_file.attached?
+        errors.add(:csv_file, 'must be attached to Csv Task.')
```
etiennebarrie, line 191:
```suggestion
        errors.add(:csv_file, 'must be attached to CSV Task.')
```

app/models/maintenance_tasks/run.rb
```diff
@@ -177,6 +179,20 @@ def cancel
...
+      elsif !(Task.named(task_name) < CsvCollection) && csv_file.attached?
+        errors.add(:csv_file, 'should not be attached to non-Csv Task.')
```
etiennebarrie, line 193:
```suggestion
        errors.add(:csv_file, 'should not be attached to non-CSV Task.')
```

etiennebarrie (approved):

PR #326, opened 2021-02-09
Introduce Task#enumerator_builder

### LINKED ISSUE #207: Provide support for enumerable collections beyond Array and ActiveRecord::Relation in TaskJob

sambostock:
This introduces the `Task#enumerator_builder` method discussed in #307, #310, and "IRL", as a mechanism to allow tasks requiring custom enumerators (e.g. over API records), as described in #207, **without introducing any new modules or classes.**

`Task#enumerator_builder` inspects the type of `Task#collection` and wraps it in an object which can produce an enumerator on demand (given a `cursor`). This allows `TaskJob#build_enumerator` to be simplified.

## ⚠️ Before Merging

- [ ] Rebase into meaningful commits
- [ ] Add tests for each kind of enumerator builder
- [ ] Update README & other documentation

----

Closes #307 (redundant)
Closes #310 (redundant)
Closes #207 (implemented)

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,9 +10,11 @@ A Rails engine for queuing and managing maintenance tasks.
 * [Usage](#usage)
   * [Creating a Task](#creating-a-task)
   * [Creating a CSV Task](#creating-a-csv-task)
+  * [Creating a custom Task](#creating-a-custom-task)
   * [Considerations when writing Tasks](#considerations-when-writing-tasks)
   * [Writing tests for a Task](#writing-tests-for-a-task)
   * [Writing tests for a CSV Task](#writing-tests-for-a-csv-task)
+  * [Writing tests for a custom Task](#writing-tests-for-a-custom-task)
   * [Running a Task](#running-a-task)
   * [Monitoring your Task's status](#monitoring-your-tasks-status)
   * [How Maintenance Tasks runs a Task](#how-maintenance-tasks-runs-a-task)
@@ -142,6 +144,12 @@ My Title,Hello World!
 
 ### Creating a custom Task
 
+If you have a special use case requiring iteration over an unsupported resource, you can implement the `enumerator_builder` method instead. For example, you may need to iterate over external resources fetch from some API.
+
+This method should return an object responding to `enumerator(context:)` with an Enumerator, yielding pairs of `[item, item_cursor]`. In order for your enumerator to support resuming iteration part way through, you may use the `context.cursor`.
+
+You may optionally provide an implementation for `count`, if appropriate.
+
 ```ruby
 # app/tasks/maintenance/pay_last_months_invoices_task.rb
 module Maintenance
@@ -184,6 +192,33 @@ module Maintenance
 end
 ```
 
+In some cases, you may have no use for a cursor (e.g. iterating over some collection until it is empty), in which case your Enumerator may yield `nil` cursors (i.e. pairs of `[item, nil]`).
+
+```ruby
+# app/tasks/maintenance/pay_last_months_invoices_task.rb
+module Maintenance
+  class UnbanUsersTask < MaintenanceTasks::Task
+    def enumerator_builder
+      ExpiredBansEnumerator.new
+    end
+
+    def process(expired_ban)
+      expired_ban.user.unban!
+    end
+
+    class ExpiredBansEnumerator
+      def enumerator(context:)
+        UserManagementAPI
+          .resource(:bans)
+          .filter("expires_at <= #{Time.now}")
+          .auto_paginate
+          .lazy
+          .map { |ban| [ban, nil] }
+      end
+    end
+  end
+end
+```
 
 ### Considerations when writing Tasks
 
@@ -253,6 +288,57 @@ module Maintenance
       post = Post.last
       assert_equal 'My Title', post.title
       assert_equal 'Hello World!', post.content
+    end
+  end
+end
+```
+
+### Writing test for a custom Task
+
+As with other tasks, you should write tests for your `#process` method. It will receive the first item in each pair your custom Enumerator yields (`[item, item_cursor]`).
+
+You should also ensure your Enumerator is tested by unit testing it in isolation, testing your `#enumerator_builder` method, or both. Make sure you test how your Enumerator handles the absence or presence of a `context.cursor`, if applicable.
+
+```ruby
+# app/tasks/maintenance/shopping_list_task.rb
+module Maintenance
+  class ShoppingListTask < ActiveSupport::TestCase
+    test "#process adds the recipe's ingredients to the shopping list" do
+      ShoppingList.clear
+      recipe = recipes(:tacos)
+      ingredient = recipe.ingredient.first
+
+      Maintenance::ShoppingListTask.process(ingredient)
+
+      assert_includes ShoppingList, ingredient
+    end
+
+    test "#build_enumerator's enumerator yields all ingredients for all recipes, when no cursor provided" do
+      all_ingredient_pairs = ingredient_pairs
+      enumerator = Maintenance::ShoppingListTask.build_enumerator.enumerator(context: stub(cursor: nil))
+
+      assert_equal all_ingredient_pairs, enumerator.to_a
+    end
+
+    test "#build_enumerator's enumerator yields remaining ingredients for all recipes, when cursor provided" do
+      remaining_ingredient_pairs = ingredient_pairs
+      previous_ingredient_pairs = remaining_ingredient_pairs.shift(5)
+      cursor = previous_ingredient_pairs.last.last
+
+      enumerator = Maintenance::ShoppingListTask.build_enumerator.enumerator(context: stub(cursor: cursor))
+
+      assert_equal remaining_ingredient_pairs, enumerator.to_a
+    end
+
+    private
+
+    def ingredient_pairs
+      recipes.flat_map do |recipe|
+        recipe.ingredients.map do |ingredient|
+          cursor = { recipe_id: recipe.id, ingredient_id: ingredient.id }
+          [ingredient, cursor]
+        end
+      end
     end
   end
 end
--- a/README.md
+++ b/README.md
@@ -140,6 +140,51 @@ title,content
 My Title,Hello World!
 ```
 
+### Creating a custom Task
+
+```ruby
+# app/tasks/maintenance/pay_last_months_invoices_task.rb
+module Maintenance
+  class PayLastMonthsInvoicesTask < MaintenanceTasks::Task
+    def enumerator_builder
+      InvoiceEnumerator.new
+    end
+
+    def count
+      InvoiceAPI.last_month.count
+    end
+
+    def process(invoice)
+      invoice.pay!
+    end
+
+    class InvoiceEnumerator
+      def enumerator(context:)
+        Enumerator.new do |yielder|
+          cursor = context.cursor
+
+          loop do
+            page = if cursor.nil?
+              InvoiceAPI.last_month(max: 100)
+            else
+              InvoiceAPI.last_month(max: 100, after: context.cursor)
+            end
+
+            page.entries.each do |invoice|
+              cursor = invoice.id
+              yielder.yield [invoice, cursor]
+            end
+
+            break unless page.has_next?
+          end
+        end
+      end
+    end
+  end
+end
+```
+
+
 ### Considerations when writing Tasks
 
 MaintenanceTasks relies on the queue adapter configured for your application to
--- a/README.md
+++ b/README.md
@@ -144,6 +144,8 @@ My Title,Hello World!
 
 ### Creating a custom Task
 
+TODO: Add generation instructions
+
 If you have a special use case requiring iteration over an unsupported resource, you can implement the `enumerator_builder` method instead. For example, you may need to iterate over external resources fetch from some API.
 
 This method should return an object responding to `enumerator(context:)` with an Enumerator, yielding pairs of `[item, item_cursor]`. In order for your enumerator to support resuming iteration part way through, you may use the `context.cursor`.
--- a/README.md
+++ b/README.md
@@ -153,36 +153,44 @@ This method should return an object responding to `enumerator(context:)` with an
 You may optionally provide an implementation for `count`, if appropriate.
 
 ```ruby
-# app/tasks/maintenance/pay_last_months_invoices_task.rb
+# app/tasks/maintenance/shopping_list_task.rb
 module Maintenance
-  class PayLastMonthsInvoicesTask < MaintenanceTasks::Task
+  class ShoppingListTask < MaintenanceTasks::Task
     def enumerator_builder
-      InvoiceEnumerator.new
+      IngredientsEnumerator.new
     end
 
-    def count
-      InvoiceAPI.last_month.count
+    def process(ingredient)
+      ShoppingList.add(ingredient)
     end
 
-    def process(invoice)
-      invoice.pay!
-    end
-
-    class InvoiceEnumerator
+    class IngredientEnumerator
       def enumerator(context:)
         Enumerator.new do |yielder|
           cursor = context.cursor
 
+          if cursor.nil?
+            recipe = FancyRecipeAPI.random_recipe
+            ingredient_id = nil
+          else
+            recipe_id, ingredient_id = cursor.split(':', 2)
+            recipe = FancyRecipeAPI.recipe(recipe_id)
+          end
+
           loop do
-            page = if cursor.nil?
-              InvoiceAPI.last_month(max: 100)
+            page = if ingredient_id.nil?
+              FancyRecipeAPI
+                .paginated_ingredients(recipe.id, max: 5)
             else
-              InvoiceAPI.last_month(max: 100, after: context.cursor)
+              FancyRecipeAPI
+                .paginated_ingredients(recipe.id, max: 5, after: ingredient_id)
             end
 
-            page.entries.each do |invoice|
-              cursor = invoice.id
-              yielder.yield [invoice, cursor]
+            page.entries.each do |ingredient|
+              ingredient_id = ingredient.id
+              cursor = "#{recipe.id}:#{ingredient.id}"
+
+              yielder.yield([ingredient, cursor])
             end
 
             break unless page.has_next?
@@ -197,25 +205,30 @@ end
 In some cases, you may have no use for a cursor (e.g. iterating over some collection until it is empty), in which case your Enumerator may yield `nil` cursors (i.e. pairs of `[item, nil]`).
 
 ```ruby
-# app/tasks/maintenance/pay_last_months_invoices_task.rb
+# app/tasks/maintenance/ingredient_purge_task.rb
 module Maintenance
-  class UnbanUsersTask < MaintenanceTasks::Task
+  class IngredientPurgeTask < MaintenanceTasks::Task
     def enumerator_builder
-      ExpiredBansEnumerator.new
+      ExpiredIngredientsEnumerator.new
     end
 
-    def process(expired_ban)
-      expired_ban.user.unban!
+    def count
+      PantryAPI.ingredients(expired: true).count +
+        FridgeAPI.ingredients(expired: true).count +
+        FreezerAPI.ingredients(expired: true).count
     end
 
-    class ExpiredBansEnumerator
-      def enumerator(context:)
-        UserManagementAPI
-          .resource(:bans)
-          .filter("expires_at <= #{Time.now}")
-          .auto_paginate
-          .lazy
-          .map { |ban| [ban, nil] }
+    def process(ingredient)
+      ingredient.compost!
+    end
+
+    class ExpiredIngredientsEnumerator
+      def enumerator(*)
+        Enumerator.chain(
+          PantryAPI.ingredients(expired: true).auto_paginate,
+          FridgeAPI.ingredients(expired: true).auto_paginate,
+          FreezerAPI.ingredients(expired: true).auto_paginate,
+        ).lazy.map { |ingredient| [ingredient, nil] }
       end
     end
   end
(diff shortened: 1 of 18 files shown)


app/jobs/maintenance_tasks/task_job.rb
```diff
@@ -25,21 +25,14 @@ def retry_on(*, **)
...
-        JobIteration::CsvEnumerator.new(collection).rows(cursor: cursor)
-      else
-        raise ArgumentError, "#{@task.class.name}#collection must be either "\
-          'an Active Record Relation, Array, or CSV.'
-      end
+      context = EnumerationContext.new(
+        cursor: cursor || @run.cursor,
+      )
+
+      @task.enumerator_builder.enumerator(context: context)
```
sambostock, line 35: We build a context object instead of coupling to a `cursor:` keyword argument to allow forward compatibility, should we need to pass in more in the future, as @rafaelfranca mentioned in https://github.com/Shopify/maintenance_tasks/pull/310#issuecomment-767783835.

app/models/maintenance_tasks/csv_collection.rb
```diff
@@ -13,6 +13,18 @@ module CsvCollection
...
+    CsvEnumeratorBuilder = Struct.new(:csv) do
+      def enumerator(context:)
+        JobIteration::CsvEnumerator.new(csv).rows(cursor: context.cursor)
+      end
+    end
+    private_constant :CsvEnumeratorBuilder
+
+    def enumerator_builder
+      CsvEnumeratorBuilder.new(collection)
+    end
```
sambostock, line 26: Since this module exists, it felt like the natural place to put the CSV enumerator code, instead of directly in `Task#enumerator_builder`.

test/dummy/app/tasks/maintenance/custom_enumerating_task.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+module Maintenance
+  class CustomEnumeratingTask < MaintenanceTasks::Task
```
sambostock, line 3: I created this for use in tests, but I wasn't able to figure out where the test would belong. It seems like a lot of the functionality of `Task` doesn't have unit tests, and is instead "integration" tested by actually running a task. 🤔

sambostock (commented):
I realized I was kind of stuck moving forward on this branch, so I figured I'd just throw up a draft and go from there.

test/dummy/app/tasks/maintenance/custom_enumerating_task.rb
adrianna-chang-shopify: Yeah, I think in keeping with how we test our other Task "types", I'd write a test `.perform_now can use custom enumerator instead of collection` in `task_job_test.rb` for now to act as our integration test (it can go with `test '.perform_now accepts Active Record Relations as collection'` and `test '.perform_now accepts CSVs as collection'`), and have it run this new Task. We could write some unit tests for the new enumeration building logic that's been added to `task.rb`, but it is all tested through `task_job_test.rb`, and since it's all private API I'd be okay to just continue relying on the tests in TaskJob to exercise that functionality.

README.md
```diff
@@ -146,6 +146,51 @@ title,content
 My Title,Hello World!
 ```
 
+### Creating a custom Task
```
adrianna-chang-shopify, line 149: Can you add this section in the [Table of Contents](https://github.com/Shopify/maintenance_tasks#table-of-contents) too?

app/tasks/maintenance_tasks/task.rb
```diff
@@ -74,6 +83,44 @@ def load_constants
...
+      case collection
+      when ActiveRecord::Relation
+        ActiveRecordEnumeratorBuilder.new(collection)
+      when Array
+        ArrayEnumeratorBuilder.new(collection)
+      else
+        raise ArgumentError, "#{self.class.name}#collection must be either "\
+          'an Active Record Relation, or Array.' # TODO: update
+        # If you want CSVs, do this
+        # If you want custom enum, do this
```
adrianna-chang-shopify, line 120: I'm actually inclined to leave the error message exactly as you have it `("#{self.class.name}#collection must be either an Active Record Relation or Array."`) . Since neither CSV Tasks nor custom enumeration Tasks require a `#collection` method to be filled in, I think it's reasonable for the error to reference the only two types of objects that `collection` actually supports.

Maybe we can add some additional wording in the Task generator to help folks understand the different types of Tasks available to them instead.

adrianna-chang-shopify (commented):
Thanks Sam! 🙌  A couple of comments, this is looking really solid to me though.
Here are the things I would want to see before prepping to merge this:
- A little blurb before the example Task is the README that describes why you might want to use a custom enumerator (simple example like reading off a queue would suffice) and explaining what method(s) need to be implemented (similar to [the CSV Task](https://github.com/Shopify/maintenance_tasks#creating-a-csv-task))
- We may want to give an example of how to write a test for a custom enumerator Task to go with our others, but I'd be okay to leave this out for now as well.
- A test in `task_job_test.rb` that exercises the functionality of the new custom enum Task. I'll defer to you whether you'd also like to unit test the enumerator builder logic that's been added to Task - I'd be okay without it.
- It would be really cool to expand the [Task Generator](https://github.com/Shopify/maintenance_tasks/blob/main/lib/generators/maintenance_tasks/task_generator.rb) to have a `--custom` option that gives you a template with `enumerator_builder`, `process` and `count` laid out as methods to define. I can do this as a follow up though if you'd prefer.

adrianna-chang-shopify:
Oh, also the new `Maintenance::CustomEnumeratingTask` just needs to be added to the lists of Tasks in the two tests that are failing CI right now for that to be green ✅

test/dummy/app/tasks/maintenance/custom_enumerating_task.rb
sambostock: `Task#enumerator_builder` is `public` though, and definitely part of the public API if we're telling consumers to override it to implement custom tasks. I'm not sure if that changes anything, since the unit tests for the other public methods are pretty lightweight.

adrianna-chang-shopify: Ah true,`Task#enumerator_builder` is indeed public now. And it does have branching, so that could be an indication that we should test it (even if it's implicitly tested in `TaskJob`). The unit tests for the other methods are mostly laying out the template for a Task, which is why they're so lightweight.

Anyways, I'll leave it up to you, IMHO it comes down to personal preference mostly 😅  We have a test for each collection type, and another for the error scenario all in `TaskJobTest`, so I consider it to have test coverage, but totally up to you!

README.md
```diff
@@ -146,6 +148,86 @@ title,content
...
+
+TODO: Add generation instructions
+
+If you have a special use case requiring iteration over an unsupported resource, you can implement the `enumerator_builder` method instead. For example, you may need to iterate over external resources fetch from some API.
```
adrianna-chang-shopify, line 155:
```suggestion
If you have a special use case requiring iteration over an unsupported collection type, such as external resources fetched from some API, you can implement the `enumerator_builder` method instead.
```

README.md
```diff
@@ -146,6 +148,86 @@ title,content
...
+
+This method should return an object responding to `enumerator(context:)` with an Enumerator, yielding pairs of `[item, item_cursor]`. In order for your enumerator to support resuming iteration part way through, you may use the `context.cursor`.
+
+You may optionally provide an implementation for `count`, if appropriate.
```
adrianna-chang-shopify, line 159: Maybe we should add a line about needing to implement `process` as well?

README.md
```diff
@@ -219,6 +301,57 @@ module Maintenance
...
+### Writing test for a custom Task
+
+As with other tasks, you should write tests for your `#process` method. It will receive the first item in each pair your custom Enumerator yields (`[item, item_cursor]`).
+
+You should also ensure your Enumerator is tested by unit testing it in isolation, testing your `#enumerator_builder` method, or both. Make sure you test how your Enumerator handles the absence or presence of a `context.cursor`, if applicable.
+
+```ruby
+# app/tasks/maintenance/shopping_list_task.rb
+module Maintenance
+  class ShoppingListTask < ActiveSupport::TestCase
```
adrianna-chang-shopify, line 313: THESE TEST EXAMPLES THO! ❤️  🌮  😋

README.md
```diff
@@ -146,6 +148,86 @@ title,content
...
+TODO: Add generation instructions
+
+If you have a special use case requiring iteration over an unsupported resource, you can implement the `enumerator_builder` method instead. For example, you may need to iterate over external resources fetch from some API.
+
+This method should return an object responding to `enumerator(context:)` with an Enumerator, yielding pairs of `[item, item_cursor]`. In order for your enumerator to support resuming iteration part way through, you may use the `context.cursor`.
+
+You may optionally provide an implementation for `count`, if appropriate.
+
+```ruby
+# app/tasks/maintenance/pay_last_months_invoices_task.rb
```
adrianna-chang-shopify, line 162: Love both of these examples, but WDYT about having a `Maintenance::ShoppingListTask` example that readers can refer to for the `Writing test for a custom Task` section? I think it would improve the understandability of the test examples to see a corresponding Task 😄

README.md
sambostock: I wasn't sure about that one, since it seems like a given. Maybe if I integrate it into the previous paragraph?

> ...yielding pairs of `[item, item_cursor]` **to the `process` method your `Task` defines.** In order for...

README.md
sambostock: Yeah, I was wondering the same thing after I wrote them out. I'll throw something together.

It's funny writing fake code for made up problems 🤣

sambostock: Done. I replaced `PayLastMonthsInvoicesTask` with `ShoppingListTask`, and `UnbanUsersTask` with `PurgeIngredientsTask`, which I then included tests for. I took the opportunity to describe testing the enumerator builder inline and in isolation, respectively. If it's too wordy, I figure we can collapse them using `<details />`.

I actually had to change up my hypothetical `ShoppingListTask`, as the behaviour I had been "testing" for would have been really gnarly to implement because of the nested iteration and composite cursor 😅

README.md
adrianna-chang-shopify: Perfect!

README.md
```diff
@@ -146,6 +148,106 @@ title,content
...
+          end
+        end
+      end
+    end
+  end
+end
+```
+
+In some cases, you may have no use for a cursor (e.g. iterating over some
+collection until it is empty), in which case your Enumerator may yield `nil`
```
adrianna-chang-shopify, line 217:
```suggestion
In some cases, you may have no use for a cursor (e.g. taking items off a queue
until it is empty), in which case your Enumerator may yield `nil`
```
Maybe `queue` instead of `collection` to make things clearer?

README.md
```diff
@@ -219,6 +321,163 @@ module Maintenance
...
+      recipe.ingredients.map do |ingredient|
+        cursor = "#{recipe.id}:#{ingredient.id}"
+        [ingredient, cursor]
+      end
+    end
+  end
+end
+```
+
+Dependaing on its complexity, you may choose to test your Enumerator builder in
```
adrianna-chang-shopify, line 391:
```suggestion
Depending on its complexity, you may choose to test your Enumerator builder in
```

README.md
```diff
@@ -219,6 +321,163 @@ module Maintenance
...
+  class ShoppingListTaskTest < ActiveSupport::TestCase
+    test '#process adds ingredients to the shopping list' do
+      ingredient = recipes(:tacos).first
+      ShoppingList.expects(:add).with(ingredient)
+
+      ShoppingListTask.process(ingredient)
+    end
+
+    test '#enumerator_builder.enumerator enumerates ingredients for a random recipe' do
+      FancyRecipeApi.fake_it_till_you_make_it do
```
adrianna-chang-shopify, line 347:
```suggestion
      FancyRecipeAPI.fake_it_till_you_make_it do
```

README.md
```diff
@@ -219,6 +321,163 @@ module Maintenance
...
+        context = stub('context', cursor: nil)
+        actual_ingredient_pairs = ShoppingListTask.enumerator_builder
+          .enumerator(context: context).to_a
+
+        assert_equal expected_ingredient_pairs, actual_ingredient_pairs
+      end
+    end
+
+    test '#enumerator_builder.enumerator enumerates remaining ingredients for the cursor recipe' do
+      FancyRecipeApi.fake_it_till_you_make_it do
```
adrianna-chang-shopify, line 362:
```suggestion
      FancyRecipeAPI.fake_it_till_you_make_it do
```

README.md
```diff
@@ -219,6 +321,163 @@ module Maintenance
...
+
+  test '#enumerator is empty if no expired ingredients' do
+    PantryAPI.stock(ingredients(:tortillas))
+
+    assert_empty ExpiredIngredientsEnumerator.enumerator(context: context).to_a
+  end
+end
+```
+
+</details>
```
adrianna-chang-shopify, line 479: This is a good idea! 👍  I think we might need to wrap the entire thing in a paragraph element to get the code blocks to work inside a collapsible:

```
<details><p>

{{code examples}}
</p></details>

**Review by adrianna-chang-shopify — COMMENTED** [review]
Thanks for the care you put into crafting the README ❤️  The examples are awesome. I'm a tiny bit concerned that they might be erring on the complex side, but then again, developing a custom enumerator is a complex thing, and I like that the README shows a working example and the different test cases to cover. We could always pull these out into a separate guide like [JobIteration](https://github.com/Shopify/job-iteration/blob/master/guides/custom-enumerator.md) has done if we feel they're making the README too lengthy.

Couple tiny nits, let's get those tests in, and then this is ✅  from my perspective. The generator stuff can go out separately.

@etiennebarrie could you give this a review when you have a sec please? 🙏

File: README.md (hunk 14 above)
Comment (sambostock) [reply README.md:479]: Yeah, I've run into that issue before, but I think the blank line is enough.

Indeed,


    <details><summary>Busted code snippet</summary>
    ```ruby
    puts ':('
    ```
    </details>

produces

<details><summary>Busted code snippet</summary>
```ruby
puts ':('
```
</details>

but in my experience all it needs to work is a blank line before it, as you've incidentally included in your example:

    <details><summary>Working code snippet</summary>

    ```ruby
    puts ':)'
    ```
    </details>
    
produces

<details><summary>Working code snippet</summary>

```ruby
puts ':)'
```
</details>

The `<p>` doesn't seem to be needed, unless I missed something.

app/jobs/maintenance_tasks/task_job.rb
etiennebarrie: Yet we're now coupling it to the `context:` keyword argument. I think `enumerator`should have an normal parameter.

app/tasks/maintenance_tasks/task.rb
```diff
@@ -74,9 +83,57 @@ def load_constants
...
+    # .enumerator will respond to .cursor, returning the previous item's cursor,
+    # or nil (if this is the first iteration), to support resuming enumeration.
+    def enumerator_builder
+      collection = self.collection
+
+      case collection
+      when ActiveRecord::Relation
+        ActiveRecordEnumeratorBuilder.new(collection)
+      when Array
+        ArrayEnumeratorBuilder.new(collection)
```
etiennebarrie, line 124: I'm not sure I like this. Previously we were calling collection and:
* creating an enumerator builder
* creating an enumerator with it, passing it the collection and cursor


Now we are:
* creating a enumerator builder with the collection
* calling enumerator on it with the context/cursor, which in turn is
* creating an enumerator builder
* creating an enumerator with the collection/cursor

I get that it's basically an adapter pattern, and it's neat for CSV to not be leaking here, but maybe it's not worth wrapping the ActiveRecord and Array enumerators here?

README.md
```diff
@@ -140,6 +142,106 @@ title,content
...
+        end
+      end
+    end
+  end
+end
+```
+
+In some cases, you may have no use for a cursor (e.g. taking items off a queue
+until it is empty), in which case your Enumerator may yield `nil`
+cursors (i.e. pairs of `[item, nil]`).
```
etiennebarrie, line 212: Wouldn't that work by simply yielding the item by itself?
```ruby
foo = Enumerator.new { |y| y.yield :item }
# => #<Enumerator: #<Enumerator::Generator:0x00007faf8dac0b90>:each>
foo.each do |item, cursor| p item, cursor end
# :item
# nil
# => [:item, nil]
```

sambostock: Yeah, we identified both
- wanting to have consumers be explicit about not having a cursor, as well as
- the existence of a "footgun", in the case where the `element` they want to yield is an array
  (say they're implementing a cursorless custom batched task, or forget to include a cursor for a custom task yielding pairs)

```ruby
def example(&block)
  Enumerator.new(&block).each do |element, cursor|
    puts "element: #{element.inspect.ljust(16)} cursor: #{cursor.inspect}"
  end
end

example { |y| y.yield :item, :cursor }              # element: :item            cursor: :cursor
example { |y| y.yield :item, nil }                  # element: :item            cursor: nil
example { |y| y.yield :item }                       # element: :item            cursor: nil

example { |y| y.yield [:item, :cursor] }            # element: :item            cursor: :cursor
example { |y| y.yield [:item, nil] }                # element: :item            cursor: nil
example { |y| y.yield [:item] }                     # element: :item            cursor: nil

example { |y| y.yield [[:pair1, :pair2], :cursor] } # element: [:pair1, :pair2] cursor: :cursor
example { |y| y.yield [[:pair1, :pair2], nil] }     # element: [:pair1, :pair2] cursor: nil
example { |y| y.yield [[:pair1, :pair2]] }          # element: [:pair1, :pair2] cursor: nil

example { |y| y.yield [:pair1, :pair2], :cursor }   # element: [:pair1, :pair2] cursor: :cursor
example { |y| y.yield [:pair1, :pair2], nil }       # element: [:pair1, :pair2] cursor: nil
example { |y| y.yield [:pair1, :pair2] }            # element: :pair1           cursor: :pair2
                                                    # cursor forgotten or omitted?      ^^^^^^
```

Without an explicit cursor, we're unable to distinguish between yielding an array, or yielding and item and cursor. This occurs for any case where they array has more than one element:

```ruby
example { |y| y.yield [:pair1, :pair2, :pair3] }    # element: :pair1           cursor: :pair2
                                                    # cursor forgotten or omitted?      ^^^^^^
```

sambostock: That said, I'm not sure how to implement that restriction, given the way block variables are auto-unpacked, so we might have to leave it. 🤷

app/jobs/maintenance_tasks/task_job.rb
sambostock: The reasoning here wasn't that we didn't want to couple to a keyword argument, it was that if we ever needed to add more to the context, we'd have to make a breaking change to the signature.


```ruby
class SomeConsumersEnumeratorBuilder
  def enumerator(cursor:)
  end
end

@task.enumerator_builder.enumerator(cursor: cursor, new_thing: 123) # 💥 ArgumentError
```

```ruby
class SomeConsumersEnumeratorBuilder
  def enumerator(cursor)
  end
end

@task.enumerator_builder.enumerator(cursor, 123) # 💥 ArgumentError
```

By instead passing in a context object, we can simply add methods to it, and if the consumer doesn't care, they don't call it; no breaking change.

```ruby
class SomeConsumersEnumeratorBuilder
  def enumerator(context:)
  end
end

@task.enumerator_builder.enumerator(context: fancy_new_context) # ✅
```

As for the choice of a keyword argument, I've found it clarifies what to expect when it isn't obvious from the method name (compared to something like `delete_user(user)`). Here it makes it entirely unambiguous to consumers that they're receiving a context object, not a cursor or something else.

I can certainly change it to a positional argument if you think that's more appropriate, but that's why I chose `context:`.

app/tasks/maintenance_tasks/task.rb
sambostock: I see what you mean, however foregoing the enumerator builder for ActiveRecord & Array would mean we'd have to have conditionals to treat those as special cases either by doing something like
- calling `collection` first and then falling back to `enumerator_builder` in `TaskJob`, or
- returning an enumerator directly from `enumerator_builder` and then checking the return type in `TaskJob`

At that point, I'd be tempted to suggest we go back to my original implementation in #307

```ruby
def collection
  ->(context:) { ...build the enumerator...}
end
```

I think the reason we decided not to go with that, and to instead look at the intermediate enumerator builder abstraction was so that **only authors of custom enumerators would have to know about cursors**, and not authors of tasks simply consuming such enumerators, unlike JobIteration's API, which as @rafaelfranca pointed out requires everyone to know about cursors, even for simple cases:

```ruby
def build_enumerator(cursor:)
  enumerator_builder.array([1, 2, 3], cursor: cursor)
end
```

It seems a shame to introduce an API to address those concerns, while sidestepping it internally and fragmenting the flow.

adrianna-chang-shopify: Yeah I agree with Sam, although there's this additional cost of complexity in introducing this new abstraction, it's the price we pay to keep our simple, existing API intact while enabling custom enumeration. IMHO it's nice to have our internal enumerator builders conform to a single interface

app/jobs/maintenance_tasks/task_job.rb
adrianna-chang-shopify: Also +1 to Sam's comments here, I think foregoing the keyword argument results in too much ambiguity. Especially with users who are coming from having used JobIteration, I think it's highly likely they'd write `def enumerator(cursor)` and expect their method to receive a cursor object, since JobIteration's API was `build_enumerator(cursor:)`

I also think coupling it to something as generic as `context` gives us the flexibility to expand this to our liking in the future without breaking changes, as Sam mentioned.

README.md
```diff
@@ -213,6 +315,163 @@ module Maintenance
...
+        cursor = "#{recipe.id}:#{ingredient.id}"
+        [ingredient, cursor]
+      end
+    end
+  end
+end
+```
+
+Depending on its complexity, you may choose to test your Enumerator builder in
+isolation, in which case you can simplify the tests for your Task.
```
etiennebarrie, line 386: I don't think we should go so much into details about how people should test their apps.

README.md
```diff
@@ -140,6 +142,106 @@ title,content
...
+    class IngredientEnumerator
+      def enumerator(context:)
+        Enumerator.new do |yielder|
+          cursor = context.cursor
+
+          if cursor.nil?
+            recipe = FancyRecipeAPI.random_recipe
+            ingredient_id = nil
+          else
+            recipe_id, ingredient_id = cursor.split(':', 2)
```
etiennebarrie, line 181: I feel like this example with the two levels of id is a bit too complex. Also a bit weird that the ingredients would be paginated. Can we make it just about a list of recipes being paginated maybe?

Better yet we could use something commerce related in a nod to Shopify? https://shopify.dev/docs/admin-api/rest/reference/orders/order?api[version]=2020-04#index-2020-04 using `Order`, `since_id` and `limit` as the name/params in the example?

README.md
```diff
@@ -213,6 +315,163 @@ module Maintenance
...
 
+### Writing test for a custom Task
+
+As with other tasks, you should write tests for your `#process` method. It will
+receive the first item in each pair your custom Enumerator yields (`[item,
+item_cursor]`).
+
+You should also ensure your Enumerator is tested, by unit testing it in
+isolation, testing your `#enumerator_builder` method, or both. Make sure you
+test how your Enumerator handles the absence or presence of a `context.cursor`,
```
etiennebarrie, line 326: I think this is the main advice we should give them about testing their EnumeratorBuilder.

etiennebarrie (commented):
Ok hopefully I'm not wrong about ruby signatures stuff, but even if I am I think it should be enough to get the gist of it. The proposed API is:

```ruby
class Context[Cursor]
  def cursor: () -> Cursor
  # later we can add other things maybe?
  # def job: () -> TaskJob
end

interface _EnumeratorBuilder[Element, Cursor]
  def enumerator: (context: Context[Cursor]) -> Enumerable[[Element, Cursor]]
end
```

My preferred API would be:

```ruby
interface _EnumeratorBuilder[Element, Cursor]
  def call: (Context[Cursor]) -> Enumerable[[Element, Cursor]]
end
```

At its core, what we want from `Task#enumerator_builder` is for it to return something that can be called with a cursor/context to get an Enumerable, and IMO `call` is the best API in Ruby when you have some code to execute with arguments. If we had that, the `ActiveRecord::Relation` and `Array` cases would be much simpler.  
Yet, I can hear the argument that sometimes we should have more APIs than just `call` 🤣. So I can somewhat reluctantly get behind using `enumerator` for API, and having to define things like `ActiveRecordEnumeratorBuilder` and `ArrayEnumeratorBuilder`, and not being able to rely on closures.

Still, even if we go that way, I’d rather have:

```ruby
interface _EnumeratorBuilder[Element, Cursor]
  def enumerator: (Context[Cursor]) -> Enumerable[[Element, Cursor]]
end
```

i.e. no keyword arguments. I don’t think we should use keyword arguments when there’s only one parameter and it’s not going to change because Context will be the one to change if we need to add something.

---

Last thing, we need to find a name for this, since it’s not really the task that is custom. The API is about the Enumerator (being built by an Enumerator Builder object that is returned by `Task#enumerator_builder`, which responds to `#enumerator` itself). So maybe “Task with a custom Enumerator”?

README.md
adrianna-chang-shopify: Let's leave it, we'll make things clear in the documentation!

adrianna-chang-shopify (commented):
Sorry to keep prolonging the conversation on this 🙈  I was thinking about this some more, and feel like having `Task's` conform to a `#build_enumerator` method that returns an object that implements the `_EnumeratorBuilder` interface is unnecessarily complex when we could just have it build the `Enumerable` directly. @etiennebarrie maybe that's what you were getting at [here](https://github.com/Shopify/maintenance_tasks/pull/326#discussion_r580436345). The current API is, as you mentioned,  like this:

```ruby
class Context[Cursor]
  def cursor: () -> Cursor
end

class Task
  def enumerator_builder: () -> _Enumerator_Builder
end

interface _EnumeratorBuilder[Element, Cursor]
  def enumerator: (context: Context[Cursor]) -> Enumerable[[Element, Cursor]]
end
```

But why not just make it like this, and cut out the intermediate `_EnumeratorBuilder` interface?
```ruby
class Context[Cursor]
  def cursor: () -> Cursor
end

class Task
  def enumerator_builder: (context: Context[Cursor]) -> Enumerable[[Element, Cursor]]
end
```

We talked about this requiring special treatment to distinguish between our default cases (AR / Array / CSV) and what the author of a custom Task writes, but I don't think this is the case:


```ruby
  class TaskJob < ActiveJob::Base
    EnumerationContext = Struct.new(:cursor, keyword_init: true)

    def build_enumerator(_run, cursor:)
      context = EnumerationContext.new(cursor: cursor || @run.cursor)
      @task.enumerator_builder(context: context)
    end
  end

  class Task
    def enumerator_builder(context:)
      collection = self.collection

      case collection
      when ActiveRecord::Relation
        JobIteration::EnumeratorBuilder.new(nil).active_record_on_records(
          collection,
          cursor: context.cursor,
        )
      when Array
        JobIteration::EnumeratorBuilder.new(nil).build_array_enumerator(
          collection,
          cursor: context.cursor,
        )
      else
          'an Active Record Relation, or Array.'
      end
    end
  end
# Task#enumerator_builder returns JobIteration::EnumeratorBuilder::Wrapper, which implements `_Enumerable` interface

  module CsvCollection
    def enumerator_builder(context:)
      JobIteration::CsvEnumerator.new(self.collection).rows(cursor: context.cursor)
    end
  end
# CsvCollection#enumerator_builder returns Enumerator::Lazy, which implements `_Enumerable` interface

  class CustomEnumeratingTask < MaintenanceTasks::Task
    def enumerator_builder(context:)
      drop = context.cursor.nil? ? 0 : context.cursor + 1

      %i(a b c).lazy.with_index.drop(drop)
    end
  end
# CustomEnumeratingTask#enumerator_builder returns Enumerator::Lazy, which implements `_Enumerable` interface
```

So no extra type checks should be required. IMHO this simplifies the language in the README / templates / etc... Instead of "you can implement the `enumerator_builder` method instead, which should return an object responding to `enumerator(context:)` with an Enumerator yielding pairs of `[item, item_cursor]`", it's "you can implement the `enumerator_builder(context:)` method instead, which should return an Enumerator yielding pairs of `[item, item_cursor]`."

On the subject of naming:
> So maybe “Task with a custom Enumerator”?

Agreed, Sam had used `CustomEnumeratingTask` to describe the sample Task in the dummy, maybe this would work?

gmcgibbon:
Superseded by https://github.com/Shopify/maintenance_tasks/pull/944, so I am going to close this for now.

PR #329, opened 2021-02-10
Add rubygems source to Gemfile line in README

adrianna-chang-shopify:
Users at Shopify encounter a warning during `bundle install` due to the fact that the gem is registered at rubygems.org and in our internal package system:
```bash
Warning: the gem 'maintenance_tasks' was found in multiple sources.
Installed from: https://rubygems.org/
Also found in:
  * https://packages.shopify.io/shopify/gems/
You should add a source requirement to restrict this gem to your preferred source.
For example:
    gem 'maintenance_tasks', :source => 'https://rubygems.org/'
Then uninstall the gem 'maintenance_tasks' (or delete all bundled gems) and then install again.
```

 IMHO we should tell users to source the gem from the public source in the README to prevent this issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,16 +34,10 @@ Watch this demo video to see the gem in action:
 
 ## Installation
 
-Add this line to your application's Gemfile:
-
-```ruby
-gem 'maintenance_tasks'
-```
-
-And then execute:
+To install the gem and run the install generator, execute:
 
 ```bash
-$ bundle
+$ bundle add maintenance_tasks
 $ rails generate maintenance_tasks:install
 ```
 


etiennebarrie (commented):
Even when there's a warning it installs the latest version, which is from rubygems, so I'm not sure it's a problem.

Could we fix it by adding a version requirement instead? For most users, this seems redundant to add the source, whereas adding the major version may be useful:
`gem 'maintenance_tasks', '~> 1.0'` (`~> 1` also works but it's a special case). That way as long as we don't make breaking changes, people can bundle update freely.

Dependabot for example is able to go past this and also update the requirement if needed, so it's not blocking anybody either.

Another, maybe better, way to fix this is to point to `bundle add`, which does the same (add a major/minor pessimistic version requirement), and we can remove the whole paragraph and replace the call to `bundle` below instead.

adrianna-chang-shopify:
> Even when there's a warning it installs the latest version, which is from rubygems, so I'm not sure it's a problem.

Yep, I don't think it's actually problematic per se, but it's come up a couple of times now as a source of confusion for adopters. Given that most of our users are internal at this point, I thought maybe it was simplest to specify the version at the cost of being redundant for external users.

I think going with `bundle add` is good, that way there isn't confusion / people not updating if we ever do bump the major version.

etiennebarrie (approved):
> I think going with `bundle add` is good, that way there isn't confusion / people not updating if we ever do bump the major version.

I'm not sure I understand "people not updating", but to be clear, this adds the version requirement to the Gemfile (i.e. `~> 1.0`). But I think it's ok anyway, `bundle outdated` which shows new versions even if they don't match the requirement, dependabot also updates requirements by default, so I don't think people will get stuck in 1.0 because of that.

adrianna-chang-shopify:
Ah okay great, I was indeed thinking that dependabot might not update if a major version is released (although given `bundle add` adds the version requirement to the Gemfile anyways, not sure why I thought they would behave differently 🤦‍♀️ ). I think this is still better, it prevent us from having to change the version in the README whenever we do a major version release, and it makes things a bit simpler by removing the extra `bundle` line.

Issue #298, opened 2021-01-13
Leverage JobIteration's ThrottleEnumerator


adrianna-chang-shopify:
On one of the apps that has been trying out this gem in production, we noticed a big spike in MySQL queries that corresponded with a task starting to run.We should make use of [Job Iteration's ThrottleEnumerator](https://github.com/Shopify/job-iteration/blob/master/lib/job-iteration/throttle_enumerator.rb) in `TaskJob`, allowing users to specify (likely at the level of a Task) the conditions under which to throttle, and maybe even offer a customized backoff.

See https://github.com/Shopify/job-iteration/blob/master/guides/throttling.md for more information.

PR #330, opened 2021-02-10
[Explore] Allow Tasks to specify throttling

adrianna-chang-shopify:
For #298 

Here's how we could implement throttling. Users could write a Task like this:

```ruby
module Maintenance
  class MyThrottlingTask < MaintenanceTasks::Task
    throttle_on -> { DatabaseStatus.healthy? }, backoff: 5.minutes

    def collection
       User.all
    end

    def count
      collection.count
    end

    def process(user)
      # Do something
    end
  end
end
```

If we're good with this API, I'll add some docs and clean this up a bit - if not, let's discuss 😁 
I think it makes sense to specify throttling on the Task instead of on a custom Job, since users should have to think about Jobs as little as possible. I also feel like users should be able to define throttling per Task, rather than having a single throttle condition across all their Tasks (which would be the case if we required throttling to be specified on the Job instead). The downside to this is that details of the job are not available, so if users wanted to use any data from the job itself to determine whether it should throttle, this wouldn't be possible.

I've built this on top of Sam's branch, since [his PR](https://github.com/Shopify/maintenance_tasks/pull/326) extracts all our enumerator-building logic into the `Task` class instead of having it in `TaskJob`, so throttling needs to work this those two classes accordingly.

### DIFF
--- a/app/jobs/maintenance_tasks/task_job.rb
+++ b/app/jobs/maintenance_tasks/task_job.rb
@@ -32,7 +32,15 @@ module MaintenanceTasks
         cursor: cursor || @run.cursor,
       )
 
-      @task.enumerator_builder.enumerator(context: context)
+      if @task.throttle_condition
+        JobIteration::EnumeratorBuilder.new(self).build_throttle_enumerator(
+          @task.enumerator_builder.enumerator(context: context),
+          throttle_on: @task.throttle_condition,
+          backoff: @task.throttle_backoff
+        )
+      else
+        @task.enumerator_builder.enumerator(context: context)
+      end
     end
 
     # Performs task iteration logic for the current input returned by the
--- a/app/jobs/maintenance_tasks/task_job.rb
+++ b/app/jobs/maintenance_tasks/task_job.rb
@@ -25,21 +25,14 @@ module MaintenanceTasks
 
     private
 
-    def build_enumerator(_run, cursor:)
-      cursor ||= @run.cursor
-      collection = @task.collection
+    EnumerationContext = Struct.new(:cursor, keyword_init: true)
 
-      case collection
-      when ActiveRecord::Relation
-        enumerator_builder.active_record_on_records(collection, cursor: cursor)
-      when Array
-        enumerator_builder.build_array_enumerator(collection, cursor: cursor)
-      when CSV
-        JobIteration::CsvEnumerator.new(collection).rows(cursor: cursor)
-      else
-        raise ArgumentError, "#{@task.class.name}#collection must be either "\
-          'an Active Record Relation, Array, or CSV.'
-      end
+    def build_enumerator(_run, cursor:)
+      context = EnumerationContext.new(
+        cursor: cursor || @run.cursor,
+      )
+
+      @task.enumerator_builder.enumerator(context: context)
     end
 
     # Performs task iteration logic for the current input returned by the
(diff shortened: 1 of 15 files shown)


app/jobs/maintenance_tasks/task_job.rb
```diff
@@ -25,20 +25,21 @@ def retry_on(*, **)
...
+      if @task.throttle_condition
+        JobIteration::EnumeratorBuilder.new(self).build_throttle_enumerator(
+          @task.enumerator_builder.enumerator(context: context),
+          throttle_on: @task.throttle_condition,
+          backoff: @task.throttle_backoff
+        )
       else
-        raise ArgumentError, "#{@task.class.name}#collection must be either "\
-          'an Active Record Relation, Array, or CSV.'
+        @task.enumerator_builder.enumerator(context: context)
```
sambostock, line 42: We could extract the wrapping, that way you wouldn't need to repeat the initial enumerator creation.

```ruby
enumerator = @task.enumerator_builder.enumerator(context: context)

if @task.throttle_condition
  enumerator = JobIteration::EnumeratorBuilder.new(self).build_throttle_enumerator(
    enumerator,
    throttle_on: @task.throttle_condition,
    backoff: @task.throttle_backoff,
  )
end

enumerator
```

app/tasks/maintenance_tasks/task.rb
```diff
@@ -65,6 +76,18 @@ def count
...
+      #
+      # @param condition [Proc] defines the condition under which the Task
+      #   should be throttled.
+      # @param backoff [ActiveSupport::Duration] optionally, a custom backoff
+      #   can be specified. This is the time to wait before retrying the Task.
+      #   If no value is specified, it defaults to 30 seconds.
+      def throttle_on(condition, backoff: 30.seconds)
+        @throttle_condition = condition
+        @throttle_backoff = backoff
+      end
```
sambostock, line 89: Is it possible you might have more than one throttle condition?

For instance, if I've got a task dealing with some API and with the DB, maybe it would be nice to be able to do something like

```ruby
throttle_on -> { DB.unhealthy? }
throttle_on -> { API.near_limit? }
```

You could store the pairs of conditions and backoffs in an array, and then instead of wrapping the enumerator `if @task.throttle_on`, you could use a loop

```ruby
# assume throttle_spec = [] in initialize, and appended to in throttle_on

@task.throttle_specs.reduce(
  @task.enumerator_builder.enumerator(context: context),
) do |enumerator, {condition:, backoff:}|
  JobIteration::EnumeratorBuilder.new(self).build_throttle_enumerator(
    enumerator,
    throttle_on: condition,
    backoff: backoff,
  )
end
```

This would allow Task superclasses to specify throttles without worry that a child class might clobber it. For instance, if a consumer creates an intermediate class that enumerates over an API, they could throttle on the API limit, and then a child class using the DB could add another throttle, with both being respected.

app/jobs/maintenance_tasks/task_job.rb
adrianna-chang-shopify: Yes, definitely!

app/tasks/maintenance_tasks/task.rb
adrianna-chang-shopify: Ooh, that's a great point Sam! Yeah, supplying multiple conditions for throttling was something I was thinking about initially, but I ended up sticking with a single condition because I assumed that doing something like this
```ruby
throttle_on -> { DB.unhealthy? }
throttle_on -> { API.near_limit? }
```
could just be achieved by combining them:
```ruby
throttle_on -> { DB.unhealthy? || API.near_limit? }
```
But the superclass scenario you described is indeed a perfect example of why registering multiple conditions for throttling would be nice ❤️  I'll extend the PR 😄

adrianna-chang-shopify:
Closing in favour of https://github.com/Shopify/maintenance_tasks/pull/400

PR #336, opened 2021-02-16
Explicitly include view helpers

shouichi:
Problem:

When `config.action_controller.include_all_helpers = false`, helpers are
not available by default which leads to `undefined method 'status_tag'`.

Solution:

Include view helpers explicitly.

### DIFF
--- a/app/controllers/maintenance_tasks/application_controller.rb
+++ b/app/controllers/maintenance_tasks/application_controller.rb
@@ -7,6 +7,9 @@ module MaintenanceTasks
   class ApplicationController < ActionController::Base
     include Pagy::Backend
 
+    helper ApplicationHelper
+    helper TaskHelper
+
     BULMA_CDN = 'https://cdn.jsdelivr.net'
 
     content_security_policy do |policy|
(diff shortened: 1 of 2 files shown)


shouichi:
I signed the CLA but the status is not updated, how can I re-run the check?

Thanks for the nice gem by the way 😃

etiennebarrie (commented):
Thank you so much for the contribution! 

I think we can avoid explicitly including the helper modules by naming them correctly, see the review comments.

Thanks for taking care of this case, I think it should even be the default for the dummy app for a Rails engine tests. 🤔 

I got the CLA to pass green by clicking some button in the Checks tab I think, not sure if you could have done the same.

shouichi:
Thanks for the review! So the task would be

- Set `config.action_controller.include_all_helpers = false` to the test dummy app.
- Rename TaskHelper to TasksHelper

Correct?

etiennebarrie:
Yes I think that should be enough to show the bug and prevent us from re-introducing it and fixing it at the same time.

shouichi:
Cool, I'll recreate a PR.

PR #337, opened 2021-02-16
Follow the rails naming convention

shouichi:
Problem:

When `config.action_controller.include_all_helpers = false`, helpers are
not available by default which leads to `undefined method 'status_tag'`.

Solution:

By following the rails naming convention, necessary helpers are
automatically included by rails.


shouichi:
This PR aims to solve the same problem of https://github.com/Shopify/maintenance_tasks/pull/336 but it solves the problem by following the rails naming convention. Thanks to the kind help from @etiennebarrie 😃

adrianna-chang-shopify (approved):
Thanks for the contribution! ❤️

etiennebarrie (approved):
I just tweaked the comment a bit! Thank you very much!

PR #340, opened 2021-02-19
Ensure cursor is converted into integer as needed

sambostock:
To ensure #339's change to string cursors can occur smoothly, we ensure we always convert the cursor into an integer when needed.

Specifically, the `Array` and `CSV` cursors **must** be integers when forwarded to JobIteration's enumerator builders, or things break. The ActiveRecord enumerator builder handles the conversion gracefully.

This should allow consumers to perform the cursor schema change from `bigint` to `string` without worrying about stale code blowing up.

## ⚠️ After Merging
- [ ] Ensure this is released before #339, so consumers can update to the transition version

### DIFF
--- a/test/jobs/maintenance_tasks/task_job_test.rb
+++ b/test/jobs/maintenance_tasks/task_job_test.rb
@@ -198,6 +198,15 @@ module MaintenanceTasks
       TaskJob.perform_now(@run)
     end
 
+    # Remove this once string cursors are shipped (#339)
+    test '.perform_now handles string cursor when Array task resumes' do
+      @run.expects(:cursor).returns('0')
+
+      TaskJob.perform_now(@run)
+
+      assert_predicate @run, :succeeded?
+    end
+
     test '.perform_now accepts Active Record Relations as collection' do
       Maintenance::TestTask.any_instance.stubs(collection: Post.all)
       Maintenance::TestTask.any_instance.expects(:process).times(Post.count)
@@ -205,6 +214,16 @@ module MaintenanceTasks
       TaskJob.perform_now(@run)
 
       assert_predicate @run.reload, :succeeded?
+    end
+
+    # Remove this once string cursors are shipped (#339)
+    test '.perform_now handles string cursor when Active Record Relation task resumes' do
+      run = Run.create!(task_name: 'Maintenance::UpdatePostsTask')
+      run.expects(:cursor).returns(Post.pick(:id)&.to_s)
+
+      TaskJob.perform_now(run)
+
+      assert_predicate run, :succeeded?
     end
 
     test '.perform_now accepts CSVs as collection' do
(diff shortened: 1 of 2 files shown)


sambostock:
Hopefully I'm right about this allowing actively running jobs to handle the schema change. The thing I'm unsure about is if there would be any issues with ActiveRecord not detecting the change while the host app is running and running into issues with improperly constructed queries... 🤔

test/jobs/maintenance_tasks/task_job_test.rb
```diff
@@ -207,6 +216,16 @@ class TaskJobTest < ActiveJob::TestCase
       assert_predicate @run.reload, :succeeded?
     end
 
+    # Remove this once string cursors are shipped (#339)
+    test '.perform_now handles string cursor when Active Record Relation task resumes' do
+      run = Run.create!(task_name: 'Maintenance::UpdatePostsTask')
+      run.expects(:cursor).returns(Post.pick(:id)&.to_s)
```
adrianna-chang-shopify, line 222: So you were entirely right about Post fixtures not getting loaded in these tests 😂  Apparently Rails doesn't support loading fixtures from both the engine + the dummy. I'm gonna put up a PR to create some Post records for these tests.

adrianna-chang-shopify (approved):
These changes LGTM!

> Hopefully I'm right about this allowing actively running jobs to handle the schema change. The thing I'm unsure about is if there would be any issues with ActiveRecord not detecting the change while the host app is running and running into issues with improperly constructed queries...

I need to tophat further, but I think there might be a couple problematic scenarios here and I'd like to test with MySQL + PostgresQL. I'm going to take a look and make sure we have all our bases covered. Don't think there's any harm in shipping this, I think we'll want to make sure everything is 🆗   before shipping #339 though.

test/jobs/maintenance_tasks/task_job_test.rb
adrianna-chang-shopify: @sambostock I attempted to just move the `runs.yml` fixture into the dummy's fixtures and change the path in the test helper, but Mocha went 💥  😅 . I instead decided to go the route of manually creating Posts for any tests that needed them, and actually realized that it's really only this test:
https://github.com/Shopify/maintenance_tasks/blob/4ae6699466cc726dd73906fd3711171534786e63/test/jobs/maintenance_tasks/task_job_test.rb#L219-L226
that needs Posts to be created to be truly testing what we expect. Even our system tests are designed to test interactions with the web interface and actually don't need any Post data to operate on to be considered correct.

I think it just makes sense to create some Posts inline for any test that needs them (this test here and the other Active Record one). We might be able to go straight to #339 anyways, so no need to update this, just an update 😄  (Our OG ActiveRecord test should be fixed, so thanks again for catching this)

adrianna-chang-shopify:
Closing, we can go straight to #339

PR #343, opened 2021-02-22
Change generator `--csv` to `--type=csv`

sambostock:
## TL;DR

```diff
-$ rails generate maintenance_tasks:task NAME --csv
+$ rails generate maintenance_tasks:task NAME --type=csv # or --type csv
```

## Details

Working on #326 and wanting to include a generator for tasks with a custom enumerator builder, @adrianna-chang-shopify and I decided it might be best to change the `--csv` generator option to `--type=csv`, rather than have a bunch of mutually exclusive options (e.g. `--csv` & `--custom`).

IF we decide to provide additional built-in enumerators in the future, this would also provide a natural extension point.

Included in this PR are various refactorings to the generator code.

This is technically a breaking change to the developer API, but not to the actual task/application code. As such, it doesn't feel like it warrants a major version bump.

### ⚠️ Before Merging
- [x] Get feedback on task type names (`--type=generic` & `--type=csv`)
- [x] Update README and any other documentation mentioning `--csv`

### DIFF
--- a/lib/generators/maintenance_tasks/task_generator.rb
+++ b/lib/generators/maintenance_tasks/task_generator.rb
@@ -9,23 +9,28 @@ module MaintenanceTasks
     desc 'This generator creates a task file at app/tasks and a corresponding '\
       'test.'
 
-    class_option :csv, type: :boolean, default: false,
-      desc: 'Generate a CSV Task.'
+    TASK_TYPES = ['generic', 'csv']
+    class_option :type, type: :string, default: 'generic',
+      desc: "Specify the type of Task to generate (#{TASK_TYPES.join(', ')})"
 
     check_class_collision suffix: 'Task'
 
+    # Ensure a valid task type has been provided
+    def validate_task_type
+      return if TASK_TYPES.include?(task_type)
+
+      raise(Thor::Error, "Unknown task type #{task_type.inspect}. " \
+            "Must be one of: #{TASK_TYPES.join(', ')}")
+    end
+
     # Creates the Task file.
     def create_task_file
-      template_file = File.join(
+      task_file = File.join(
         "app/tasks/#{tasks_module_file_path}",
         class_path,
         "#{file_name}_task.rb"
       )
-      if options[:csv]
-        template('csv_task.rb', template_file)
-      else
-        template('task.rb', template_file)
-      end
+      template(task_template_file, task_file)
     end
 
     # Creates the Task test file, according to the app's test framework.
@@ -44,25 +49,52 @@ module MaintenanceTasks
     private
 
     def create_task_test_file
-      template_file = File.join(
+      test_file = File.join(
         "test/tasks/#{tasks_module_file_path}",
         class_path,
         "#{file_name}_task_test.rb"
       )
-      template('task_test.rb', template_file)
+      template(test_template_file, test_file)
     end
 
     def create_task_spec_file
-      template_file = File.join(
+      spec_file = File.join(
         "spec/tasks/#{tasks_module_file_path}",
         class_path,
         "#{file_name}_task_spec.rb"
       )
-      template('task_spec.rb', template_file)
+      template(spec_template_file, spec_file)
+    end
+
+    def task_template_file
+      case task_type
+      when 'generic'
+        'task.rb'
+      when 'csv'
+        'csv_task.rb'
+      end
+    end
+
+    def test_template_file
+      case task_type
+      when 'generic', 'csv'
+        'task_test.rb'
+      end
+    end
+
+    def spec_template_file
+      case task_type
+      when 'generic', 'csv'
+        'task_spec.rb'
+      end
     end
 
     def file_name
       super.sub(/_task\z/i, '')
+    end
+
+    def task_type
+      options[:type]
     end
 
     def tasks_module
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ is specified in your application's Gemfile, and that you've followed the
 Generate a CSV Task by running:
 
 ```bash
-$ rails generate maintenance_tasks:task import_posts --csv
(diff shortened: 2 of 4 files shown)


lib/generators/maintenance_tasks/task_generator.rb
```diff
@@ -9,23 +9,29 @@ class TaskGenerator < Rails::Generators::NamedBase
     desc 'This generator creates a task file at app/tasks and a corresponding '\
       'test.'
 
-    class_option :csv, type: :boolean, default: false,
-      desc: 'Generate a CSV Task.'
+    TASK_TYPES = ['collection', 'csv'].freeze
```
adrianna-chang-shopify, line 12: I don't think `.freeze` is necessary 😄

lib/generators/maintenance_tasks/task_generator.rb
```diff
@@ -9,23 +9,29 @@ class TaskGenerator < Rails::Generators::NamedBase
...
+    private_constant :TASK_TYPES
```
adrianna-chang-shopify, line 13: Any reason this needs to be private in addition to the class being private?

adrianna-chang-shopify (commented):
A couple small comments, looks great though!

I am having second thoughts about using `collection` as the default type. (Since `--type=csv` means a collection of type CSV, it's a bit odd to have `--type=collection`). I wonder if it should just be `default` or `generic` instead...

lib/generators/maintenance_tasks/task_generator.rb
sambostock: Habit, since it's really an implementation detail. I can remove it if it's inappropriate.

lib/generators/maintenance_tasks/task_generator.rb
sambostock: I'm just used to freezing all mutable constants (so they're actually "constant"). Do you see a use case for modifying it?

sambostock:
@adrianna-chang-shopify I've
- removed `private_constant :TASK_TYPES`
- changed `--type=collection` to `--type=generic`
- changed `--csv` to `--type=csv` in `README.md`

lib/generators/maintenance_tasks/task_generator.rb
adrianna-chang-shopify: The argument I've heard from Rafael & others is that using `.freeze` is defensive and that we should assume developers are well-intended with the code they write instead of protecting against something that they likely won't do. The fact that it only warns instead of raising if you redefine the constant means that we're not really protected from mutation anyways. We haven't frozen any of our other constants in the gem for this reason 😄

adrianna-chang-shopify (approved):
I would drop `.freeze` on `TASK_TYPES` still to avoid defensive programming, but other than that this looks great! ❤️

adrianna-chang-shopify:
@sambostock @etiennebarrie how would you guys feel about starting off a feature branch for custom enumeration and shipping these incremental PRs into it so that we can release it in completion? `main` has got some patches on it currently that I'll release as a patch version soon, but I've still got a couple fixes to finish. I think it's probably best to not have bits and pieces of the custom enum stuff going out until it's ready in entirety, right?

sambostock:
The following PRs exist right now around supporting custom enumeration:

- Ensure cursor is converted into integer as needed (#340)
- Change cursor from bigint to string (#339)
- Change generator `--csv` to `--type=csv`(#343, this PR)
- Introduce Task#enumerator_builder (#326)

#340 would have to be released in a prior release to the one that includes #339, since it's concerned with forwards compatibility while the app is running, so we wouldn't want to include it in a feature branch. Though we'd have to validate #339 will work before merging #340.

#343 I think could technically ship on its own, though I can understand the value of shipping it in the same release as custom enumeration.

Usually I'm opposed to feature branches and prefer to continuously integrate small changes. For instance, we could always ship part of the custom enumerator change as an internal refactor that doesn't break anything.

That said, I've done most of my development for apps, not libraries/gems, so maybe this is different. If we think there's value in a feature branch here, I'm glad to do it. 🤷

adrianna-chang-shopify:
I think we're okay to just ship incrementally in that case 👍  I was recalling a previous experience working on the Packwerk gem where a fix had to be shipped ASAP and there was a feature that was half-functional on the main branch and caused all sorts of headaches 😅  But you're right that the most of these changes are internal refactors that should be OK to ship without fear of having incomplete functionality on main.
This one (#343) is totally fine to go out on its own IMO. I was thinking maybe we'd release it as a minor version (since it is a change in the developer API) after releasing a patch (v1.1.1) but I suppose it's okay just to ship all our bug fixes to 1.2 instead.
#340 can go in with that as well, since it's just an internal refactor.

#339, #326 and a PR to add `custom` to our Task Generator can all go out together as the final piece of custom enumeration, once we're sure everything will work and our bases are covered in terms of backwards compatibility + making sure nothing breaks when changing the column type.

So TL;DR ignore my feature branch request, let me see if I can get the patch version out by the end of the day - if not, I think we can just start merging some of these and go straight to 1.2.

adrianna-chang-shopify:
@etiennebarrie you'd mentioned preferring to keep `--csv` and support only `--csv` and `--custom` as mutually exclusive options moving forward. I believe that this PR gives us more flexibility, but agree that it might be premature to assume we'll have other Task types beyond CSV and the new custom enumeration stuff. 

Can you 👍  / 👎  this so we either ship or close?

etiennebarrie:
Yeah I'm not really in favor of this PR because:
* I'd like to keep `--csv` around
* `--type=generic` can just be the default, no need to add a name for that
* whenever we introduce `--enumerator` (`custom` is a bit unclear I find), we can just deal with the fact that they're exclusive
* I don't think we're going to need more than that

sambostock:
Fair enough

Issue #334, opened 2021-02-15
Hooks for when a job starts and completes?


tgwizard:
What do you think about adding custom hooks, similar to `MaintenanceTasks.error_handler`, to allow the hosting service to be notified when a new maintenance job is started or completed? Similar to the `start` and `complete` callbacks from job-iteration.

The main use-case I have in mind is to send a notification to Slack whenever a task is started or completes. We have similar Slack posts for other system changes (feature rollout changes, DB migrations, etc.)

PR #344, opened 2021-02-23
Avoid usage of private_constant

rafaelfranca:
This just make hard to use reflection on Ruby classes breaking tools like tapioca. The value we want to provide with this call is telling people to not use this class, but that is already achieved by `@api private`.

If people really want to use this constant even if we told them to not do it, isn't `private_constant` that will stop them.

### DIFF
--- a/app/controllers/maintenance_tasks/application_controller.rb
+++ b/app/controllers/maintenance_tasks/application_controller.rb
@@ -3,7 +3,7 @@
 module MaintenanceTasks
   # Base class for all controllers used by this engine.
   #
-  # @api private
+  # Can be extended to add different authentication and authorization code.
   class ApplicationController < ActionController::Base
     include Pagy::Backend
 
--- a/app/controllers/maintenance_tasks/application_controller.rb
+++ b/app/controllers/maintenance_tasks/application_controller.rb
@@ -22,5 +22,4 @@ module MaintenanceTasks
 
     protect_from_forgery with: :exception
   end
-  private_constant :ApplicationController
 end
--- a/app/controllers/maintenance_tasks/runs_controller.rb
+++ b/app/controllers/maintenance_tasks/runs_controller.rb
@@ -30,5 +30,4 @@ module MaintenanceTasks
       @run = Run.find(params.fetch(:id))
     end
   end
-  private_constant :RunsController
 end
--- a/app/controllers/maintenance_tasks/tasks_controller.rb
+++ b/app/controllers/maintenance_tasks/tasks_controller.rb
@@ -42,5 +42,4 @@ module MaintenanceTasks
       @refresh = 3
     end
   end
-  private_constant :TasksController
 end
--- a/app/helpers/maintenance_tasks/application_helper.rb
+++ b/app/helpers/maintenance_tasks/application_helper.rb
@@ -29,5 +29,4 @@ module MaintenanceTasks
       end
     end
   end
-  private_constant :ApplicationHelper
 end
--- a/app/helpers/maintenance_tasks/tasks_helper.rb
+++ b/app/helpers/maintenance_tasks/tasks_helper.rb
@@ -114,5 +114,4 @@ module MaintenanceTasks
       )
     end
   end
-  private_constant :TasksHelper
 end
--- a/app/models/maintenance_tasks/application_record.rb
+++ b/app/models/maintenance_tasks/application_record.rb
@@ -6,5 +6,4 @@ module MaintenanceTasks
   class ApplicationRecord < ActiveRecord::Base
     self.abstract_class = true
   end
-  private_constant :ApplicationRecord
 end
--- a/app/models/maintenance_tasks/application_record.rb
+++ b/app/models/maintenance_tasks/application_record.rb
@@ -2,7 +2,8 @@
 module MaintenanceTasks
   # Base class for all records used by this engine.
   #
-  # @api private
+  # Can be extended to setup different database where all tables related to
+  # maintenance tasks will live.
   class ApplicationRecord < ActiveRecord::Base
     self.abstract_class = true
   end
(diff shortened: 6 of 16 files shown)


adrianna-chang-shopify (commented):
(closes https://github.com/Shopify/maintenance_tasks/issues/314)

I'm 👍  to avoiding `private_constant`, is there a reason this PR is only removing `private_constant` for `ApplicationRecord` though and not all our constants?

rafaelfranca:
> I'm 👍 to avoiding `private_constant`, is there a reason this PR is only removing `private_constant` for `ApplicationRecord` though and not all our constants?

Not really. This was breaking tapioca RBI generation, that is why I opened only for it, but I can remove all of them.

etiennebarrie:
Sounds like a bug in tapioca to me, even if we remove its usage here, we will always find a gem somewhere that uses `private_constant`. 🤷

rafaelfranca:
> Sounds like a bug in tapioca to me, even if we remove its usage here, we will always find a gem somewhere that uses `private_constant`. 🤷

Yes, I agree, and we are fixing this bug.

But I don't think this gem, or any ruby code really, should be using `private_constant` as API visibility. It is just like defining all method that we don't want users to use `private` and we calling `send` everywhere in our code because we want to use.

In my opinion `private_constant` is not protecting from anything, that `@api private` is not already doing.

adrianna-chang-shopify (approved):

etiennebarrie:
> It is just like defining all method that we don't want users to use `private` and we calling `send` everywhere in our code because we want to use

It's not, we don't have to use const_get ourselves to get to our private constants.

> In my opinion `private_constant` is not protecting from anything that `@api private` is not already doing.

It's not about opinion though, it is protecting against accessing a constant without acknowledging that it is private by having to use `const_get`, which `@api private` doesn't do, unless we had some static typing tool that used this declaration.

I'm not against removing them if they're blocking, like it is for tapioca, though in that case I think we should probably have a test generating the rbi with tapioca, the same way we have one testing generating the documentation.

rafaelfranca:
The point is, do we really need to protect against it?

I'm not removing because it is blocking something. I removing because I don't think it is bringing us or our users anything. If people can use `const_get` like you said, `private_constant` is not protecting against anything, just creating more trouble for users. If they want to use something that we clearly marked as private API let them use. There is no need to also require them to use `const_get`.

rafaelfranca:
Ah, not to say that

```ruby
module MaintenanceTask
  ApplicationController.include(Something)
end
```

Will also work without problem. No need for `const_get`, no "acknowledging that it is private". Making users write unnatural code just to tell them something that we already told in the documentation is unnecessary, in my opinion.

And before you bring "there is no opinion", software is opinion, and we write opinionated tools.

sambostock:
@rafaelfranca out of curiosity, are there any use cases where you think `private_constant` is appropriate?

----

Personally, I think there's value in the constant not showing up when introspecting `.constants`:

```ruby
module Foo
  BAR = 123
  BAZ = 456
  private_constant :BAZ
end
Foo.constants # => [:BAR]
```

It's also nice for users of `pry` when using `ls`:
```ruby
[2] pry(main)> ls Foo
constants: BAR
```

Neither of these cases "honour" `@api private`.

paracycle:
> ... out of curiosity, are there any use cases where you think `private_constant` is appropriate?

@sambostock Not Rafael here, but I also have an opinion on this. The way I see it, `private_constant` is appropriate if the constant is either a value constant (i.e. not a class/module) or if its usage is totally contained within the surrounding context.

In the example you give `BAZ` is a value constant and I think it is fine for that to be private. If `BAZ` was a temp module/class that was necessary for just the `Foo` scope, then, in my view, that would also be appropriate.

The bigger problem with making class/module constants private is that they inevitably leak via the public API of your gem. For example, if you return a value `val` of the type of a private constant, then people can always do `NewSubclass = Class.new(val.class)` and subclass from the private constant. I am not saying that this is what this gem is doing but it is likely that might be the case down the line.

We had that exact same tension when building Tapioca. We felt like we shouldn't be generating private constants, but then also sometimes some public type ended up including a private constant or subclassed from it. That is why I will relax Tapioca so that it generates definitions for all constants regardless of visibility.

Anyway, that's my 2 cents.

etiennebarrie:
Writing unnatural code like this is another way of acknowledging that it is private. I can clearly see someone reviewing code written like that and asking "Why is it re-opening the module?" "Because the constant is private".

You're right that there is an opinion, mine is that people don't necessarily read all the documentation and especially reviewers don't necessarily double check that everything used is a public API. I think the core goal for me is to lower my worry of changing things in the future, I feel like with private constants I get the peace of mind that I did all I could.

I guess I should probably be more comfortable taking the risk of breaking people stuff, hoping they read the documentation. So I could get behind removing them, even though I thought a brand new gem was the occasion to start cleanly on that point.

Regarding removing the `@api private` to ApplicationController, I'm less sure, because that raises my worries. Maybe they will add permission based on the controller/action, and prevent running tasks based on `tasks#run` for example, and then we will change it to `runs#create` and silently break their stuff, whereas if they talk to us we can build what they need, and they can always do what they do with sidekiq, which is adding authentication at the router level with constraints.

ApplicationRecord, I think it's a good idea, they could use the `connects_to` macro, change the table_name_prefix or suffix. So I feel like it's justified by the needs. They could also add callbacks which seems like it could break things but that seems more far-fetched, we already have one case of monkey-patching to support things like #334 and they went straight for the Run even if it was private.

etiennebarrie (approved):
Given that it's breaking tapioca, I think it's fair to remove the `private_constant` calls. But in the future I'd like to bring them back, which shouldn't be a problem they're private API and we're free to rename, hide them if we want.

Regarding ApplicationController and ApplicationRecord, it makes sense.

Issue #335, opened 2021-02-15
Race condition when a task is interrupted leads to exceptions


etiennebarrie:
First Job | Second Job | Run status 
------------ | ------------- | --
Task is created, job enqueued in Runner.run |  | enqueued
TaskJob#before_enqueue | | running
Job is re-enqueued in JobIteration::Iteration#reenqueue_iteration_job | | 
~ | TaskJob#before_enqueue | running
TaskJob#after_perform | | interrupted
~ | Task is processing | interrupted

At this point you have a running Task which a job is actively processing, so in the UI it's a bit weird because you can see progress being made to a seemingly interrupted Task.
An exception is then raised when the Task succeeds, because we try to move the status from interrupted to succeeded. 

Before 1.1.0, this exception is caught and another one is raised when we try to move from interrupted to errored.

What happened next with 1.0.0 and sidekiq is that the job was retried because there was an unhandled exception (with the previous cursor as explained in the README). That finally led the task to finish and succeed. Re-using the last cursor meant that elements were processed multiple times, luckily the process was idempotent.

With 1.1.0, the exception is caught, and the task is moved to errored thanks to the second commit of #293, which was meant for deleted tasks but applies here as well: https://github.com/Shopify/maintenance_tasks/pull/293/commits/dbe323439c312c3dfbaf7324bb7515e763081d57.
The only solution at this point is to run the task again, which may skip the already processed items with the `collection` method, or last line of defense hopefully the `process` method is idempotent.

adrianna-chang-shopify:
Okay, so if I understand correctly, the current situation with 1.1 is this:
- A Task can get interrupted and reenqueued so fast that you have an "interrupted" Task that is actually running, because the status gets set to running when it is starts performing again, and then is set to interrupted afterwards when the initial job goes through it's shutdown callback
- When the Task finishes successfully, it will get set to "succeeded", which is invalid
- This triggers a transition from "interrupted" -> "errored", which is now valid in 1.1

So, with 1.1 we won't run into the issue of Sidekiq retrying the job because of an unhandled exception, because we're rescuing appropriately now. But, we may have missed some iterations because the cursor was not correct when the job was reenqueued (since this happened before the shutdown callback).

Can you clarify?
>The only solution at this point is to run the task again, which may skip the already processed items with the collection method

[...]

PR #345, opened 2021-02-25
Postpone reenqueuing the iteration job until after callbacks are done

adrianna-chang-shopify:
For: https://github.com/Shopify/maintenance_tasks/issues/335

This PR fixes a bug that occurs when a job gets interrupted and is reenqueued faster than the original job takes to shut down. This results in race conditions with the `@run` and results in errors due to invalid status transitions (and notably results in a run showing up as interrupted when it is actually running).

To solve this, we should delay reenqueuing the job until after all callbacks have completed, which can be done by calling `#reenqueue_iteration_job` in a prepended after_perform callback in TaskJob, and skipping the call that JobIteration performs in [`#iterate_with_enumerator`](https://github.com/Shopify/job-iteration/blob/ae9f11e316c99eaf0c5a478c1fa0a659fcd7730a/lib/job-iteration/iteration.rb#L131).

## Considerations
Arguably callback ordering and ensuring that a job shuts down and runs its callbacks successfully before the new one starts up is something that should be upstreamed to `JobIteration`. I experimented with [a PR](https://github.com/Shopify/job-iteration/pull/66) for this, but the issue is that a number of jobs relying on JobIteration in Core are also dependent on the point at which the job is reenqueued in order to resume batch processing jobs where an error has occurred. (These jobs are currently able to resume from the last cursor when something goes wrong, but changing the order results in the [new job not being pushed back to the queue](https://buildkite.com/shopify/shopify-selective-tests/builds/337355), and things simply failing).

I think this is worth reinvestigating upstream at a later point, but I think we should get this patch out in the meantime, given it's affecting a number of users at the moment.

## Tophatting
Try without the changes and then with to see the difference.
Steps to reproduce:
- `bundle add sidekiq`
```ruby
+++ test/dummy/config/application.rb
     # Application configuration can go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded after loading
     # the framework and any gems in your application.
+    config.active_job.queue_adapter = :sidekiq
   end
 end
```
```ruby
+++ app/jobs/maintenance_tasks/task_job.rb
    def on_shutdown
       if @run.cancelling?
         @run.status = :cancelled
         @run.ended_at = Time.now
       else
+        sleep(3)
         @run.status = @run.pausing? ? :paused : :interrupted
```

```ruby
+++ test/dummy/config/initializers/maintenance_tasks.rb
+JobIteration.max_job_runtime = 4.seconds
```

And then run `Maintenance::UpdatePostsTask` from the UI

### DIFF
--- a/app/jobs/maintenance_tasks/task_job.rb
+++ b/app/jobs/maintenance_tasks/task_job.rb
@@ -97,8 +97,16 @@ module MaintenanceTasks
       @ticker.persist
     end
 
+    def reenqueue_iteration_job(should_ignore: true)
+      super() unless should_ignore
+      @reenqueue_iteration_job = true
+    end
+
     def after_perform
       @run.save!
+      if defined?(@reenqueue_iteration_job) && @reenqueue_iteration_job
+        reenqueue_iteration_job(should_ignore: false)
+      end
     end
 
     def on_error(error)
--- a/app/jobs/maintenance_tasks/task_job.rb
+++ b/app/jobs/maintenance_tasks/task_job.rb
@@ -97,6 +97,17 @@ module MaintenanceTasks
       @ticker.persist
     end
 
+    # We are reopening a private part of Job Iteration's API here, so we should
+    # ensure the method is still defined upstream. This way, in the case where
+    # the method changes upstream, we catch it at load time instead of at
+    # runtime while calling `super`.
+    unless private_method_defined?(:reenqueue_iteration_job)
+      error_message = <<~HEREDOC
+        JobIteration::Iteration#reenqueue_iteration_job is expected to be
+        defined. Upgrading the maintenance_tasks gem should solve this problem.
+      HEREDOC
+      raise error_message
+    end
     def reenqueue_iteration_job(should_ignore: true)
       super() unless should_ignore
       @reenqueue_iteration_job = true
--- a/lib/maintenance_tasks.rb
+++ b/lib/maintenance_tasks.rb
@@ -9,6 +9,10 @@ require 'maintenance_tasks/engine'
 require 'pagy'
 require 'pagy/extras/bulma'
 
+# Force the TaskJob class to load so we can verify upstream compatibility with
+# the JobIteration gem
+require_relative '../app/jobs/maintenance_tasks/task_job'
+
 # The engine's namespace module. It provides isolation between the host
 # application's code and the engine-specific code. Top-level engine constants
 # and variables are defined under this module.
--- a/test/jobs/maintenance_tasks/task_job_test.rb
+++ b/test/jobs/maintenance_tasks/task_job_test.rb
@@ -164,6 +164,24 @@ module MaintenanceTasks
       assert_equal 'Task Maintenance::DeletedTask not found.', run.error_message
     end
 
+    test '.perform_now delays reenqueuing the job after interruption until all callbacks are finished' do
+      JobIteration.stubs(interruption_adapter: -> { true })
+
+      AnotherTaskJob = Class.new(TaskJob) do
+        after_perform { self.class.times_interrupted = times_interrupted }
+
+        class << self
+          attr_accessor :times_interrupted
+        end
+      end
+      AnotherTaskJob.perform_now(@run)
+
+      # The job should not yet have been reenqueued, so times_interrupted should
+      # be 0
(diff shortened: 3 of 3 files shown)


app/jobs/maintenance_tasks/task_job.rb
```diff
@@ -97,8 +97,18 @@ def on_shutdown
...
 
+    def reenqueue_iteration_job
+      if defined?(@reenqueue_iteration_job) && @reenqueue_iteration_job
+        super
+      end
+    end
+
     def after_perform
       @run.save!
+      if @run.interrupted?
```
etiennebarrie, line 108: Instead of relying on the state of the run, maybe we should just have to code only focused on delaying the enqueuing? i.e. instead of checking for the situation where the job would have be re-enqueued (`@run.interrupted?`), remember if re-enqueuing was tried when it's prevented and then later check whether it would have been re-enqueued and only re-enqueuing if it was.

app/jobs/maintenance_tasks/task_job.rb
```diff
@@ -97,8 +97,18 @@ def on_shutdown
...
+        @reenqueue_iteration_job = true
+        reenqueue_iteration_job
```
etiennebarrie, line 110: If we're keeping the logic around `@run.interrupted?`, we could simplify by not using an instance variable, just a parameter to `reenqueue_iteration_job`.

app/jobs/maintenance_tasks/task_job.rb
adrianna-chang-shopify: Sounds good, I was torn between introducing a new ivar to handle this vs using something we already had (the run status), but if we upstream later it might be better to shift this logic to be focused on the delayed enqueue instead 👍

adrianna-chang-shopify: Oh we don't even need another variable to do this, we can just set `@reenqueue_iteration_job` in `reenqueue_iteration_job` if it's not yet set, duh 🤦‍♀️

app/jobs/maintenance_tasks/task_job.rb
```diff
@@ -97,8 +97,23 @@ def on_shutdown
       @ticker.persist
     end
 
+    # We are reopening a private part of Job Iteration's API here, so we should
+    # ensure the method is still defined upstream. This way, in the case where
+    # the method changes upstream, we catch it at load time instead of at
+    # runtime while calling `super`.
+    unless private_method_defined?(:reenqueue_iteration_job)
+      raise 'JobIteration::Iteration#reenqueue_iteration_job must be defined'
```
etiennebarrie, line 105: Maybe we could tell them to upgrade maintenance_tasks, because we should have a solution to this on our side.

app/jobs/maintenance_tasks/task_job.rb
```diff
@@ -97,8 +97,23 @@ def on_shutdown
...
+    end
+    def reenqueue_iteration_job(should_ignore: true)
```
etiennebarrie, line 107: Technically we're not ignoring it, more like postponing it but ok 😅

lib/maintenance_tasks.rb
```diff
@@ -9,6 +9,10 @@
 require 'pagy'
 require 'pagy/extras/bulma'
 
+# Force the TaskJob class to load so we can verify upstream compatibility with
+# the JobIteration gem
+require_relative '../app/jobs/maintenance_tasks/task_job'
```
etiennebarrie, line 14: Maybe instead of requiring the file we could to the check here instead? But that separates it from the patch itself so it's probably worse.

etiennebarrie (approved):

app/jobs/maintenance_tasks/task_job.rb
adrianna-chang-shopify: I thought about using "postpone" instead, but then figured it's not really postponed because we call it again explicitly, it is more just outright ignored the first time 😛

app/jobs/maintenance_tasks/task_job.rb
adrianna-chang-shopify: Yes, good idea 👍

lib/maintenance_tasks.rb
adrianna-chang-shopify: Yeah I think it might be less confusing to keep them altogether for now

Issue #342, opened 2021-02-22
Issues with indexes


adrianna-chang-shopify:
While generating a new db migration, @sambostock and I noticed that the descending order was dropped from the `["task_name", "created_at"]` index when the migration ran.

Some of our adopters also seem to not have the order on the index present:
- https://github.com/Shopify/arrive-server/blob/master/db/schema.rb#L393
- https://github.com/Shopify/accio/blob/master/db/schema.rb#L192

@etiennebarrie had also flagged that the index on just `task_name` is potentially redundant - we should double check whether we need it or not.

adrianna-chang-shopify:
Some updates on this:
- Confirmed that our `task_name` index is redundant and can be dropped. Will open a PR
- Adopters not having the sort order present on the index is actually not a bug, but instead due to the fact that most of our apps are using MySQL 5.7, and ordering is only present on `MySQL 8.0+` 🤦‍♀️ 
- The sort order disappearing from SQLite3 when running a new migration is actually the result of a [bug](https://github.com/rails/rails/pull/41548) in the SQLite3 Adapter in Rails.

PR #346, opened 2021-02-25
Remove redundant index on :task_name

adrianna-chang-shopify:
For: https://github.com/Shopify/maintenance_tasks/issues/342

Confirmed that the index on just `:task_name` is redundant given our composite index on `[:task_name, :created_at]`. We can remove it.

I think we need to remove the index from the table itself in a `change_table` block (instead of using https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-remove_index) due to LHM constraints (as mentioned [here](https://github.com/Shopify/maintenance_tasks/pull/239), modifying indexes outside of table blocks seems to pose problems for LHMs).

### DIFF
--- a/db/migrate/20210225152418_remove_index_on_task_name.rb
+++ b/db/migrate/20210225152418_remove_index_on_task_name.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+class RemoveIndexOnTaskName < ActiveRecord::Migration[6.1]
+  def up
+    change_table(:maintenance_tasks_runs) do |t|
+      t.remove_index(:task_name)
+    end
+  end
+
+  def down
+    change_table(:maintenance_tasks_runs) do |t|
+      t.index(:task_name)
+    end
+  end
+end
--- a/test/dummy/db/schema.rb
+++ b/test/dummy/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_01_12_141723) do
+ActiveRecord::Schema.define(version: 2021_02_25_152418) do
 
   create_table "active_storage_attachments", force: :cascade do |t|
     t.string "name", null: false
@@ -56,7 +56,6 @@ ActiveRecord::Schema.define(version: 2021_01_12_141723) do
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
     t.index ["task_name", "created_at"], name: "index_maintenance_tasks_runs_on_task_name_and_created_at", order: { created_at: :desc }
-    t.index ["task_name"], name: "index_maintenance_tasks_runs_on_task_name"
   end
 
   create_table "posts", force: :cascade do |t|


sambostock (approved):

sambostock:
I suspect the reason #239 was able to get around LHM constraints was because it consolidated everything into a `create_table` block. I feel like `change_table` or `add_index` would both have the same problems for LHM (don't modify existing tables without using LHM). 🤔

adrianna-chang-shopify:
Oof yeah good shout @sambostock, whether we use `change_table` or not likely won't solve the LHM issue because it's still modifying the table 🤦‍♀️  Didn't think that one through very hard lol. 

Given the apps that use LHMs are a minority, we should just be able to ship this, and any app that requires use of LHMs can delete this migration and rewrite it as an LHM if need be (I think only Partners had an issue with this)?

sambostock:
While I do agree it's probably fine to ship as is and let consumers rewrite migrations as LHM, Shopify's data stores best practices recommend everyone use LHM for all table changes, so I'm not sure how uncommon it is internally. That said, the only other thing we could do would be to try to detect LHM and generate an appropriate migration, but I don't know if that makes sense.

db/migrate/20210225152418_remove_index_on_task_name.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+class RemoveIndexOnTaskName < ActiveRecord::Migration[6.1]
+  def up
+    change_table(:maintenance_tasks_runs) do |t|
```
etiennebarrie, line 4: Any reason not to use `remove_index` and `add_index` directly?

etiennebarrie (approved):

adrianna-chang-shopify:
Yeah, it's a really good point @sambostock - it's tricky to find the balance between making this meet our internal best practices vs trying to keep things conventional for use externally. Looking at the [dependency list](https://services.shopify.io/dependencies/detail?dependency=lhm) in SDB, there are a number of gems using `Shopify/lhm` 😬  But I'm not sure it's worth the complexity of having to detect and generate new LHM migrations.. Maybe a good middle point would be to document rewriting one of our migrations as an LHM in the README?

db/migrate/20210225152418_remove_index_on_task_name.rb
adrianna-chang-shopify: The only reason I could see using this over `remove_index` and `add_index` directly is that it makes it easier to change it to an LHM:

```ruby
class RemoveIndexOnTaskName < ActiveRecord::Migration[6.1]
  def up
      t.remove_index(:task_name)
    end
  end

  def down
      t.index(:task_name)
    end
  end
end
```

```ruby
require 'lhm'

class RemoveIndexOnTaskName < ActiveRecord::Migration[6.1]
  def up
    Lhm.change_table :maintenance_tasks_runs do |t|
      t.remove_index(:task_name)
    end
  end

  def down
    Lhm.change_table :maintenance_tasks_runs do |t|
      t.index(:task_name)
    end
  end
end
```

Not sure if if this is something we care about that much though 😄

etiennebarrie: Yup it's really nice if it lets people adapt it more easily to LHM.

PR #353, opened 2021-03-01
Don't show indefinite progress bar when finishing a task

etiennebarrie:
When a Task is finished processing, but the `JobIteration.max_job_runtime` has been reached, the Task gets interrupted, then when it starts it finally gets marked as succeeded, because there's no additional element to process.

When that happens, the tick count is equal to the tick total and the progress bar appears as indefinite. We were missing a test at the boundary.


adrianna-chang-shopify (approved):

adrianna-chang-shopify (approved):

etiennebarrie:
I've signed the CLA!

PR #355, opened 2021-03-01
Release v1.1.1

adrianna-chang-shopify:
Let's ship this patch version will all our 🐞  fixes.

I've updated the release notes, but let's make sure we haven't missed anything.


etiennebarrie (approved):
👌

PR #359, opened 2021-03-02
Check if ActiveStorage used before fetching Run's csv_file

adrianna-chang-shopify:
Right now, both:
https://github.com/Shopify/maintenance_tasks/blob/0c5813cd2d5acd8a4457d9ddfe6072dd9bd48ed6/app/views/maintenance_tasks/runs/_info.html.erb#L12
and
https://github.com/Shopify/maintenance_tasks/blob/0c5813cd2d5acd8a4457d9ddfe6072dd9bd48ed6/app/models/maintenance_tasks/run.rb#L191
will blow up when called in an app that does not have ActiveStorage set up.

Arguably 
https://github.com/Shopify/maintenance_tasks/blob/0c5813cd2d5acd8a4457d9ddfe6072dd9bd48ed6/app/models/maintenance_tasks/run.rb#L189
and
https://github.com/Shopify/maintenance_tasks/blob/0c5813cd2d5acd8a4457d9ddfe6072dd9bd48ed6/app/models/maintenance_tasks/runner.rb#L48
also blow up, but these are scenarios where the app is attempting to use CSV functionality, so I think it's okay that these result in the missing table error.

IMHO the way to solve this is to just return early from `Run#csv_file` if the `ActiveStorage::Attachment` table doesn't exist, and to then use the safe access operator for the first two lines above.

I figured we wouldn't want to use the safe access operator for the second two scenarios (L189 in `Run` and L48 in `Runner`), since ActiveStorage should be configured here if the user wants to write CSV Tasks, so it makes sense for this to 💥 .

EDIT: The preloading query in `TaskData` also fails if the table doesn't exist, so I've tweaked that to not preload unless it's the table exists there too.

### DIFF
--- a/app/models/maintenance_tasks/run.rb
+++ b/app/models/maintenance_tasks/run.rb
@@ -188,11 +188,21 @@ module MaintenanceTasks
     def csv_attachment_presence
       if Task.named(task_name) < CsvCollection && !csv_file.attached?
         errors.add(:csv_file, 'must be attached to CSV Task.')
-      elsif !(Task.named(task_name) < CsvCollection) && csv_file.attached?
+      elsif !(Task.named(task_name) < CsvCollection) && csv_file.present?
         errors.add(:csv_file, 'should not be attached to non-CSV Task.')
       end
     rescue Task::NotFoundError
       nil
+    end
+
+    # Fetches the attached ActiveStorage CSV file for the run. Checks first
+    # whether the ActiveStorage::Attachment table exists so that we are
+    # compatible with apps that are not using ActiveStorage.
+    #
+    # @return [ActiveStorage::Attached::One] the attached CSV file
+    def csv_file
+      return unless ActiveStorage::Attachment.table_exists?
+      super
     end
   end
 end
--- a/app/models/maintenance_tasks/run.rb
+++ b/app/models/maintenance_tasks/run.rb
@@ -41,6 +41,11 @@ module MaintenanceTasks
 
     scope :active, -> { where(status: ACTIVE_STATUSES) }
 
+    # Ensure ActiveStorage is in use before preloading the attachments
+    scope :with_attached_csv, -> do
+      with_attached_csv_file if ActiveStorage::Attachment.table_exists?
+    end
+
     validates_with RunStatusValidator, on: :update
 
     has_one_attached :csv_file
--- a/app/models/maintenance_tasks/task_data.rb
+++ b/app/models/maintenance_tasks/task_data.rb
@@ -135,7 +135,7 @@ module MaintenanceTasks
     private
 
     def runs
-      Run.where(task_name: name).with_attached_csv_file.order(created_at: :desc)
+      Run.where(task_name: name).with_attached_csv.order(created_at: :desc)
     end
   end
 end
--- a/app/views/maintenance_tasks/runs/_info.html.erb
+++ b/app/views/maintenance_tasks/runs/_info.html.erb
@@ -9,7 +9,7 @@
   <%= render "maintenance_tasks/runs/info/#{run.status}", run: run %>
 </div>
 
-<% if run.csv_file.attached? %>
+<% if run.csv_file.present? %>
   <div class="block">
     <%= link_to('Download CSV', csv_file_download_path(run)) %>
   </div>


app/models/maintenance_tasks/task_data.rb
```diff
@@ -135,7 +135,13 @@ def csv_task?
     private
 
     def runs
-      Run.where(task_name: name).with_attached_csv_file.order(created_at: :desc)
+      if ActiveStorage::Attachment.table_exists?
```
etiennebarrie, line 138: Can we keep this logic inside Run? Maybe overriding `with_attached_csv_file`?

app/models/maintenance_tasks/run.rb
```diff
@@ -188,11 +188,21 @@ def stuck?
     def csv_attachment_presence
       if Task.named(task_name) < CsvCollection && !csv_file.attached?
         errors.add(:csv_file, 'must be attached to CSV Task.')
-      elsif !(Task.named(task_name) < CsvCollection) && csv_file.attached?
+      elsif !(Task.named(task_name) < CsvCollection) && csv_file&.attached?
```
etiennebarrie, line 191: Since `ActiveStorage::Attached::One` implements `blank?`, we may be able to use `present?` instead. But the safe operator is fine otherwise.

etiennebarrie (approved):
We also don't depend on activestorage in our gemspec, which means it would fail with an app that doesn't even include the framework. We could either add it as our dependency like we do for activejob, but I think for this case it may be nice to our users to make it entirely optional. It's worth shipping this anyway as I think a lot of apps just depend on rails and have activestorage, even if they haven't installed it.

app/models/maintenance_tasks/run.rb
adrianna-chang-shopify: Sure, let's change it!

app/models/maintenance_tasks/task_data.rb
adrianna-chang-shopify: Actually overriding `with_attached_csv_file` involves overriding the scope, which complains noisily when I try to do it. I'll create our own scope instead, let me know what you think.

adrianna-chang-shopify:
@etiennebarrie Yeah I'm still open to making ActiveStorage an explicit dependency of the gem! IMHO it's kind of an unrelated conversation to this patch. Maybe it's something we don't want to force on our users (also given it's a rare occurrence that this would be an issue, since most Rails apps include the library by default anyways), but I don't think it's a problem to add it either.

I just wanted to get your opinion on https://github.com/Shopify/maintenance_tasks/pull/359/commits/3c39814fb93d7910d34948eb3b281321d4a59926 before we ship - attempting to override the scope causes all sorts of oddities (I'm seeing the tests blow up with
```bash
RuntimeError: /Users/adriannachang/src/github.com/Shopify/maintenance_tasks/app/models/maintenance_tasks/run.rb:8: warning: already initialized constant MaintenanceTasks::Run::STATUSES
```
)
And also just seems to be discouraged by Rails:
https://github.com/rails/rails/blob/7b9cfde741662603f7fdd78698190b2fdefead17/activerecord/lib/active_record/scoping/named.rb#L197-L203

We can just create our own scope instead, which is what I've done, but there is a slight concern that the original scope would get used by mistake. WDYT?

app/models/maintenance_tasks/run.rb
```diff
@@ -41,6 +41,13 @@ class Run < ApplicationRecord
 
     scope :active, -> { where(status: ACTIVE_STATUSES) }
 
+    # Ensure ActiveStorage is in use before preloading the attachments
+    scope :with_attached_csv, -> do
+      if ActiveStorage::Attachment.table_exists?
+        includes(csv_file_attachment: :blob)
```
etiennebarrie, line 47: Why do we have to re-implement it ourselves? Can we rely on `with_attached_csv_file` instead?

etiennebarrie (approved):
> override the scope causes all sorts of oddities

I believe it's just a side-effect of loading failing because of the warning, then the file being loaded again.

> there is a slight concern that the original scope would get used by mistake

I don't think that's an issue, we control that code entirely. But if we really don't want to have two methods, we can:
```ruby
    # Ensure ActiveStorage is in use before preloading the attachments
    module OptionalWithAttachedCsvFile
      def with_attached_csv_file
        if ActiveStorage::Attachment.table_exists?
          super
        else
          all
        end
      end
    end
    singleton_class.prepend OptionalWithAttachedCsvFile
```

etiennebarrie:
No tests, not sure how to add some that won't be just stubbing `table_exists?`, but to tophat, I did:

```sh-session
$ rails db:migrate:down VERSION=20210112141723
$ rm test/dummy/db/migrate/20210112141723_create_active_storage_tables.active_storage.rb
```

PR #361, opened 2021-03-02
Always return relation from `with_attached_csv`

sambostock:
Follow up to #359.

If the attachments table exists, we return a relation, but if not, we return `nil`.

We should return an empty relation instead, so we don't have to worry about follow up calls to methods like `.order` failing.

### DIFF
--- a/app/models/maintenance_tasks/run.rb
+++ b/app/models/maintenance_tasks/run.rb
@@ -43,7 +43,11 @@ module MaintenanceTasks
 
     # Ensure ActiveStorage is in use before preloading the attachments
     scope :with_attached_csv, -> do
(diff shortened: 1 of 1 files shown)


etiennebarrie:
`scope` handles `nil` for us:
```sh-session
$ rails db:migrate:down VERSION=20210112141723
$ rm test/dummy/db/migrate/20210112141723_create_active_storage_tables.active_storage.rb
$ rails c
>> MaintenanceTasks::Run.with_attached_csv.includes_values
=> []
>> MaintenanceTasks::Run.with_attached_csv.count
   (110.1ms)  SELECT COUNT(*) FROM "maintenance_tasks_runs"
=> 100126
```

sambostock:
Today I learned 🚀

Issue #362, opened 2021-03-03
Remove dependency on Pagy


adrianna-chang-shopify:
[Pagy](https://github.com/ddnexus/pagy) is a nice, simple pagination gem that we turned to for paginating previous runs. However, it uses offset, which is prohibited in Shopify Core:
![Screen Shot 2021-03-03 at 2 48 46 PM](https://user-images.githubusercontent.com/22918438/109877595-35aa3e00-7c41-11eb-9d6b-603d9175d0ee.png)

and just generally won't scale well. Plus, most apps don't really need paginated runs since they'll just run a Task once. It might be worthwhile to investigate removing Pagy and implementing some sort of value-based pagination, or even dropping pagination and allowing users to load more data on their page with a button click at the bottom of their page.

simonlevasseur:
I saw your linked PR and it seems you have a pretty good handle on things but wanted to link a similar issue we had in Partners. We analyzed the performance implications of value-based pagination vs offset-based pagination. Not only is the performance better but it's also more reliable as documents change.

See https://github.com/Shopify/partners/pull/23097

adrianna-chang-shopify:
That's an awesome writeup, thanks for sharing @simonlevasseur ! ❤️  The more I read up on the two, the more apparent it became that we should make the switch, but it's really neat to see the performance implications in an actual production application.

PR #363, opened 2021-03-05
Switch to cursor-based pagination with geared_pagination

adrianna-chang-shopify:
Closes: https://github.com/Shopify/maintenance_tasks/issues/362

As mentioned in the issue, Pagy uses offset-based pagination, which decreases in performance as the collection of elements being paginated grows. It doesn't scale very well, and queries that use `OFFSET` are outright banned in Core for performance reasons. A cursor based pagination strategy queries based on a specific point in the dataset (the cursor), and works well for 
sequential datasets / datasets with ordered attributes (`maintenance_tasks_runs` table is guaranteed to be sortable by `created_at` ).

Basecamp has actually written a gem that supports [cursor-based pagination](https://github.com/basecamp/geared_pagination#cursor-based-pagination), which I think is a really good candidate for us to use, replacing Pagy.

### Implementation
- As the gem's README notes, we use `set_page_and_extract_portion_from` with the `ordered_by`  option to perform cursor-based pagination
- Using this form of pagination requires the original relation to be unordered, so I've dropped the ordering on `#previous_runs`, since the pagination will handle this for us now
- This provides us with an instance var `@page`, which contains the recordset of paginated elements which we can render in the view (similar to what was happening with Pagy)
- **What is config/initializers/geared_pagination.rb all about?**
     - The gem works by [including `GearedPagination::Controller` in `ActionController::Base`](https://github.com/basecamp/geared_pagination/blob/140de067b283dfc6d564fa91b8860b2badc57286/lib/geared_pagination/engine.rb#L7) when the gem is initialized
      -  `GearedPagination::Controller` sets an `after_action` filter, `set_paginated_headers`
      - This currently breaks our test suite, because `set_paginated_headers` checks an ivar that doesn't exist (`@page`) which warns and thus raises in our test suite. I also don't really like that this filter runs before every single controller action in the engine _and_ the host app. 
      - So, I've added an initializer that removes this filter and puts the onus on us to call `set_paginated_headers` when we need to (for now, this is only in `TasksController#show`)
      - Arguably this is a little brittle. We could write a wrapper to isolate the call to `set_paginated_headers` in case we want to reuse pagination elsewhere, but I thought maybe that was premature

### To 🎩 
- Create a bunch of `runs` in the dummy (ie. `rails db:seed` a couple times)
- Check out http://localhost:3000/maintenance_tasks/tasks/Maintenance::TestTask and verify no pagination
- Check out http://localhost:3000/maintenance_tasks/tasks/Maintenance::UpdatePostsTask and test the pagination

Screen Shot 2021-03-05 at 10 33 13 AM

### Tradeoffs
- We lose the ability to show individual "pages" and to go back to a previous page. However, I think it will be pretty rare that a ton of pages will be needed
- We also gain the ability to provide a URL to a given page based on a cursor now which is kind of nice

### DIFF
--- a/config/initializers/geared_pagination.rb
+++ b/config/initializers/geared_pagination.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+GearedPagination::Engine.config.after_initialize do
+  ActiveSupport.on_load(:action_controller) do
+    ActionController::Base.skip_after_action(:set_paginated_headers)
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -14,3 +14,5 @@ gem 'selenium-webdriver'
 gem 'sqlite3'
 gem 'webdrivers', require: false
 gem 'yard'
+
+gem 'geared_pagination', '~> 1.1'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -6,7 +6,6 @@ PATH
       activejob (>= 6.0)
       activerecord (>= 6.0)
       job-iteration (~> 1.1)
-      pagy (~> 3.9)
       railties (>= 6.0)
 
 GEM
@@ -115,7 +114,6 @@ GEM
     nokogiri (1.11.1)
       mini_portile2 (~> 2.5.0)
       racc (~> 1.4)
-    pagy (3.11.0)
     parallel (1.20.1)
     parser (2.7.2.0)
       ast (~> 2.4.1)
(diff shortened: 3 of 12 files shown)


config/initializers/geared_pagination.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+GearedPagination::Engine.config.after_initialize do
+  ActiveSupport.on_load(:action_controller) do
+    ActionController::Base.skip_after_action(:set_paginated_headers)
```
etiennebarrie, line 4: Given this kind of implementation detail, I wonder if we're not better off implementing it ourselves.

adrianna-chang-shopify:
Opted for https://github.com/Shopify/maintenance_tasks/pull/371 instead

Issue #350, opened 2021-02-26
Dummy fixtures aren't loaded - tests not running as expected


adrianna-chang-shopify:
@sambostock kindly noted that `Post.count` returns `0` in `test/jobs/maintenance_tasks/task_job_test.rb` 😅  
Turns out our dummy fixtures are not loaded by the test helper, because only a single fixture path is permitted and naturally it points to our engine fixtures. Consequently, any of our tests that were relying on `Post`s having been created are not doing much at all (I think this is just our `TaskJobTest` + system tests).

Options:
1) We manually create some posts in any of the tests that need them
2) We move all of our fixtures to be in the dummy, and then change our fixture load path in `test_helper.rb` to use the dummy instead

I think we might be okay just to go with 2 - the only fixture we have for the engine right now is `runs.yml`, which are sample runs that should be perceived as being within the context of the dummy anyways.

PR #366, opened 2021-03-08
Move dummy fixtures into the engine's fixtures directory

adrianna-chang-shopify:
Closes: https://github.com/Shopify/maintenance_tasks/issues/350

Let's move the dummy fixtures to the engine so they get loaded in the tests.


sambostock (approved):

etiennebarrie (commented):
Shouldn't we get rid of the fixtures in the dummy app if they're misleading?
Regarding system tests, they seem to be using the seeds anyway:

```diff
diff --git i/test/system/maintenance_tasks/tasks_test.rb w/test/system/maintenance_tasks/tasks_test.rb
index cc3426c..7a52af3 100644
--- i/test/system/maintenance_tasks/tasks_test.rb
+++ w/test/system/maintenance_tasks/tasks_test.rb
@@ -50,6 +50,7 @@ class TasksTest < ApplicationSystemTestCase
 
     test 'visit main page through iframe' do
       visit root_path
+      take_screenshot
 
       within_frame('maintenance-tasks-iframe') do
         assert_content 'Maintenance Tasks'
```

I get posts in the system tests:
image

The seeds seem to be coming from the db:setup call, and they seem necessary for at least one system test: 

```
rails test test/system/maintenance_tasks/runs_test.rb -n test_resume_a_Run
```

which fails if you haven't run `rake db:setup` before, for example:

```sh-session
$ rails db:test:prepare
$ rails test test/system/maintenance_tasks/runs_test.rb -n test_resume_a_Run
```


Whereas moving the fixtures in the right spot seems to be doing the right thing, fixtures get loaded and there are 2 posts even in tests (the seeds run even for test, but then they get overwritten by the fixtures).

etiennebarrie:
> I don't really like moving the posts fixture to the engine though, since it's dummy code.

The dummy app is meant for tests really, it sits in `test/`, so it's not entirely misplaced to have fixtures in `test/fixtures`.

We also have things like the list of tasks from the dummy app tested in test/, which is a very similar situation IMO:
https://github.com/Shopify/maintenance_tasks/blob/5581f73270e88fa02209279a6f11926affbcd24c/test/tasks/maintenance_tasks/task_test.rb#L8-L13

I don't feel so good about using seeds in tests, there was a discussion in https://discuss.rubyonrails.org/t/should-dbprepare-also-call-db-seed-by-default/74835 around this, it's not clear cut, but I think we ought to use `db:test:prepare`. One thing where it's worse than `db:setup` is that it doesn't really check for pending migrations, and the error happens when the tests run instead of when `db:setup` runs:

https://github.com/rails/rails/blob/90d0b42bd8206e942597b64163837287caf7119d/activerecord/lib/active_record/railties/databases.rake#L362-L363

That `abort_if_pending_migrations` task is not documented so I don't think we should depend on it either: https://github.com/rails/rails/blob/90d0b42bd8206e942597b64163837287caf7119d/activerecord/lib/active_record/railties/databases.rake#L314-L315

But overall it's fine, db:test:prepare loads the schema, and rails test or rails test:system fail because a migration is pending.

adrianna-chang-shopify:
> We also have things like the list of tasks from the dummy app tested in test/

Yeah, I guess the thing I'm stuck on is that at least the dummy-related data we have in `test/` (list of Tasks) and the engine's fixtures (Runs) are data models related to the engine, whereas Posts are entirely the business logic of the dummy. 

But you're right, the argument on using seeds for the test db is murky and not a great option either (thanks for linking the thread, was an insightful read ❤️ ). I'm on board with using `db:test:prepare` instead of `db:setup`, but I'd still prefer to keep the fixtures where they are. What do you think about explicitly trying to load the `posts.yml` fixtures in either `test_helper.rb`, or even just in `application_system_test_helper.rb` for the system tests like this:

```ruby
    fixtures_dir = 'test/dummy/test/fixtures'
    names = Dir["#{fixtures_dir}/**/*.yml"].map do |path|
      File.basename(path, '.yml')
    end
    ActiveRecord::FixtureSet.create_fixtures(fixtures_dir, names)
```

etiennebarrie:
I guess we could do this: (no need to support other fixture sets since we're probably not going to add any)
```ruby
ActiveRecord::FixtureSet.create_fixtures(Rails.root.join('test', 'dummy', 'test', 'fixtures'), :posts)
```
But I still don't see what the problem is with moving the file 😆

etiennebarrie (approved):

PR #370, opened 2021-03-09
[WIP] Value-based pagination

adrianna-chang-shopify:
Here's a proof of concept for doing value-based pagination ourselves. Not sure how general we want it to be. Right now it assumes that:
- `id` is used for the cursor
- the relation is ordered by `id: :desc` (our `runs` are actually ordered by `created_at: :desc`, we'll have to think about what to do here... Since id is auto-incrementing these mean essentially the same thing, but using `created_at` as our cursor is tricky since there is no uniqueness clause on this column right now)
- the relation is already ordered when it is passed to the `Page` initializer

We could make this customizable by allowing the column / columns for the cursor to be specified, and the direction of the relation to be specified, but not sure if this will be reused beyond showing runs. I feel like at that point we also have to start validating whether the column has a unique order, ensuring the relation is not already ordered before it is passed in OR checking that the order matches what is specified, etc... Not sure it's worth the effort.

Let me know what you think.

### DIFF
--- a/app/controllers/maintenance_tasks/tasks_controller.rb
+++ b/app/controllers/maintenance_tasks/tasks_controller.rb
@@ -20,7 +20,11 @@ module MaintenanceTasks
     def show
       @task = TaskData.find(params.fetch(:id))
       set_refresh if @task.last_run&.active?
-      @pagy, @previous_runs = pagy(@task.previous_runs)
+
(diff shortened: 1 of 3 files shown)


adrianna-chang-shopify:
Closing in favour of https://github.com/Shopify/maintenance_tasks/pull/371

Issue #369, opened 2021-03-08
Allow ActiveStorage service to be configured


adrianna-chang-shopify:
In Rails 6.1, ActiveStorage allows separate cloud storage services to be configured in an application. Attachments can specify which service should be used (see [the docs](https://guides.rubyonrails.org/v6.1/active_storage_overview.html#attaching-files-to-records)).

We should expose a config option on the gem that allows users to specify which storage service they would like the gem's active storage blobs to be uploaded to.

cc @simonlevasseur

PR #372, opened 2021-03-10
Allow configuration of Active Storage service

adrianna-chang-shopify:
Closes: https://github.com/Shopify/maintenance_tasks/issues/369

Rails 6.1+ allows multiple services to be configured per environment for Active Storage. When [defining an attachment on a model](https://edgeapi.rubyonrails.org/classes/ActiveStorage/Attached/Model.html#method-i-has_one_attached), the `service` option can be used to specify which service config to use.

Let's allow users to specify which service to use for `csv_file` attachments on our `Run` model by making it a config option on the gem.

🎩 

```
# test/dummy/config/initializers/maintenance_tasks.rb
MaintenanceTasks.active_storage_service = :local
```

Try running the `Maintenance::ImportPostsTask` and ensure that the blob gets uploaded to `test/dummy/storage` rather than `test/dummy/tmp/storage`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ A Rails engine for queuing and managing maintenance tasks.
     * [Customizing the maintenance tasks module](#customizing-the-maintenance-tasks-module)
     * [Customizing the underlying job class](#customizing-the-underlying-job-class)
     * [Customizing the rate at which task progress gets updated](#customizing-the-rate-at-which-task-progress-gets-updated)
+    * [Customizing which Active Storage service to use](#customizing-which-active-storage-service-to-use)
 * [Upgrading](#upgrading)
 * [Contributing](#contributing)
 * [Releasing new versions](#releasing-new-versions)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -400,6 +401,36 @@ MaintenanceTasks.ticker_delay = 2.seconds
...
+
+```yaml
+# config/storage.yml
+gcs_primary:
+  service: GCS
+  credentials: <%= Rails.root.join("path/to/primary/keyfile.json") %>
+  project: "my-project"
+  bucket: "primary-bucket"
+
+gcs_secondary:
```
etiennebarrie, line 419: Maybe we could use more realistic names like `user_data` and `internal` or something like that?

etiennebarrie (approved):
Great job! 👏 

I did the opposite for the 🎩 , by default development was already using `:local`, so I changed it to `:test` to have files show up in `test/dummy/tmp/storage/`.

PR #376, opened 2021-03-19
Use double quoted strings

adrianna-chang-shopify:
Shopify's Ruby Style Guide was [recently changed](https://github.com/Shopify/ruby-style-guide/pull/238) to prefer double quotes over single quotes for string literals, so I've changed the rule for Rubocop and changed our strings to be compliant 😁

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -24,7 +24,7 @@ Style/DocumentationMethod:
     - lib/**/*.rb
 
 Style/StringLiterals:
-  EnforcedStyle: single_quotes
+  EnforcedStyle: double_quotes
 
 Style/SymbolArray:
   EnforcedStyle: brackets
--- a/Gemfile
+++ b/Gemfile
@@ -1,16 +1,16 @@
 # frozen_string_literal: true
 
-source 'https://rubygems.org'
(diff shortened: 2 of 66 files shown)


.rubocop.yml
```diff
@@ -24,7 +24,7 @@ Style/DocumentationMethod:
     - lib/**/*.rb
 
 Style/StringLiterals:
-  EnforcedStyle: single_quotes
+  EnforcedStyle: double_quotes
```
etiennebarrie, line 27: Can we wait until the gem is updated to entirely remove that section instead? Or we can ship and fix that later with the update.

etiennebarrie (approved):

.rubocop.yml
adrianna-chang-shopify: I'd prefer to ship now so we don't have to continue making changes / maintaining this PR as new strings get added. Since the current version doesn't specify a rule for `Style/StringLiterals`, we could just remove this now ahead of the new version being released, but it's nice to have it for autoformatting / external contributors IMHO. I'll go ahead and ship this, but leave a note that we should remove this rule when `rubocop-shopify` gets bumped.

PR #377, opened 2021-03-19
Make progress information visible in the info card

adrianna-chang-shopify:
Having the elapsed ticks / total ticks hidden away in the title attribute of the progress bar is not very discoverable for users. For long-running tasks where progress is slow and not very obvious by the progress bar, it helps to see the tick count to have confidence that the Task is still progressing.

I've moved the progress info out of the `title` attribute of the `progress` element, and into a span underneath the progress bar. Changes are as follows:
- Renamed `Progress#title` to `Progress#text` since it's no longer really a title
- Cleaned up `Progress#text` a bit, and made the returned string more consistent (notably `Processed 10 out of 10 (100%)` => `Processed 10 out of 10 items (100%).`
- Changed `TasksHelper#progress` to return a div wrapping both the progress element and a span containing the progress text. Open to styling options on this if we're not happy with it, I went with having the text in italics for now.

Samples from Tophat:

Screen Shot 2021-03-19 at 8 56 21 AM
Screen Shot 2021-03-19 at 8 56 35 AM
Screen Shot 2021-03-19 at 8 56 49 AM
Screen Shot 2021-03-19 at 8 57 01 AM
Screen Shot 2021-03-19 at 8 58 52 AM

### DIFF
--- a/app/helpers/maintenance_tasks/tasks_helper.rb
+++ b/app/helpers/maintenance_tasks/tasks_helper.rb
@@ -49,8 +49,8 @@ module MaintenanceTasks
         max: progress.max,
         class: ["progress"] + STATUS_COLOURS.fetch(run.status)
       )
-      progress_text = tag.i(progress.text)
-      tag.div(progress_bar.concat(progress_text), class: "block")
+      progress_text = tag.p(tag.i(progress.text))
+      tag.div(progress_bar + progress_text, class: "block")
     end
 
     # Renders a span with a Run's status, with the corresponding tag class
--- a/app/helpers/maintenance_tasks/tasks_helper.rb
+++ b/app/helpers/maintenance_tasks/tasks_helper.rb
@@ -44,12 +44,13 @@ module MaintenanceTasks
 
       progress = Progress.new(run)
 
-      tag.progress(
+      progress_bar = tag.progress(
         value: progress.value,
         max: progress.max,
-        title: progress.title,
         class: ["progress"] + STATUS_COLOURS.fetch(run.status)
       )
+      progress_text = tag.i(progress.text)
+      tag.div(progress_bar.concat(progress_text), class: "block")
     end
 
     # Renders a span with a Run's status, with the corresponding tag class
--- a/app/models/maintenance_tasks/progress.rb
+++ b/app/models/maintenance_tasks/progress.rb
@@ -4,6 +4,7 @@ module MaintenanceTasks
   # This class generates progress information about a Run.
   class Progress
     include ActiveSupport::NumberHelper
+    include ActionView::Helpers::TextHelper
 
     # Sets the Progress initial state with a Run.
     #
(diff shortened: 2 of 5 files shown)


app/helpers/maintenance_tasks/tasks_helper.rb
```diff
@@ -44,12 +44,13 @@ def progress(run)
 
       progress = Progress.new(run)
 
-      tag.progress(
+      progress_bar = tag.progress(
         value: progress.value,
         max: progress.max,
-        title: progress.title,
```
sambostock, line 50: Is there accessibility value in continuing to have a title of some sort on the progress bar?

app/helpers/maintenance_tasks/tasks_helper.rb
```diff
@@ -44,12 +44,13 @@ def progress(run)
...
         class: ['progress'] + STATUS_COLOURS.fetch(run.status)
       )
+      progress_text = tag.i(progress.text)
+      tag.div(progress_bar.concat(progress_text), class: 'block')
```
etiennebarrie, line 53: Maybe just `+` would be simpler than `concat`?

app/helpers/maintenance_tasks/tasks_helper.rb
```diff
@@ -44,12 +44,13 @@ def progress(run)
...
       progress = Progress.new(run)
 
-      tag.progress(
+      progress_bar = tag.progress(
         value: progress.value,
         max: progress.max,
-        title: progress.title,
         class: ['progress'] + STATUS_COLOURS.fetch(run.status)
       )
+      progress_text = tag.i(progress.text)
```
etiennebarrie, line 52: I think we should wrap all text into `<p>`.

app/models/maintenance_tasks/progress.rb
```diff
@@ -73,5 +74,9 @@ def estimatable?
     def over_total?
       @run.tick_count > @run.tick_total
     end
+
+    def ticks_in_words(amount, unit = 'item')
```
etiennebarrie, line 78: I don't think we need the `unit` parameter here.

app/models/maintenance_tasks/progress.rb
```diff
@@ -73,5 +74,9 @@ def estimatable?
...
+      "#{amount} #{unit.pluralize(amount)}"
```
etiennebarrie, line 79: There's a `pluralize` helper that does that: https://api.rubyonrails.org/classes/ActionView/Helpers/TextHelper.html#method-i-pluralize

app/helpers/maintenance_tasks/tasks_helper.rb
adrianna-chang-shopify: Moving the text into a `<span>` and out of the title is probably a _plus_ on the accessibility side of things - from https://accessible-app.com/pattern/vue/progress:

> Using progress element alone is not the worst thing you could do, but there are always way to improve. If you think, for example, that the presence of a <progress> bar alone leads to automatic screen reader output, when it updates – nope, that is not the case. A screen reader's virtual cursor has to be placed on the progressbar in order to hear its information. 

Whereas I believe pulling the text out into a span makes it a readable element. 

We could keep the title with the same value, but I'm not sure it's worth having that information duplicated 🤷‍♀️

app/models/maintenance_tasks/progress.rb
adrianna-chang-shopify: Nice, forgot this was a thing! 🙌

adrianna-chang-shopify: Oh, i remember why we didn't use this before - it's from `action_view`. Are we okay with including the `TextHelper` module?

etiennebarrie (approved):
👍

PR #379, opened 2021-03-26
Bump gem to v1.2

adrianna-chang-shopify:
Draft release here: https://github.com/Shopify/maintenance_tasks/releases/edit/untagged-799da69791c7b7acc355

I decided to bump us up to v1.2.0 rather than v1.1.3, since we had a couple minor changes that were not strictly bug fixes.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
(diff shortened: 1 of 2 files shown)


etiennebarrie (approved):
:shipit:

PR #385, opened 2021-03-29
Maintain backwards compatibility when using #has_one_attached

adrianna-chang-shopify:
`v1.2.0` is broken for all Rails versions < 6.1.0 due to `has_one_attached: unknown keyword: service (ArgumentError)`, introduced [in this PR](https://github.com/Shopify/maintenance_tasks/pull/372).

~~Couldn't think of a better way to patch this and preserve functionality for Rails < 6.1.0 while allowing service configuration for Rails 6.1+ apps than checking the kwargs on the method and either passing in a service or not 😅  Let me know if you can think of a better way!~~

Opting instead to check for the presence of the `MaintenanceTasks.active_storage_service` config instead to check if we should specify the `service` kwarg, rather than doing parameter introspection.

I think we may also want to expand our test suite to test against a dummy app that use Rails 6.0, and then Rails 6.1 to make sure we catch things like this moving forward.

# 🎩 
- Change the dummy to use Rails 6.0
- Without these changes, `has_one_attached: unknown keyword: service (ArgumentError)` should be raised
- With these changes, ActiveStorage should be fine
- Moving back to Rails 6.1, these changes should be fine

### DIFF
--- a/app/models/maintenance_tasks/run.rb
+++ b/app/models/maintenance_tasks/run.rb
@@ -48,7 +48,12 @@ module MaintenanceTasks
 
     validates_with RunStatusValidator, on: :update
 
-    has_one_attached :csv_file, service: MaintenanceTasks.active_storage_service
+    if MaintenanceTasks.active_storage_service.present?
+      has_one_attached :csv_file,
+        service: MaintenanceTasks.active_storage_service
+    else
+      has_one_attached :csv_file
+    end
 
     # Sets the run status to enqueued, making sure the transition is validated
     # in case it's already enqueued.


app/models/maintenance_tasks/run.rb
```diff
@@ -48,7 +48,12 @@ class Run < ApplicationRecord
 
     validates_with RunStatusValidator, on: :update
 
-    has_one_attached :csv_file, service: MaintenanceTasks.active_storage_service
```
etiennebarrie, line 51: Maybe we could use check `MaintenanceTasks.active_storage_service` instead? If it's not set, we don't pass it which works the same? I feel like it's less hack-ish than doing introspection on the method?

That doesn't solve the problem if a user tries to set the configuration and is not using Rails 6.1 but then that happens at development time, not when upgrading the gem.

adrianna-chang-shopify: This seems sensible to me 👍 I think if the user tries to set it and it fails because they're not yet on 6.1 it's fine.

etiennebarrie (approved):

PR #391, opened 2021-04-12
Make Active Storage optional

etiennebarrie:
This allows loading the gem and running regular Tasks when `active_storage/engine` is not required in `config/application.rb`.

Not having required `active_storage/engine` will have the same result as having it required but not configured. An error is raised when a CSV file is uploaded, but no earlier.

This is handled in the README so I'm not sure we want to do more about this.

### DIFF
--- a/app/models/maintenance_tasks/run.rb
+++ b/app/models/maintenance_tasks/run.rb
@@ -43,6 +43,7 @@ module MaintenanceTasks
 
     # Ensure ActiveStorage is in use before preloading the attachments
     scope :with_attached_csv, -> do
+      return unless defined?(ActiveStorage)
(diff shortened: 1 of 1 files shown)


adrianna-chang-shopify (approved):
Nice! ✨

sambostock (commented):
Would it make sense to raise if `MaintenanceTasks::Task.csv_collection` is called and active storage isn't setup? It shouldn't be expensive, as it would only happen when defining task classes.

That way, you catch it at the point when someone creates a CSV task in the first place, instead of when they try to run it.

etiennebarrie:
> you catch it at the point when someone creates a CSV task in the first place

Right. I think that makes sense, but for this PR I just kept the existing behavior. I'll follow up!

PR #393, opened 2021-04-16
Release 1.2.2

etiennebarrie:
I noticed we weren't shipping the license file which I think we are supposed to. The gem wasn't specified as MIT-licensed either.

And a bump for making Active Storage optional.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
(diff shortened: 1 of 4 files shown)


adrianna-chang-shopify (approved):
Good catch on the license stuff ❤️

etiennebarrie:
image
⬆️ MIT

PR #394, opened 2021-04-19
Tweak workflow labelling and triggers

sambostock:
This is almost identical to the change I made in https://github.com/Shopify/job-iteration/pull/74.

- Only run on push. Running on Pull Request seems redundant, since all PRs would include a push.
- Rename the workflow from "Ruby" to "CI" to avoid redundant prefix
- Append Gemfile from matrix to job name, so they are distinguishable

#### :warning: Before Merging

- [x] Squash commits and update commit messages

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: Ruby
+name: CI
 
 on: [push, pull_request]
 
@@ -6,7 +6,7 @@ jobs:
   build:
 
     runs-on: ubuntu-latest
-    name: Ruby ${{ matrix.ruby }}
+    name: Ruby ${{ matrix.ruby }} / ${{ matrix.gemfile }}
     strategy:
       matrix:
         gemfile: [Gemfile, gemfiles/rails_6_0.gemfile, gemfiles/rails_edge.gemfile]


.github/workflows/ci.yml
```diff
@@ -1,12 +1,12 @@
-name: Ruby
+name: CI
```
etiennebarrie, line 1: Is there some kind of de-prefixing going on? We don't have `Ruby / Ruby` at the moment, but this adds the `CI / ` at the beginning 🤔.

I don't really like the `CI / ` prefix.

.github/workflows/ci.yml
```diff
@@ -1,12 +1,12 @@
...
 
-on: [push, pull_request]
+on: [push]
```
etiennebarrie, line 3: Yeah I think we want to keep that, as @adrianna-chang-shopify mentioned it's not running the same code.

.github/workflows/ci.yml
```diff
@@ -1,12 +1,12 @@
...
 
 jobs:
   build:
 
     runs-on: ubuntu-latest
-    name: Ruby ${{ matrix.ruby }}
+    name: Ruby ${{ matrix.ruby }} | Gemfile ${{ matrix.gemfile }}
```
etiennebarrie, line 9: This adds a whole lot of duplication with `Gemfile [gG]emfile` at the begining of each label. Maybe: 
```suggestion
    name: Ruby ${{ matrix.ruby }}: ${{ matrix.gemfile }}
```

.github/workflows/ci.yml
adrianna-chang-shopify: I think it may be something that changed recently about GH actions. #395 that was just opened seems to prefix with `Ruby / Ruby`:
Screen Shot 2021-04-20 at 1 20 36 PM

adrianna-chang-shopify: If there's no way to omit the name from the Github check (which it doesn't look like there is), IMHO `CI` is an improvement over having `Ruby` twice 😁

.github/workflows/ci.yml
sambostock: Yeah, interesting. I didn't realize it used a separate branch. I wish the GitHub Actions showed examples. Maybe it's just me, but I feel like it could have been clearer.

Anyway, I'll undo the change. I wonder if we should ensure other projects also do the same (namely Job Iteration).

.github/workflows/ci.yml
sambostock: Yeah, that works. I just did the same thing as `Job Iteration`, but it [uses a matrix over the filename basenames](https://github.com/Shopify/job-iteration/blob/6d6af27b0bae5bd6742bff2f5c1473be8be86559/.github/workflows/ci.yml#L17):

```yml
gemfile: [rails_5_2, rails_6_0, rails_edge]
```

Given our setup here, dropping the prefix makes sense.

adrianna-chang-shopify (approved):

.github/workflows/ci.yml
```diff
@@ -1,12 +1,12 @@
...
+name: CI
 
 on: [push, pull_request]
 
 jobs:
   build:
 
     runs-on: ubuntu-latest
-    name: Ruby ${{ matrix.ruby }}
+    name: Ruby ${{ matrix.ruby }} | ${{ matrix.gemfile }}
```
etiennebarrie, line 9: Maybe `/` like GitHub does?

etiennebarrie (approved):
Tiny comment on style but feel free to squash and ship.

sambostock:
Right, @etiennebarrie, @adrianna-chang-shopify do you know how we can change the list of required statuses?

etiennebarrie:
Yes we can do it.

PR #395, opened 2021-04-20
Enhance usability of task_spec generator

RouzbehMajidi:
Just a quick fix, we use the task_spec generator a lot and it is handy that it generates a few lines to get you started, so I thought we might as well be able to use those lines in our code rather than needing to delete them.

#### Problem
The current generator generates something like
```ruby
# frozen_string_literal: true
require 'rails_helper'

module Maintenance
  RSpec.describe SomeMaintenanceTask do
    # describe '#process' do
      # it 'performs a task iteration' do
        # Maintenance::SomeMaintenanceTask.process(element)
      # end
    # end
  end
end
```
If we're generating this starter code, I think we might as well generate it in a somewhat usable manner, rather than generating it for it to be deleted later 

#### Changes
This is a minor change but I just rearranged the .process and cleaned it up a bit so it matches up with the [RSpec/DescribedClass convention](https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/DescribedClass). The generator will now generate something like this
```ruby
# frozen_string_literal: true
require 'rails_helper'

module Maintenance
  RSpec.describe SomeMaintenanceTask do
    # describe '#process' do
      # subject(:process) { described_class.process(element) }
      # it 'performs a task iteration' do
        # expect something here
      # end
    # end
  end
end
```
This makes the generated code somewhat more usable and I believe it is more helpful. Let me know what you guys think! maybe this is all unnecessary..

### DIFF
--- a/lib/generators/maintenance_tasks/templates/task_spec.rb.tt
+++ b/lib/generators/maintenance_tasks/templates/task_spec.rb.tt
@@ -4,11 +4,13 @@ require 'rails_helper'
 module <%= tasks_module %>
 <% module_namespacing do -%>
   RSpec.describe <%= class_name %>Task do
-    # describe '#process' do
-      # it 'performs a task iteration' do
-        # <%= tasks_module %>::<%= class_name %>Task.process(element)
-      # end
-    # end
+    describe "#process" do
+      subject(:process) { described_class.process(element) }
+      let(:element) { 
+        # Object to be processed in a single iteration of this task
+      }
+      pending "add some examples to (or delete) #{__FILE__}"
+    end
   end
 <% end -%>
 end


lib/generators/maintenance_tasks/templates/task_spec.rb.tt
```diff
@@ -5,8 +5,9 @@ module <%= tasks_module %>
 <% module_namespacing do -%>
   RSpec.describe <%= class_name %>Task do
     # describe '#process' do
+      # subject(:process) { described_class.process(element) }
       # it 'performs a task iteration' do
-        # <%= tasks_module %>::<%= class_name %>Task.process(element)
+        # expect something here
```
etiennebarrie, line 10: Should we make this also more useful? This line is not going to be directly usable either.

lib/generators/maintenance_tasks/templates/task_spec.rb.tt
```diff
@@ -5,8 +5,9 @@ module <%= tasks_module %>
 <% module_namespacing do -%>
   RSpec.describe <%= class_name %>Task do
     # describe '#process' do
+      # subject(:process) { described_class.process(element) }
```
etiennebarrie, line 8: `element` is not defined, so this is not going to be ideal either, is it?

etiennebarrie (commented):
We could take example from rspec-rails maybe?
https://github.com/rspec/rspec-rails/blob/main/lib/generators/rspec/job/templates/job_spec.rb.erb
https://github.com/rspec/rspec-rails/blob/main/lib/generators/rspec/model/templates/model_spec.rb

It uses `pending` to get something immediately useful.

RouzbehMajidi:
Thanks for the great feedback @etiennebarrie. I've updated the following
- used `pending` instead of an empty example
- added an empty `element` item with a comment that reads `Object to be processed in a single iteration of this task` similar to the one here https://github.com/Shopify/maintenance_tasks/blob/0b0d44221c73ecb95438282843584e00cb612556/lib/generators/maintenance_tasks/templates/task.rb.tt#L12

Below is the new output
```ruby
# frozen_string_literal: true
require 'rails_helper'

module Maintenance
  RSpec.describe SleepyTask do
    # describe "#process" do
    #   subject(:process) { described_class.process(element) }
    #   let(:element) { 
    #     # Object to be processed in a single iteration of this task
    #   }
    #   pending "add some examples to (or delete) #{__FILE__}"
    # end
  end
end
```
once this code is uncommented it will produce the following
```ruby
# frozen_string_literal: true
require 'rails_helper'

module Maintenance
  RSpec.describe SleepyTask do
    describe "#process" do
      subject(:process) { described_class.process(element) }
      let(:element) { 
        # Object to be processed in a single iteration of this task
      }
      pending "add some examples to (or delete) #{__FILE__}"
    end
  end
end
```
which will raise the appropriate pending warning. Let me know what you guys think!

etiennebarrie (approved):
Maybe we could not comment it now that it works out of the box?

RouzbehMajidi:
> Maybe we could not comment it now that it works out of the box?

Good point, updated!

adrianna-chang-shopify:
Nice work @RouzbehMajidi ! Feel free to ship :shipit:

PR #400, opened 2021-05-04
Introduce Mechanism for Throttling

adrianna-chang-shopify:
Closes: https://github.com/Shopify/maintenance_tasks/issues/298

I revisited adding throttling! I think adding a simple DSL like this optimizes for ease of use from a user's perspective:

```ruby
throttle_on -> { <condition> }, backoff: <backoff>
```

The PR is pretty simple - JobIteration's [Throttle Enumerator](https://github.com/Shopify/job-iteration/blob/4e4ed66c0c325d6e28ecddf84399f1b291ae8ae4/lib/job-iteration/throttle_enumerator.rb) does most of the work for us, so we just need to wrap the enumerator we build using the collection in a throttle enumerator Job Iteration constructs via [`#build_throttle_enumerator`](https://github.com/Shopify/job-iteration/blob/4e4ed66c0c325d6e28ecddf84399f1b291ae8ae4/lib/job-iteration/enumerator_builder.rb#L113)

As discussed in https://github.com/Shopify/maintenance_tasks/pull/330, storing `throttle_specs` as an array allows Tasks to define multiple throttle conditions, and to inherit throttle specs from superclasses / mixins without clobbering those conditions if the Task defines new ones. Does this approach still make sense?

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,46 @@ title,content
 My Title,Hello World!
 ```
 
+### Throttling
+
+Maintenance Tasks often modify a lot of data and can be taxing on your database.
+The gem provides a throttling mechanism that can be used to throttle a Task when
+a given condition is met. If a Task is throttled, it will be interrupted and
+retried after a backoff period has passed. The default backoff is 30 seconds.
+Specify the throttle condition as a block:
+
+```ruby
+# app/tasks/maintenance/update_posts_throttled_task.rb
+module Maintenance
+  class UpdatePostsThrottledTask < MaintenanceTasks::Task
+    throttle_on(backoff: 1.minute) do
+      DatabaseStatus.unhealthy?
+    end
+
+    def collection
+      Post.all
+    end
+
+    def count
+      collection.count
+    end
+
+    def process(post)
+      post.update!(content: "New content added on #{Time.now.utc}")
+    end
+  end
+end
+```
+
+Note that it's up to you to define a throttling condition that makes sense for
+your app. Shopify implements `DatabaseStatus.healthy?` to check various MySQL
+metrics such as replication lag, DB threads, whether DB writes are available, 
+etc.
+
+Tasks can define multiple throttle conditions. Throttle conditions are inherited
+by descendants, and new conditions will be appended without impacting existing
+conditions.
+
 ### Considerations when writing Tasks
 
 MaintenanceTasks relies on the queue adapter configured for your application to
--- a/app/tasks/maintenance_tasks/task.rb
+++ b/app/tasks/maintenance_tasks/task.rb
@@ -1,11 +1,18 @@
 # frozen_string_literal: true
-
+puts "LOADING TASK CLASS"
 module MaintenanceTasks
   # Base class that is inherited by the host application's task classes.
   class Task
     extend ActiveSupport::DescendantsTracker
 
     class NotFoundError < NameError; end
+
+    # The throttle conditions for a given Task. This is provided as an array of
+    # hashes, with each hash specifying two keys: throttle_condition and
+    # backoff. Note that Tasks inherit conditions from their superclasses.
+    #
+    # @api private
+    class_attribute :throttle_conditions, default: []
 
     class << self
       # Finds a Task with the given name.
@@ -70,6 +77,19 @@ module MaintenanceTasks
       # @return the count of items.
       def count
         new.count
+      end
+
+      # Add a condition under which this Task will be throttled.
+      #
+      # @param backoff [ActiveSupport::Duration] optionally, a custom backoff
+      #   can be specified. This is the time to wait before retrying the Task.
+      #   If no value is specified, it defaults to 30 seconds.
+      # @yieldreturn [Boolean] where the throttle condition is being met,
+      #   indicating that the Task should throttle.
+      def throttle_on(backoff: 30.seconds, &condition)
+        self.throttle_conditions += [
+          { throttle_on: condition, backoff: backoff },
+        ]
       end
 
       private
--- a/test/dummy/app/tasks/maintenance/update_posts_throttled_task.rb
+++ b/test/dummy/app/tasks/maintenance/update_posts_throttled_task.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+module Maintenance
+  class UpdatePostsThrottledTask < MaintenanceTasks::Task
+    class << self
+      attr_accessor :throttle
+    end
+
+    throttle_on { throttle }
+
+    def collection
+      Post.all
+    end
+
+    def count
+      collection.count
+    end
+
+    def process(post)
+      post.update!(content: "New content added on #{Time.now.utc}")
+    end
+  end
+end
--- a/test/tasks/maintenance_tasks/task_test.rb
+++ b/test/tasks/maintenance_tasks/task_test.rb
@@ -11,6 +11,7 @@ module MaintenanceTasks
         "Maintenance::ImportPostsTask",
         "Maintenance::TestTask",
         "Maintenance::UpdatePostsTask",
+        "Maintenance::UpdatePostsThrottledTask",
       ]
       assert_equal expected,
         MaintenanceTasks::Task.available_tasks.map(&:name).sort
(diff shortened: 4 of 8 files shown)


README.md
```diff
@@ -112,6 +112,43 @@ title,content
...
+    end
+
+    def process(post)
+      post.update!(content: "New content added on #{Time.now.utc}")
+    end
+  end
+end
+```
+
+Optionally, a `backoff` can be supplied. The default is 30 seconds.
```
etiennebarrie, line 142: It's a bit weird that the example uses `backoff` before it's explained. I guess it makes sense to only want one example, but maybe we can leave the details to the API documentation?

README.md
```diff
@@ -112,6 +112,43 @@ title,content
...
+Note that it's up to you to define a throttling condition that makes sense for
+your app. Shopify implements `DatabaseStatus.healthy?` to check various MySQL
+metrics such as replication lag, DB threads, whether DB writes are available, 
+etc.
+
+Tasks can define multiple throttle conditions. Throttle conditions are inherited
+by descendants, meaning new conditions will be appended, rather than clobbering
+existing specs.
```
etiennebarrie, line 150: `specs` sounds like an implementation term. `conditions` maybe?

README.md
```diff
@@ -112,6 +112,43 @@ title,content
...
+```
+
+Optionally, a `backoff` can be supplied. The default is 30 seconds.
+Note that it's up to you to define a throttling condition that makes sense for
+your app. Shopify implements `DatabaseStatus.healthy?` to check various MySQL
+metrics such as replication lag, DB threads, whether DB writes are available, 
+etc.
+
+Tasks can define multiple throttle conditions. Throttle conditions are inherited
+by descendants, meaning new conditions will be appended, rather than clobbering
```
etiennebarrie, line 149: The present participle on `clobbering` is weird. Maybe we could reword to avoid it? `and won't remove existing conditions`?

app/jobs/maintenance_tasks/task_job.rb
```diff
@@ -40,6 +40,14 @@ def build_enumerator(_run, cursor:)
         raise ArgumentError, "#{@task.class.name}#collection must be either "\
           "an Active Record Relation, Array, or CSV."
       end
+
+      @task.throttle_specs.reduce(collection_enumerator) do |enumerator, spec|
```
etiennebarrie, line 44: Maybe we could extract that logic to a method?

app/tasks/maintenance_tasks/task.rb
```diff
@@ -72,6 +74,27 @@ def count
...
       end
 
+      # Returns the list of the throttle specs for a given Task. This is
+      # provided as an array of hashes, with each hash specifying two keys:
+      # throttle_condition and backoff. Note that Tasks inherit throttle specs
+      # from their superclasses.
+
+      # @return [Array<Hash{Symbol => Proc, ActiveSupport::Duration}>]
+      def throttle_specs
+        @throttle_specs ||= superclass.throttle_specs.dup
```
etiennebarrie, line 84: Could we use `class_attribute` instead?

app/tasks/maintenance_tasks/task.rb
```diff
@@ -72,6 +74,27 @@ def count
         new.count
       end
 
+      # Returns the list of the throttle specs for a given Task. This is
+      # provided as an array of hashes, with each hash specifying two keys:
+      # throttle_condition and backoff. Note that Tasks inherit throttle specs
+      # from their superclasses.
+
```
etiennebarrie, line 81:
```suggestion
```

app/tasks/maintenance_tasks/task.rb
```diff
@@ -72,6 +74,27 @@ def count
...
+      # @return [Array<Hash{Symbol => Proc, ActiveSupport::Duration}>]
+      def throttle_specs
```
etiennebarrie, line 83: So this is a public API, in case people don't want to inherit? (They could use `throttle_specs.clear` to get rid of any inherited value).
Maybe we could avoid making it public, even though it would remove some possibilities?
If we keep it public I think we may want to think more about naming? `specs` is a bit fuzzy maybe?

app/tasks/maintenance_tasks/task.rb
```diff
@@ -110,5 +133,12 @@ def process(_item)
     # @return [Integer, nil]
     def count
     end
+
+    # Returns the list of the throttle specs for a Task instance.
+    #
+    # @return [Array<Hash{Symbol => Proc, ActiveSupport::Duration}>]
+    def throttle_specs
```
etiennebarrie, line 140: Hum maybe actually we do want the instance-level to be public? That gives Task authors complete control over the throttling, and the class-level macro just makes it more convenient for the common use-cases?

README.md
```diff
@@ -112,6 +112,43 @@ title,content
...
+
+Maintenance Tasks often modify a lot of data and can be taxing on your database.
+The gem provides a throttling mechanism that can be used to throttle a Task when
+a given condition is met. Specify the throttle condition as a Proc:
+
+```ruby
+# app/tasks/maintenance/update_posts_throttled_task.rb
+module Maintenance
+  class UpdatePostsThrottledTask < MaintenanceTasks::Task
+    throttle_on -> { DatabaseStatus.unhealthy? }, backoff: 1.minute
```
etiennebarrie, line 125: Do we really need a Proc? I feel like a block would be more Rubyish?

README.md
adrianna-chang-shopify: SGTM! I had trouble wording this section in a way that made sense, thanks! 👍

app/tasks/maintenance_tasks/task.rb
adrianna-chang-shopify: Ah you're right, don't think this needs to be public. I don't think being able to customize inheritance warrants making this public API - Task authors should better configure their parent Tasks / modules if they don't want certain throttling conditions inherited.

app/tasks/maintenance_tasks/task.rb
adrianna-chang-shopify: I think I'd prefer to leave this as private API for now until we're convinced that the class-level macro is insufficient for users.

adrianna-chang-shopify: (I didn't intend this to be public API, just forgot to add the private tag 😅 )

README.md
adrianna-chang-shopify: Nope, Job Iteration says in their documentation that a Proc should be used so I went with that, but there's no reason we can't take a block for our API. I'll touch this up and you can let me know what you think.

app/tasks/maintenance_tasks/task.rb
adrianna-chang-shopify: I always forget about `class_attribute`, thanks ❤️

app/jobs/maintenance_tasks/task_job.rb
adrianna-chang-shopify: I feel like it's pretty readable as is - if I extracted it to a method I'd probably name if `reduced_throttle_conditions`, which seems to be equally legible with what's there. Feel free to push back if you feel strongly 😄

test/tasks/maintenance_tasks/task_test.rb
```diff
@@ -67,5 +68,20 @@ class TaskTest < ActiveSupport::TestCase
       message = "MaintenanceTasks::Task must implement `process`."
       assert_equal message, error.message
     end
+
+    test ".throttle_conditions inherits conditions from superclass" do
```
adrianna-chang-shopify, line 72: This test might be unnecessary now that we're using `class_attribute` - let me know if we should 🔥  it

etiennebarrie (approved):

app/tasks/maintenance_tasks/task.rb
```diff
@@ -72,6 +80,15 @@ def count
         new.count
       end
 
+      # Add a throttle spec for this Task.
+      #
+      # @param backoff [ActiveSupport::Duration] optionally, a custom backoff
+      #   can be specified. This is the time to wait before retrying the Task.
+      #   If no value is specified, it defaults to 30 seconds.
+      def throttle_on(backoff: 30.seconds, &condition)
```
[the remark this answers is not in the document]
adrianna-chang-shopify: Ah nice, yeah that makes sense 👍

README.md
```diff
@@ -112,6 +112,44 @@ title,content
...
+  end
+end
+```
+
+Note that it's up to you to define a throttling condition that makes sense for
+your app. Shopify implements `DatabaseStatus.healthy?` to check various MySQL
+metrics such as replication lag, DB threads, whether DB writes are available, 
+etc.
+
+Tasks can define multiple throttle conditions. Throttle conditions are inherited
```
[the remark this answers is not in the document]
adrianna-chang-shopify: I can mention it, doesn't hurt to be explicit!

test/dummy/app/tasks/maintenance/update_posts_throttled_task.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+module Maintenance
+  class UpdatePostsThrottledTask < MaintenanceTasks::Task
+    class << self
+      attr_accessor :throttle
```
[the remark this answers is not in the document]
adrianna-chang-shopify: IMHO this is useful mostly just for our tests. If we think we'd like a throttling task for local development in the future, we should probably write something with a more realistic throttling condition. I'm going to leave this for now because I'm not sure it warrants the time investment yet 😄

Issue #390, opened 2021-03-31
Support `ActiveRecord::Batches::BatchEnumerator` as a collection (or don't require a collection at all)


lamp:
Hi, while implementing [this job](https://github.com/Shopify/evs-shipping/blob/cdf431afc258dacc6c7c0fefd1098f85226f75ed/app/tasks/maintenance/copy_origin_postal_code_to_sample_zip_location_task.rb) for recent task, I ran into the requirement [here](https://github.com/Shopify/maintenance_tasks/blob/afadabb40a5cabdff8fc91faafb1af8c7be47c0d/app/jobs/maintenance_tasks/task_job.rb#L32) that I return a known collection type.

As you can see, to make my job work I had to provide an array with at least one value, it would be nice if I could have done:
```ruby
module Maintenance  
  class CopyOriginPostalCodeToSampleZipLocationTask < MaintenanceTasks::Task
    def collection
      ReconExtractBatchDetail.in_batches
    end

    def count
      ReconExtractBatchDetail.count
    end

    def process(batch)
        batch.update_all('sample_zip_location = origin_postal_code')
    end
  end
end
```

It looks like it should "just work" by adding the `BatchEnumerator` to the allowed list of classes, but I might be missing something, let me know if you have any questions and if needed I'd be happy to work on it with you.

Thanks a lot for the library, it works great other than this small issue.

Issue #401, opened 2021-05-04
Support batched collections


adrianna-chang-shopify:
Batch support has been requested by numerous users because it can make Tasks that need to do batch operations much more efficient.

We should allow tasks to define batch collections.

PR #409, opened 2021-05-10
Support `ActiveRecord::Batches::BatchEnumerator`  collections

adrianna-chang-shopify:
Closes: https://github.com/Shopify/maintenance_tasks/issues/390, #401

This PR allows users to specify batch collections via the following API:

```ruby
# app/tasks/maintenance/update_posts_in_batches_task.rb
module Maintenance
  class UpdatePostsInBatchesTask < MaintenanceTasks::Task
    def collection
      Post.in_batches
    end

    def count
      collection.count
    end

    def process(batch_of_posts)
      batch_of_posts.update_all(content: "New content added on #{Time.now.utc}")    
    end
  end
end
```

Users are expected to use [`#in_batches`](https://api.rubyonrails.org/v6.1.3.1/classes/ActiveRecord/Batches.html#method-i-in_batches). We use `EnumeratorBuilder#active_record_on_batch_relations ` under the hood to build an enumerator that yields `ActiveRecord::Relation`s - we get the ActiveRecord collection via `@relation` and the batch size via`@of` on the Batch Enumerator.

I've also tweaked `Ticker#tick` to take an increment, defaulting to 1. In `TaskJob#each_iteration`, I've tweaked the call to `#tick` to use input.size if input `is_a?(ActiveRecord::Relation)`. This makes it so that tick count is still based on the number of records, rather than the number of batches, if we're processing batches.

### DIFF
--- a/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
+++ b/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
@@ -32,6 +32,7 @@ module MaintenanceTasks
     def build_enumerator(_run, cursor:)
       cursor ||= @run.cursor
       collection = @task.collection
+      @enumerator = nil
 
       collection_enum = case collection
       when ActiveRecord::Relation
@@ -43,13 +44,13 @@ module MaintenanceTasks
             a batch enumerator with the "start" or "finish" options.
           MSG
         end
-        enumerator_builder.active_record_on_batch_relations(
+        # For now, only support automatic count based on the enumerator for
+        # batches
+        @enumerator = enumerator_builder.active_record_on_batch_relations(
           collection.relation,
           cursor: cursor,
           batch_size: collection.batch_size,
         )
-        @run.update!(tick_total: enumerator.size)
-        enumerator
       when Array
         enumerator_builder.build_array_enumerator(collection, cursor: cursor)
       when CSV
@@ -102,8 +103,9 @@ module MaintenanceTasks
     end
 
     def on_start
-      @run.tick_total = @task.count unless @run.tick_total
-      @run.update!(started_at: Time.now)
+      count = @task.count
+      count = @enumerator&.size if count == :no_count
+      @run.update!(started_at: Time.now, tick_total: count)
     end
 
     def on_complete
--- a/app/tasks/maintenance_tasks/task.rb
+++ b/app/tasks/maintenance_tasks/task.rb
@@ -131,6 +131,7 @@ module MaintenanceTasks
     #
     # @return [Integer, nil]
     def count
+      :no_count
     end
   end
 end
--- a/test/dummy/app/tasks/maintenance/update_posts_in_batches_task.rb
+++ b/test/dummy/app/tasks/maintenance/update_posts_in_batches_task.rb
@@ -5,10 +5,6 @@ module Maintenance
       Post.in_batches(of: 5)
     end
 
-    def count
-      Post.count
-    end
-
     def process(batch_of_posts)
       batch_of_posts.update_all(content: "New content added on #{Time.now.utc}")
     end
--- a/test/tasks/maintenance_tasks/task_test.rb
+++ b/test/tasks/maintenance_tasks/task_test.rb
@@ -54,9 +54,9 @@ module MaintenanceTasks
       assert_equal 2, Maintenance::TestTask.count
     end
 
-    test "#count is nil by default" do
+    test "#count is :no_count by default" do
       task = Task.new
-      assert_nil task.count
+      assert_equal(:no_count, task.count)
     end
 
     test "#collection raises NoMethodError" do


adrianna-chang-shopify:
@etiennebarrie I've requested your review on this. CI obviously fails because the Job Iteration changes haven't shipped yet, but tests pass with the changes from Shopify/job-iteration#86.

Was hoping for another review on the Job Iteration PR, but will ship tmrw if I don't hear from Job Patterns and put out an update. Figured there was no reason to wait on getting feedback on this 😄

app/jobs/concerns/maintenance_tasks/task_job_concern.rb
```diff
@@ -36,6 +36,15 @@ def build_enumerator(_run, cursor:)
       collection_enum = case collection
       when ActiveRecord::Relation
         enumerator_builder.active_record_on_records(collection, cursor: cursor)
+      when ActiveRecord::Batches::BatchEnumerator
+        relation = collection.instance_variable_get(:@relation)
+        batch_size = collection.instance_variable_get(:@of)
```
etiennebarrie, line 41: I think we want a test to ensure that a simple batch enumerator has those with the expected values. Additionally I think we should prevent the usage of `@start` and `@finish`, or support it properly

adrianna-chang-shopify: Ah yes good call! Let's prevent usage, I think it would be tricky to support them since we're using the base relation here to build the enumerator, which doesn't factor in the `@start` / `@finish` values.

etiennebarrie:
Hum, previously the `records.empty?` wouldn't have caused any additional query though because the records were loaded already. Since we're going to need at least one query to get the last record, I think we could change that to:
```ruby
last_record = records.last
update_from_record(last_record) if last_record
```
or change `update_from_record` to early return if given nil.

Today it's one query to load the records (in job-iteration) + one query for updating them (in the app code).
E.g. `records = relation.to_a` + `Post.where(id: posts.map(&:id)).update_all(published: true)`.

Ideally we'd move to one query to get the next cursor value + the app code.
`relation.pluck(@columns)` or something in job_iteration + `relation.update_all(published: true)`.

Hopefully even if it doesn't help much with the SQL, it helps by not loading so much stuff in memory?

adrianna-chang-shopify:
cc @Shopify/rails @Shopify/core-stewardship

app/jobs/concerns/maintenance_tasks/task_job_concern.rb
```diff
@@ -36,13 +36,27 @@ def build_enumerator(_run, cursor:)
       collection_enum = case collection
       when ActiveRecord::Relation
         enumerator_builder.active_record_on_records(collection, cursor: cursor)
+      when ActiveRecord::Batches::BatchEnumerator
+        if collection.instance_variable_get(:@start) ||
```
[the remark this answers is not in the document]
etiennebarrie: The plan also was to go to Rails and make the attributes public on `ActiveRecord::Batches::BatchEnumerator` so we could just pull them out.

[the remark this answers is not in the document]
adrianna-chang-shopify: Ah yes, that was also our plan, although I don't know if our use case is a justified reason to make those public.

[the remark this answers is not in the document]
rafaelfranca: In that case, do that first, and them backport the patch in this gem and use the public API.

rafaelfranca (approved):

lib/patches/active_record_batch_enumerator.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+# TODO: Remove this patch once https://github.com/rails/rails/pull/42312/commits/a031a43d969c87542c4ee8d0d338d55fcbb53376
```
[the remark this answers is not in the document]
adrianna-chang-shopify: True, I'll update the comment 👍

adrianna-chang-shopify:
cc @etiennebarrie

README.md
```diff
@@ -112,6 +112,39 @@ title,content
...
+
+    def process(batch_of_posts)
+      batch_of_posts.update_all(content: "New content added on #{Time.now.utc}")    
+    end
+  end
+end
+```
+
+Ensure that you've implemented the following methods:
+* `collection`: return an `ActiveRecord::Batches::BatchEnumerator`.
```
etiennebarrie, line 143:
```suggestion

* `collection`: return an `ActiveRecord::Batches::BatchEnumerator`.
```

README.md
```diff
@@ -112,6 +112,39 @@ title,content
...
+* `process`: do the work of your Task on a batch (`ActiveRecord::Relation`).
+* `count`: return the number of rows that will be iterated over (this should
+still be the number of records to be processed, not the number of batches).
```
etiennebarrie, line 146: Is it? Why not use the number of batches instead? I guess showing the number of records does look nicer.

app/jobs/concerns/maintenance_tasks/task_job_concern.rb
```diff
@@ -58,7 +69,8 @@ def build_enumerator(_run, cursor:)
     def each_iteration(input, _run)
       throw(:abort, :skip_complete_callbacks) if @run.stopping?
       task_iteration(input)
-      @ticker.tick
+      batch_size = input.is_a?(ActiveRecord::Relation) ? input.size : 1
```
etiennebarrie, line 72: This requires an unnecessary `COUNT(*)` to show records instead of batches in the progress, since the batch is not `loaded?` when it comes out of JobIteration.
Maybe we could somehow re-use the batch size and have some rounding error in the number of ticks recorded, or we should use the number of batches directly (and I'm not sure if we should explain that in the UI, but I think we have currently of persisting the information that the Task returned a relation in batches).

app/jobs/concerns/maintenance_tasks/task_job_concern.rb
```diff
@@ -36,13 +36,24 @@ def build_enumerator(_run, cursor:)
       collection_enum = case collection
       when ActiveRecord::Relation
         enumerator_builder.active_record_on_records(collection, cursor: cursor)
+      when ActiveRecord::Batches::BatchEnumerator
```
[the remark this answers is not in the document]
etiennebarrie: When you use a Relation, we batch load the records, but load all the columns and give the records to you.
For some use cases it's too slow, let's say you just added a boolean, want to set it to tru for all records before making it NOT NULL.
With relations, updating all records will results in b batch queries + b*batch_size updates (one UPDATE per record)
```ruby
def collection
  Post.all
end
def process(post)
  post.update(published: true)
end
```
With true batches, updating all records will result in b batch queries (we still need those for the cursors) + b batch updates (one UPDATE per batch).
```ruby
def collection
  Post.in_batches
end
def process(posts)
  posts.update_all(published: true)
end
```

lib/patches/active_record_batch_enumerator.rb
```diff
@@ -0,0 +1,23 @@
...
+  # The relation from which the BatchEnumerator yields batches.
+  attr_reader :relation
+
+  # The size of the batches yielded by the BatchEnumerator.
+  def batch_size
+    @of
+  end
+end
+
+ActiveRecord::Batches::BatchEnumerator.include(ActiveRecordBatchEnumerator)
```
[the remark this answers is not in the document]
etiennebarrie: Active Record is a dependency because we use it for the Run model so I think it's necessarily loaded?
https://github.com/Shopify/maintenance_tasks/blob/63ab1b104f0ec3deaca50d4c9919bf8d57f15398/lib/maintenance_tasks.rb#L5
https://github.com/Shopify/maintenance_tasks/blob/63ab1b104f0ec3deaca50d4c9919bf8d57f15398/lib/maintenance_tasks/engine.rb#L2

app/jobs/concerns/maintenance_tasks/task_job_concern.rb
etiennebarrie: Yes we can, there would be some rounding error where if there's 7 records and you have batches of 5, you'd see 10 records processed at the end.

adrianna-chang-shopify: Changes incoming! 😁

README.md
```diff
@@ -112,6 +112,44 @@ title,content
...
+
+Note that `#count` will be calculated for you based on the number of batches in
+your collection, and your Task's progress will be displayed in terms of batches
+(not the number of records in the relation).
+
+**Important!** Batches should only be used if `#process` is performing a batch
+operation such as `#update_all` or `#delete_all`, and should **not** be used
+if you need to iterate over the records in the batch. If you need to iterate
+over individual records, it is more performant to define a collection that
+[directly returns an `ActiveRecord::Relation`](#creating-a-task).
```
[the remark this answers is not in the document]
etiennebarrie:
```suggestion
**Important!** Batches should only be used if `#process` is performing a batch
operation such as `#update_all` or `#delete_all`. If you need to iterate over
individual records, you should define a collection that [directly returns an
`ActiveRecord::Relation`](#creating-a-task). This will also uses batching
internally, but load the records with only one SQL query, whereas Batch
collections will load the primary keys first and then the records, if you call
`each` or any `Enumerable` method.
```

README.md
```diff
@@ -112,6 +112,44 @@ title,content
...
+```ruby
+# app/tasks/maintenance/update_posts_in_batches_task.rb
+module Maintenance
+  class UpdatePostsInBatchesTask < MaintenanceTasks::Task
+    def collection
+      Post.in_batches
+    end
+
+    def process(batch_of_posts)
+      batch_of_posts.update_all(content: "New content added on #{Time.now.utc}")
```
etiennebarrie, line 132:
```suggestion
      batch_of_posts.update_all(content: "New content added on #{Time.now.utc}")
```

app/jobs/concerns/maintenance_tasks/task_job_concern.rb
```diff
@@ -36,13 +36,30 @@ def build_enumerator(_run, cursor:)
...
+            #{@task.class.name}#collection cannot support
+            a batch enumerator with the "start" or "finish" options.
+          MSG
+        end
+        enumerator = enumerator_builder.active_record_on_batch_relations(
+          collection.relation,
+          cursor: cursor,
+          batch_size: collection.batch_size,
+        )
+        @run.update!(tick_total: enumerator.size)
```
etiennebarrie, line 51: I think we may be able to skip persisting it here and let the rest of the job (the start callback) do it.

etiennebarrie, line 51: `build_enumerator` is called each time the run resumes after being interrupted, so we shouldn't recompute the tick total each time.

etiennebarrie, line 51: Like I mentioned above, it should be possible to avoid calling `ActiveRecordBatchEnumerator#size` which is not in their control.

README.md
```diff
@@ -112,6 +112,46 @@ title,content
 My Title,Hello World!
 ```
 
+### Processing Batch Collections
+
+The Maintenance Tasks gem supports processing Active Records in batches. This
+can reduce the number of calls your Task makes to the database. Use
+`ActiveRecord::Batches#in_batches` to specify that your Task should process
+records in batches. Active Record defaults to 1000 records, but a custom size can be
```
etiennebarrie, line 120:
```suggestion
`ActiveRecord::Batches#in_batches` on the relation returned by your collection to specify that your Task should process
batches instead of records. Active Record defaults to 1000 records by batch, but a custom size can be
```

README.md
```diff
@@ -112,6 +112,46 @@ title,content
...
+Note that `#count` is calculated automatically based on the number of batches in
+your collection, and your Task's progress will be displayed in terms of batches
+(not the number of records in the relation).
+
+**Important!** Batches should only be used if `#process` is performing a batch
+operation such as `#update_all` or `#delete_all`. If you need to iterate over
+individual records, you should define a collection that [returns an
+`ActiveRecord::Relation`](#creating-a-task). This uses batching
+internally, but loads the records with one SQL query. Conversely, batch
+collections load the primary keys first, and then perform a query to load the
```
etiennebarrie, line 152:
```suggestion
collections load the primary keys of the records of the batch first, and then perform an additional query to load the
```

etiennebarrie (approved):
Some more wording propositions, feel free to tweak. The rest is great 🎉 :shipit:

Issue #406, opened 2021-05-10
Explore using ActiveModel to power Tasks with parameters


adrianna-chang-shopify:
We've had a number of requests to support Tasks with parameters (more info in https://github.com/Shopify/maintenance_tasks/issues/325). A simple, Rails-flavoured approach to this could be to have Tasks include `ActiveModel::Model` and use attributes for parameters, which would integrate nicely with the UI and also allows Task authors to specify validations on their Task parameters.

Issue #325, opened 2021-02-09
arrive-server use cases


adrianna-chang-shopify:
Moving the conversation from https://github.com/Shopify/super_maintenance/issues/8#issuecomment-775932361 over to our existing repo

From @tgwizard :
> 👋  I want to describe a use-case that I'm not sure fits well with the current CSV Task structure. In (app), we often run migrations and backfills on a large dataset. When we write such migrations, we often want to try them out on a subset of data. This can be all orders for a specific user, or the first 10k orders with id > X. We may often want to run them in parallel too (at least the process part), which could potentially be done by launching concurrent maintenance tasks on separate subsets of the data

> Is that possible to do with a CSV Task? Seems a bit hard to do, since for CSV tasks only the process method seems like it should be implemented, whereas I imagine the parameters being applicable to the collection method.

> (I imagine we'd write 2 separate maintenance tasks for the two flows above, one for single user, one for id + limit).

> Is there some other way we could write these kind of migrations using the maintenance tasks gem?

adrianna-chang-shopify:
Hey @tgwizard , can you tell me a bit more about why you want to use a CSV for these use cases? You can run a Task with a subset of a large db table by specifying an ActiveRecord relation for `collection`:

```ruby
module Maintenance
  class SubsetOrderTask < MaintenanceTasks::Task
    def collection
        Order.where(user_id: 123)
        # or, Order.where(id: 10000..20000)
    end

    def count
      collection.count
    end

    def process(order)
      # Do something
    end
  end
end
```

As for concurrency, it's definitely something that's on our radar, but the gem isn't able to handle that yet. You can indeed write separate tasks to handle different subsets of the data and run them concurrently. I'm realizing now that CSV Tasks might have been appealing because you can customize which collection the Task is processing? So yes, if you wanted to, you could export a CSV with ids for all the orders you wanted to process, and then upload this to a CSV-processing Task. This way you could reuse a single Task across multiple datasets BUT the drawback is that you can only run a single instance of a Task at a time, so you wouldn't be able to achieve this concurrently. (So if concurrency is important, your best bet right now would be to just write separate Tasks for different datasets).

Let me know if this answers your question(s) or if I've misunderstood anything 😄

tgwizard:
Hmm, right, that `SubsetOrderTask` conceptually exactly what I want to do, but I don't want to write a specific task for each user we'd run it for.

[...]

PR #413, opened 2021-05-13
Support Tasks with custom parameters

adrianna-chang-shopify:
Closes: https://github.com/Shopify/maintenance_tasks/issues/406, https://github.com/Shopify/maintenance_tasks/issues/325

This PR introduces support for custom Task parameters by leveraging `ActiveModel` attributes & validations.

Task authors use `ActiveModel::Attributes` for their parameters, and can define their own validations to ensure any parameters that are passed from the UI conform to their expectations (ie. `Maintenance::ParamsTask` uses a presence validator and a regex to ensure that post ids is a comma-delimited list of integers).

## Approach taken
- `MaintenanceTasks::Task` includes relevant `ActiveModel` modules
- In the view, we render a text field for every parameter the Task supports
- These params are passed from controller => `Runner#run`. `MaintenanceTasks::Run` has a field for `parameters` now, and we pass the parameters to the constructor
- On run creation, we assign the parameters (as attributes) to a `Task` instance for the `run`, and then validate whether the attributes are valid. If they aren't, we take the errors and add them to the `run`, which gets propagated to the UI in the form of a flash alert.
- The `run` how has parameters persisted, and we've also set the parameters on the `Task` so that any of the methods in the `Task` (`#process`, `#collection`, `#count`) now have access to these attribute

## 🎩 
Screen Shot 2021-05-13 at 3 13 38 PM

Screen Shot 2021-05-13 at 4 00 18 PM

- Go to http://127.0.0.1:3000/maintenance_tasks/tasks/Maintenance::ParamsTask
- Run it with `1,2,3` in the `Post ids` field
- Run it with an invalid string in the `Post ids` field
- Run it with no value in the `Post ids` field

## To Do
- [x] Documentation

Let me know what you think!

### DIFF
--- a/app/controllers/maintenance_tasks/tasks_controller.rb
+++ b/app/controllers/maintenance_tasks/tasks_controller.rb
@@ -27,7 +27,8 @@ module MaintenanceTasks
     def run
       task = Runner.run(
         name: params.fetch(:id),
-        csv_file: params[:csv_file]
+        csv_file: params[:csv_file],
+        arguments: task_arguments,
       )
       redirect_to(task_path(task))
     rescue ActiveRecord::RecordInvalid => error
@@ -37,6 +38,12 @@ module MaintenanceTasks
     end
 
     private
+
+    def task_arguments
+      return {} unless params[:task_arguments].present?
+      task_attributes = Task.named(params[:id]).attribute_names
+      params.require(:task_arguments).permit(*task_attributes).to_h
+    end
 
     def set_refresh
       @refresh = 3
--- a/app/controllers/maintenance_tasks/tasks_controller.rb
+++ b/app/controllers/maintenance_tasks/tasks_controller.rb
@@ -33,6 +33,9 @@ module MaintenanceTasks
       redirect_to(task_path(task))
     rescue ActiveRecord::RecordInvalid => error
       redirect_to(task_path(error.record.task_name), alert: error.message)
+    rescue ActiveRecord::ValueTooLong => error
+      task_name = params.fetch(:id)
+      redirect_to(task_path(task_name), alert: error.message)
     rescue Runner::EnqueuingError => error
       redirect_to(task_path(error.run.task_name), alert: error.message)
     end
--- a/app/models/maintenance_tasks/run.rb
+++ b/app/models/maintenance_tasks/run.rb
@@ -38,6 +38,7 @@ module MaintenanceTasks
     attr_readonly :task_name
 
     serialize :backtrace
+    serialize :arguments, JSON
 
     scope :active, -> { where(status: ACTIVE_STATUSES) }
 
--- a/app/models/maintenance_tasks/run.rb
+++ b/app/models/maintenance_tasks/run.rb
@@ -34,6 +34,7 @@ module MaintenanceTasks
       Task.available_tasks.map(&:to_s)
     } }
     validate :csv_attachment_presence, on: :create
+    validate :validate_task_arguments, on: :create
 
     attr_readonly :task_name
 
@@ -207,6 +208,40 @@ module MaintenanceTasks
       nil
     end
 
+    # Support iterating over ActiveModel::Errors in Rails 6.0 and Rails 6.1+.
+    # To be removed when Rails 6.0 is no longer supported.
+    if Rails::VERSION::STRING.match?(/^6.0/)
+      # Performs validation on the arguments to use for the Task. If the Task is
+      # invalid, the errors are added to the Run.
+      def validate_task_arguments
+        if task.invalid?
+          error_messages = task.errors
+            .map { |attribute, message| "#{attribute.inspect} #{message}" }
+          errors.add(
+            :arguments,
+            "are invalid: #{error_messages.join("; ")}"
+          )
+        end
+      rescue Task::NotFoundError
+        nil
+      end
+    else
+      # Performs validation on the arguments to use for the Task. If the Task is
+      # invalid, the errors are added to the Run.
+      def validate_task_arguments
+        if task.invalid?
+          error_messages = task.errors
+            .map { |error| "#{error.attribute.inspect} #{error.message}" }
+          errors.add(
+            :arguments,
+            "are invalid: #{error_messages.join("; ")}"
+          )
+        end
+      rescue Task::NotFoundError
+        nil
+      end
+    end
+
     # Fetches the attached ActiveStorage CSV file for the run. Checks first
     # whether the ActiveStorage::Attachment table exists so that we are
     # compatible with apps that are not using ActiveStorage.
@@ -216,6 +251,22 @@ module MaintenanceTasks
       return unless defined?(ActiveStorage)
       return unless ActiveStorage::Attachment.table_exists?
       super
+    end
+
+    # Returns a Task instance for this Run. Assigns any attributes to the Task
+    # based on the Run's parameters. Note that the Task instance is not supplied
+    # with :csv_content yet if it's a CSV Task. This is done in the job, since
+    # downloading the CSV file can take some time.
+    #
+    # @return [Task] a Task instance.
+    def task
+      @task ||= begin
+        task = Task.named(task_name).new
+        if task.attribute_names.any? && arguments.present?
+          task.assign_attributes(arguments)
+        end
+        task
+      end
     end
   end
 end
--- a/app/models/maintenance_tasks/runner.rb
+++ b/app/models/maintenance_tasks/runner.rb
@@ -45,6 +45,8 @@ module MaintenanceTasks
     # @raise [EnqueuingError] if an error occurs while enqueuing the Run.
     # @raise [ActiveRecord::RecordInvalid] if validation errors occur while
     #   creating the Run.
+    # @raise [ActiveRecord::ValueTooLong] if the creation of the Run fails due
+    #   to a value being too long for the column type.
     def run(name:, csv_file: nil, arguments: {})
       run = Run.active.find_by(task_name: name) ||
         Run.new(task_name: name, arguments: arguments)
--- a/app/models/maintenance_tasks/runner.rb
+++ b/app/models/maintenance_tasks/runner.rb
@@ -37,14 +37,17 @@ module MaintenanceTasks
     #   for the Task to iterate over when running, in the form of an attachable
     #   (see https://edgeapi.rubyonrails.org/classes/ActiveStorage/Attached/One.html#method-i-attach).
     #   Value is nil if the Task does not use CSV iteration.
+    # @param arguments [Hash] the arguments to persist to the Run and to make
+    #   accessible to the Task.
     #
     # @return [Task] the Task that was run.
     #
     # @raise [EnqueuingError] if an error occurs while enqueuing the Run.
     # @raise [ActiveRecord::RecordInvalid] if validation errors occur while
     #   creating the Run.
-    def run(name:, csv_file: nil)
-      run = Run.active.find_by(task_name: name) || Run.new(task_name: name)
+    def run(name:, csv_file: nil, arguments: {})
+      run = Run.active.find_by(task_name: name) ||
+        Run.new(task_name: name, arguments: arguments)
       run.csv_file.attach(csv_file) if csv_file
 
       run.enqueued!
--- a/app/models/maintenance_tasks/task_data.rb
+++ b/app/models/maintenance_tasks/task_data.rb
@@ -132,6 +132,15 @@ module MaintenanceTasks
       !deleted? && Task.named(name) < CsvCollection
     end
 
+    # @return [Array<String>] the names of parameters the Task accepts.
+    def parameter_names
+      if deleted?
+        []
+      else
+        Task.named(name).attribute_names
+      end
+    end
+
     private
 
     def runs
--- a/app/tasks/maintenance_tasks/task.rb
+++ b/app/tasks/maintenance_tasks/task.rb
@@ -3,6 +3,9 @@ module MaintenanceTasks
   # Base class that is inherited by the host application's task classes.
   class Task
     extend ActiveSupport::DescendantsTracker
+    include ActiveModel::Attributes
+    include ActiveModel::AttributeAssignment
+    include ActiveModel::Validations
 
     class NotFoundError < NameError; end
 
--- a/app/views/maintenance_tasks/tasks/show.html.erb
+++ b/app/views/maintenance_tasks/tasks/show.html.erb
@@ -16,6 +16,16 @@
           <%= form.file_field :csv_file %>
         </div>
       <% end %>
+      <% if @task.parameter_names.any? %>
+        <div class="block">
+          <%= form.fields_for :task_arguments do |ff| %>
+            <% @task.parameter_names.each do |parameter| %>
+              <%= ff.label parameter, "#{parameter}: ", class: "label" %>
+              <%= ff.text_area parameter, class: "textarea" %>
+            <% end %>
+          <% end %>
+        </div>
+      <% end %>
       <div class="block">
         <%= form.submit 'Run', class: "button is-success", disabled: @task.deleted? %>
       </div>
--- a/db/migrate/20210517131953_add_arguments_to_maintenance_tasks_runs.rb
+++ b/db/migrate/20210517131953_add_arguments_to_maintenance_tasks_runs.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+class AddArgumentsToMaintenanceTasksRuns < ActiveRecord::Migration[6.0]
+  def change
+    add_column(:maintenance_tasks_runs, :arguments, :text)
+  end
+end
--- a/test/application_system_test_case.rb
+++ b/test/application_system_test_case.rb
@@ -20,11 +20,13 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
   setup do
     travel_to Time.zone.local(2020, 1, 9, 9, 41, 44)
     Maintenance::UpdatePostsTask.fast_task = false
+    Maintenance::ParamsTask.fast_task = false
   end
 
   teardown do
     assert_empty page.driver.browser.manage.logs.get(:browser)
     Maintenance::UpdatePostsTask.fast_task = true
+    Maintenance::ParamsTask.fast_task = true
     FileUtils.rm_rf("test/dummy/tmp/downloads")
   end
 end
--- a/test/application_system_test_case.rb
+++ b/test/application_system_test_case.rb
@@ -19,14 +19,10 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
 
   setup do
     travel_to Time.zone.local(2020, 1, 9, 9, 41, 44)
-    Maintenance::UpdatePostsTask.fast_task = false
-    Maintenance::ParamsTask.fast_task = false
   end
 
   teardown do
     assert_empty page.driver.browser.manage.logs.get(:browser)
-    Maintenance::UpdatePostsTask.fast_task = true
-    Maintenance::ParamsTask.fast_task = true
     FileUtils.rm_rf("test/dummy/tmp/downloads")
   end
 end
--- a/test/dummy/app/tasks/maintenance/params_task.rb
+++ b/test/dummy/app/tasks/maintenance/params_task.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+module Maintenance
+  class ParamsTask < MaintenanceTasks::Task
+    attribute :post_ids, :string
+
+    validates :post_ids,
+      presence: true,
+      format: { with: /\A(\s?\d+(,\s?\d+\s?)*)\z/, allow_blank: true }
+
+    class << self
+      attr_accessor :fast_task
+    end
+
+    def collection
+      Post.where(id: post_ids_array)
+    end
+
+    def count
+      collection.count
+    end
+
+    def process(post)
+      sleep(1) unless self.class.fast_task
+
+      post.update!(content: "New content added on #{Time.now.utc}")
+    end
+
+    private
+
+    def post_ids_array
+      post_ids.split(",")
+    end
+  end
+end
--- a/test/dummy/db/schema.rb
+++ b/test/dummy/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_02_25_152418) do
+ActiveRecord::Schema.define(version: 2021_05_17_131953) do
 
   create_table "active_storage_attachments", force: :cascade do |t|
     t.string "name", null: false
@@ -55,7 +55,8 @@ ActiveRecord::Schema.define(version: 2021_02_25_152418) do
     t.text "backtrace"
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
-    t.index ["task_name", "created_at"], name: "index_maintenance_tasks_runs_on_task_name_and_created_at", order: { created_at: :desc }
+    t.text "arguments"
+    t.index ["task_name", "created_at"], name: "index_maintenance_tasks_runs_on_task_name_and_created_at"
   end
 
   create_table "posts", force: :cascade do |t|
--- a/test/jobs/maintenance_tasks/task_job_test.rb
+++ b/test/jobs/maintenance_tasks/task_job_test.rb
@@ -355,5 +355,18 @@ module MaintenanceTasks
 
       assert_predicate run.reload, :succeeded?
     end
+
+    test ".perform_now makes arguments supplied for Task parameters available" do
+      post = Post.last
+      Maintenance::ParamsTask.any_instance.expects(:process).once.with(post)
+
+      run = Run.create!(
+        task_name: "Maintenance::ParamsTask",
+        arguments: { post_ids: post.id.to_s }
+      )
+      TaskJob.perform_now(run)
+
+      assert_predicate run.reload, :succeeded?
+    end
   end
 end
--- a/test/models/maintenance_tasks/run_test.rb
+++ b/test/models/maintenance_tasks/run_test.rb
@@ -5,19 +5,27 @@ module MaintenanceTasks
   class RunTest < ActiveSupport::TestCase
     test "invalid if the task doesn't exist" do
       run = Run.new(task_name: "Maintenance::DoesNotExist")
-      refute run.valid?
+      refute_predicate run, :valid?
     end
 
     test "invalid if associated with CSV Task and no attachment" do
       run = Run.new(task_name: "Maintenance::ImportPostsTask")
-      refute run.valid?
+      refute_predicate run, :valid?
     end
 
     test "invalid if unassociated with CSV Task and attachment" do
       run = Run.new(task_name: "Maintenance::UpdatePostsTask")
       csv = Rack::Test::UploadedFile.new(file_fixture("sample.csv"), "text/csv")
       run.csv_file.attach(csv)
-      refute run.valid?
+      refute_predicate run, :valid?
+    end
+
+    test "invalid if associated Task has parameters and they are invalid" do
+      run = Run.new(
+        task_name: "Maintenance::ParamsTask",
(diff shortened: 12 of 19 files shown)


app/controllers/maintenance_tasks/tasks_controller.rb
```diff
@@ -27,7 +27,8 @@ def show
     def run
       task = Runner.run(
         name: params.fetch(:id),
-        csv_file: params[:csv_file]
+        csv_file: params[:csv_file],
+        params: task_params[:task_params],
```
etiennebarrie, line 31: Why don't we move all the params to `task_params`?

app/models/maintenance_tasks/task_data.rb
```diff
@@ -132,6 +132,15 @@ def csv_task?
       !deleted? && Task.named(name) < CsvCollection
     end
 
+    # @return [Array<String>] the names of parameters the Task accepts.
+    def params
```
etiennebarrie, line 136: Based on the documentation, shouldn't it be `parameter_names`?

test/dummy/app/tasks/maintenance/params_task.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+module Maintenance
+  class ParamsTask < MaintenanceTasks::Task
+    attribute :post_ids, :string
+
+    validates :post_ids, presence: true
+    validates_format_of :post_ids, with: /\A([0-9]+(,\s?[0-9]+)*)\z/,
```
etiennebarrie, line 7: We could use `\d` for `[0-9]`.
Why are spaces allowed after the comma but not before, and not at the end or beginning either? 😄 
-`1 ,2`-` 1,2`-`1,2 `-

test/jobs/maintenance_tasks/task_job_test.rb
```diff
@@ -13,7 +13,7 @@ class TaskJobTest < ActiveJob::TestCase
         @run.pausing!
       end
 
-      TaskJob.perform_now(@run)
+      TaskJob.perform_now(@run, {})
```
etiennebarrie, line 16: Should we default to `{}` so that we don't have to change all the tests?

app/tasks/maintenance_tasks/task.rb
```diff
@@ -3,6 +3,8 @@ module MaintenanceTasks
   # Base class that is inherited by the host application's task classes.
   class Task
     extend ActiveSupport::DescendantsTracker
+    include ActiveModel::Attributes
+    include ActiveModel::Model
```
etiennebarrie, line 7: Maybe we don't need `ActiveModel::Conversion`, or `#initialize` and `persisted?` from `ActiveSupport::Model`, so maybe we could only use what the modules we need.

etiennebarrie (commented):
I think we're missing a CSS class, the field is almost invisible:

image

> Does it make sense to start the job and raise if parameters are invalid?

Potentially the validations would be quite costly, so I think it's ok to make them in `before_perform`. In the mean time, if the validations are costly, it also means we do them at each interruption. Validations have a concept of context which we could use to decide which ones to run when.

I guess we could assume they're the same kind of validations that would happen in a regular controller flow, without any job involved, for now. If we realize there's a need for deeper more intensive validations, we could add them in the `on_start` callback to run only once.

---

Looks great! I have a few comments but overall really neat.

app/views/maintenance_tasks/tasks/show.html.erb
```diff
@@ -16,6 +16,16 @@
           <%= form.file_field :csv_file %>
         </div>
       <% end %>
+      <% if @task.params.any? %>
+        <div class="block">
+          <%= form.fields_for :task_params do |ff| %>
+            <% @task.params.each do |param| %>
+              <%= ff.label "#{param}: ", class: "label" %>
```
samuelgiles, line 23: I'm not sure if the label would tie to the text field correctly as it stands? We may need to do this:

```suggestion
              <%= ff.label param, "#{param}:", class: "label" %>
```

test/dummy/app/tasks/maintenance/params_task.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+module Maintenance
+  class ParamsTask < MaintenanceTasks::Task
+    attribute :post_ids, :string
```
samuelgiles, line 4: Nice!

samuelgiles (commented):
This is exciting! We were talking in https://github.com/Shopify/arrive-server/pull/17707 about how we needed `maintenance_tasks` to support parameters and now its here 👏

test/dummy/app/tasks/maintenance/params_task.rb
adrianna-chang-shopify: Yeah, we can make any of those valid 😂  I mostly just wanted it to work with submitting `1, 2, 3`

app/views/maintenance_tasks/tasks/show.html.erb
adrianna-chang-shopify: Actually, even in omitting the name, it seems to match the label to the text field 😄 
Screen Shot 2021-05-14 at 10 25 27 AM
I guess that's typical ✨ Rails Magic ✨ for ya.

adrianna-chang-shopify:
> I think we're missing a CSS class, the field is almost invisible:

@etiennebarrie what browser are you using? It renders with a dark outline for me in Chrome:

Screen Shot 2021-05-14 at 10 34 37 AM

> I guess we could assume they're the same kind of validations that would happen in a regular controller flow, without any job involved, for now. If we realize there's a need for deeper more intensive validations, we could add them in the on_start callback to run only once.

Good point, I overlooked the fact that these validations would run every time the job interrupts. But, this is potentially a benefit as you mentioned, since they might be evaluated in to a context that changes based on when the job first runs vs when it's being resumed. Let's leave it in `#before_perform` for now, and revisit moving it to `#on_start` if they are proving costly.

adrianna-chang-shopify:
@dirceu that's a great idea! The whole reason I went with `post_ids` for the test Task was because I noticed that Arrays of Integers was one of the most frequent use cases. 😁  It definitely makes sense to simplify this for users and offer them an API that takes care of these details for them, even if `ActiveModel::Attributes` won't give us Array support directly.

Lemme see what I can whip up

etiennebarrie:
> what browser are you using?

Firefox!

> Let's leave it in `#before_perform` for now

I feel like this is the least interesting choice. Either we're worried about the potential cost of validations and we do it once before the whole run, or we're not and we can do it at the controller level. "Validation" code can run inside `collection` to run at each interruption, or inside `process` to run for each item. So I think overall the most useful place for us to run the validations is directly before creating a Run in the controller. We should just document it clearly so that it's clear.

> Would it make sense to have first-class support for array of string and array of numbers

Yeah that would be cool, but either we introduce our own DSL (instead of being a simple Active Model), or we add a type to `ActiveModel::Registry`, which is too invasive for a gem IMO. The last solution I see would be to have an extensive example creating a Type, but that's a bit overblown I think. If we're worried about the regex being wrongly copied, we could remove it 😄 

Active Record deals pretty well with invalid data:
```ruby
Post.where(id: "2,foo,4".split(",")).to_sql
# => "SELECT \"posts\".* FROM \"posts\" WHERE \"posts\".\"id\" IN (2, 4)"
```

---

One other thing I had missed is that this doesn't persist the parameters, so we can't pause/resume unless I'm missing something?

adrianna-chang-shopify:
Okay, might be a Bulma + Firefox thing in that case @etiennebarrie ?

Gotcha, yeah I misinterpreted what you were getting at before re: validations needing to run in certain contexts. Let's just move it into the controller for now.

In terms of types, IMO a nice way to do it would be to to add `ActiveModel::Type::Value` classes that Task authors can use if they want, without actually adding them to the registry. This is what I've been mocking up:

```ruby
module MaintenanceTasks
  module Parameters
    class IntegerArrayType < ActiveModel::Type::Value
      def cast(input)
        return unless input.present?

        validate_input(input)
        input.split(",").map(&:to_i)
      end

      private

      def validate_input(input)
        unless /\A(\s?\d+(,\s?\d+\s?)*\s?)\z/.match?(input)
          error_message = "IntegerArrayType expects expects a comma-delimited "\
            "string of integers. Input received: #{input}"
          raise TypeError.new(error_message)
        end
      end
    end
  end
end
```

```ruby
module Maintenance
  class ParamsTask < MaintenanceTasks::Task
    attribute :post_ids, MaintenanceTasks::Parameters::IntegerArrayType.new
    validates :post_ids, presence: true
    ....
```

IMHO this is a _much_ better experience than requiring every user to validate / sanitize input strings from the UI or write their own type.

> One other thing I had missed is that this doesn't persist the parameters, so we can't pause/resume unless I'm missing something?

LOL I completely overlooked this 😆  Should we persist params on the run as well? It might be a nice feature to be able to see that data historically in the UI as well (beyond just logging params, making them visible in the UI should increase confidence in terms of being able to track down what happened if someone ran a Task with faulty params)

app/models/maintenance_tasks/runner.rb
```diff
@@ -37,14 +37,17 @@ def initialize(run)
     #   for the Task to iterate over when running, in the form of an attachable
     #   (see https://edgeapi.rubyonrails.org/classes/ActiveStorage/Attached/One.html#method-i-attach).
     #   Value is nil if the Task does not use CSV iteration.
+    # @param params [Hash] the parameters to make accessible to the Task and to
+    #   persist to the Run.
```
etiennebarrie, line 41: Maybe we could put that in chronological order, first the arguments are persisted to the run, and then they're made accessible to the Task?

app/views/maintenance_tasks/tasks/show.html.erb
```diff
@@ -16,6 +16,16 @@
           <%= form.file_field :csv_file %>
         </div>
       <% end %>
+      <% if @task.parameter_names.any? %>
+        <div class="block">
+          <%= form.fields_for :task_params do |ff| %>
+            <% @task.parameter_names.each do |param| %>
```
etiennebarrie, line 22:
```suggestion
            <% @task.parameter_names.each do |parameter| %>
```

app/models/maintenance_tasks/run.rb
```diff
@@ -206,6 +208,21 @@ def csv_attachment_presence
...
 
+    # Performs validation on the parameters to use for the Task. Attributes are
+    # assigned to a Task instance based on the parameters passed to the Run's
+    # constructor. If the Task is invalid, the errors are added to the Run.
+    def valid_task_parameters?
+      if task.invalid?
+        error_messages = task.errors.full_messages
+        errors.add(
+          :parameters,
+          "are invalid: #{error_messages.join("; ")}"
```
etiennebarrie, line 219: I found that the default error messages weren't great because of the call to `humanize`:
image

test/dummy/app/tasks/maintenance/params_task.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    validates :post_ids, presence: true
+    validates_format_of :post_ids, with: /\A(\s?\d+(,\s?\d+\s?)*)\z/,
+      allow_blank: true
```
etiennebarrie, line 8: Maybe we could combine them?
```suggestion
    validates :post_ids,
      presence: true,
      format: { with: /\A(\s?\d+(,\s?\d+\s?)*)\z/, allow_blank: true }
```

test/dummy/app/tasks/maintenance/params_task.rb
```diff
@@ -0,0 +1,30 @@
...
+    def collection
+      Post.where(id: post_ids_array)
+    end
+
+    def count
+      collection.count
+    end
+
+    def process(post)
+      sleep(1)
```
etiennebarrie, line 19: We should add a `fast_task` option similar to `UpdatePostsTask`.

etiennebarrie (commented):
Two main things:
* CLI support
* parameters vs arguments naming

But it's looking great! ✨

app/controllers/maintenance_tasks/tasks_controller.rb
```diff
@@ -38,6 +39,12 @@ def run
 
     private
 
+    def task_params
+      return {} unless params[:task_params].present?
+      task_attributes = Task.named(params[:id]).attribute_names
+      params.require(:task_params).permit(*task_attributes).to_h
+    end
+
```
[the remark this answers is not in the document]
adrianna-chang-shopify: This is the point of Strong Parameters though, no? To validate that inputs from the UI conform to what we expect? With the CLI, the point of failure will be `task.assign_attributes` in `run.rb` when validating the run on create, and that's probably enough, but with the controller, I feel like it makes sense to have this extra level of control and to fail sooner if an invalid key is supplied.

app/models/maintenance_tasks/run.rb
adrianna-chang-shopify: Do you have suggestions for a better message? 😄

test/dummy/app/tasks/maintenance/params_task.rb
adrianna-chang-shopify: Sure, this is going to change with https://github.com/Shopify/maintenance_tasks/pull/414 though 😉

app/controllers/maintenance_tasks/tasks_controller.rb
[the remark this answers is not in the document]
adrianna-chang-shopify: I'd like to do a follow up for CLI support, we can chat more about it then?

app/models/maintenance_tasks/run.rb
```diff
@@ -216,5 +233,19 @@ def csv_file
       return unless ActiveStorage::Attachment.table_exists?
       super
     end
+
+    # Returns a Task instance for this Run. Assigns any attributes to the Task
+    # based on the Run's parameters.
```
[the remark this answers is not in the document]
etiennebarrie: It seemed to me the only place we called `run.task` was in the job, just before setting the `csv_content`. But I see now it's also called from the validation.
It just seems a bit misleading for `Run#task` to return a Task that's only half-ready to be used.
I think alternatively, there could be a method only focused on setting the task attributes based on the run (and something else would instantiate the task, maybe `valid_task_arguments?` in that case), or maybe it should take an option to skip downloading the CSV maybe so that it's more thruthful (= "this is the task instance ready for this run, but you have an option to not download the CSV").
Or we could keep it like this, Run#task gives you a Task instance that is ready, just missing CSV, and we do the remaining work in the job. We could mention it in the comment maybe.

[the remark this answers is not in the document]
adrianna-chang-shopify: Ah, makes sense! I think I'd prefer to always explicitly download the CSV in the job, but I'm on board with naming `Run#task` something that indicates it's really only setting + validating the arguments :D

app/models/maintenance_tasks/run.rb
```diff
@@ -206,6 +208,24 @@ def csv_attachment_presence
...
 
+    # Performs validation on the arguments to use for the Task. If the Task is
+    # invalid, the errors are added to the Run.
+    def valid_task_arguments?
+      if task.invalid?
+        error_messages = []
+        task.errors.messages.each do |attribute, messages|
+          error_messages +=
+            messages.map { |message| "#{attribute.inspect} #{message}" }
+        end
```
adrianna-chang-shopify, line 219: This is unfortunately the only way to do it that is compatible with both Rails 6.0 and Rails 6.1 (since in Rails 6, error messages are [yielded as hashes](https://api.rubyonrails.org/v6.0/classes/ActiveModel/Errors.html#method-i-each))

test/application_system_test_case.rb
```diff
@@ -20,11 +20,13 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
   setup do
     travel_to Time.zone.local(2020, 1, 9, 9, 41, 44)
     Maintenance::UpdatePostsTask.fast_task = false
+    Maintenance::ParamsTask.fast_task = false
```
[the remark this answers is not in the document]
etiennebarrie: Hum interesting. I think we want to be fast all the time right? I don't think we need to set either to false here.

app/models/maintenance_tasks/run.rb
etiennebarrie: Good catch.

etiennebarrie:
> we could serialize it to JSON if you'd prefer

Maybe that would make more sense 🤔

app/models/maintenance_tasks/run.rb
```diff
@@ -34,10 +34,12 @@ class Run < ApplicationRecord
       Task.available_tasks.map(&:to_s)
     } }
     validate :csv_attachment_presence, on: :create
+    validate :valid_task_arguments?, on: :create
```
rafaelfranca, line 37:
```suggestion
    validate :validate_task_arguments, on: :create
```

Validations are actions, not queries

app/models/maintenance_tasks/run.rb
rafaelfranca: Better to define the methods conditionally based in the Rails version with different implementations so we know we can remove the 6.0 implementation when 6.0 is not supported anymore.

test/models/maintenance_tasks/run_test.rb
```diff
@@ -20,6 +20,14 @@ class RunTest < ActiveSupport::TestCase
       refute run.valid?
     end
 
+    test "invalid if associated Task has parameters and they are invalid" do
+      run = Run.new(
+        task_name: "Maintenance::ParamsTask",
+        arguments: { post_ids: "xyz" }
+      )
+      refute run.valid?
```
rafaelfranca, line 28:
```suggestion
      refute_predicate run, valid?
```

test/models/maintenance_tasks/run_test.rb
```diff
@@ -276,6 +284,22 @@ class RunTest < ActiveSupport::TestCase
       end
     end
 
+    test "#task returns Task instance for Run" do
+      run = Run.new(task_name: "Maintenance::UpdatePostsTask")
+      assert run.task.is_a?(Maintenance::UpdatePostsTask)
```
rafaelfranca, line 289:
```suggestion
      assert_kind_of Maintenance::UpdatePostsTask, run.task
```

app/models/maintenance_tasks/run.rb
adrianna-chang-shopify: Opting for `task_arguments_are_valid` to make it less redundant

etiennebarrie:
Since we have parameters, should we still prevent running multiple tasks with different arguments? It probably requires a few changes because we've always assumed "one Task = zero or one Run", but conceptually, two runs with different arguments are similar to two different Tasks.
It came up to me while writing about #422, users could work around the lack of parallelism by having start/end parameters and running the same Task multiple time with different arguments.

I don't think it's reasonable to tackle this here, but just leaving this here.

app/models/maintenance_tasks/run.rb
```diff
@@ -206,6 +208,26 @@ def csv_attachment_presence
       nil
     end
 
+    # Performs validation on the arguments to use for the Task. If the Task is
+    # invalid, the errors are added to the Run.
+    def task_arguments_are_valid
```
rafaelfranca, line 213: This method is saying that the argument are valid, but what it does is check if the arguments are valid. SInce it is a public method that people can call I still think it should be called `validate_task_arguments`. That is the action being done by the method. Or `check_if_ task_arguments_are_valid`

adrianna-chang-shopify: Isn't it redundant to repeat `validate` or to prepend with `check_if_`? I was styling it how [Rails says to write custom validations](https://guides.rubyonrails.org/active_record_validations.html#custom-methods):
```ruby
class Invoice < ApplicationRecord
  validate :active_customer, on: :create

  def active_customer
    errors.add(:customer_id, "is not active") unless customer.active?
  end
end
```

ie. `validate <condition_that_should_be_met_for_record_to_be_valid>`

rafaelfranca: It is, but those methods are usually private and you don't expect users to call `invoice.active_customer`. Our method here is public though, and it is being called in the test, so its usage, is different.

adrianna-chang-shopify: Gotcha, makes sense. Will update 👍

rafaelfranca (approved):

app/controllers/maintenance_tasks/tasks_controller.rb
```diff
@@ -38,6 +39,12 @@ def run
 
     private
 
+    def task_arguments
+      return {} unless params[:task_arguments].present?
+      task_attributes = Task.named(params[:id]).attribute_names
```
etiennebarrie, line 44: Previously a wrong Task name would result in a `ActiveRecord::RecordInvalid` raised from Runner and error caught in the action, now it will raise here and will result in a 500. We don't have tests for this because we didn't want controller tests and there's no way to test this with the UI (with a browser test, you'd need to see the UI and remove the task before clicking on Run). We could ignore that for now because like I mentioned in a previous review, parts of this logic should move to the Runner, with the CLI job needing similar handling, and fix it once we support arguments from the CLI.

app/models/maintenance_tasks/run.rb
```diff
@@ -34,10 +34,12 @@ class Run < ApplicationRecord
       Task.available_tasks.map(&:to_s)
     } }
     validate :csv_attachment_presence, on: :create
+    validate :verify_task_arguments_are_valid, on: :create
```
etiennebarrie, line 37: I think I preferred `validate_task_arguments` because that was only one verb, validate vs. verify thngs are valid, which is two verbs and the noun that we care about (valid). Naming is hard 😄 fine to keep it that way anyway.

app/models/maintenance_tasks/runner.rb
```diff
@@ -37,14 +37,17 @@ def initialize(run)
...
+    #   accessible to the Task.
     #
     # @return [Task] the Task that was run.
     #
     # @raise [EnqueuingError] if an error occurs while enqueuing the Run.
     # @raise [ActiveRecord::RecordInvalid] if validation errors occur while
     #   creating the Run.
-    def run(name:, csv_file: nil)
-      run = Run.active.find_by(task_name: name) || Run.new(task_name: name)
+    def run(name:, csv_file: nil, arguments: {})
```
etiennebarrie, line 48: It's starting to be more and more weird that we use `run` for starting or resuming a task. We previously assumed you resume without a CSV (because otherwise we would override the CSV file), and now we ignore the arguments if you resume, which is a different behavior. Nothing to do for this PR though, just commenting.

app/views/maintenance_tasks/tasks/show.html.erb
```diff
@@ -16,6 +16,16 @@
           <%= form.file_field :csv_file %>
         </div>
       <% end %>
+      <% if @task.parameter_names.any? %>
+        <div class="block">
+          <%= form.fields_for :task_arguments do |ff| %>
+            <% @task.parameter_names.each do |parameter| %>
+              <%= ff.label parameter, "#{parameter}: ", class: "label" %>
+              <%= ff.text_area parameter %>
```
etiennebarrie, line 24: It's barely visible with Firefox. Doesn't it need a class like everything does with Bulma?

db/migrate/20210517131953_add_arguments_to_maintenance_tasks_runs.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+class AddArgumentsToMaintenanceTasksRuns < ActiveRecord::Migration[6.0]
+  def change
+    add_column(:maintenance_tasks_runs, :arguments, :text)
```
etiennebarrie, line 4: We don't do any check on the size. I guess the good thing about JSON is that if the content is too long, it will not parse and won't risk updating the wrong rows etc. like it could with YAML. I'd like to test this and add a validation if necessary.

test/application_system_test_case.rb
```diff
@@ -19,12 +19,10 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
 
   setup do
     travel_to Time.zone.local(2020, 1, 9, 9, 41, 44)
-    Maintenance::UpdatePostsTask.fast_task = false
   end
 
   teardown do
     assert_empty page.driver.browser.manage.logs.get(:browser)
-    Maintenance::UpdatePostsTask.fast_task = true
```
etiennebarrie, line 27: 🚤 💨

test/dummy/db/schema.rb
```diff
@@ -48,14 +48,15 @@
     t.integer "tick_count", default: 0, null: false
     t.integer "tick_total"
     t.string "job_id"
-    t.bigint "cursor"
+    t.integer "cursor"
```
etiennebarrie, line 51: Yeah sqlite can't dump it back into a bigint, but I think we should keep it as a bigint since it's what's in the original migration.

test/system/maintenance_tasks/runs_test.rb
```diff
@@ -29,6 +29,34 @@ class RunsTest < ApplicationSystemTestCase
       assert_no_button "Run"
     end
 
+    test "run a Task that accepts parameters" do
+      visit maintenance_tasks_path
+
+      click_on("Maintenance::ParamsTask")
+      post_id = Post.first.id
+      fill_in("[task_arguments][post_ids]", with: post_id.to_s)
```
etiennebarrie, line 37: Is that the `name` of the field? It's kind of weird. Though it's probably ok since it's not an API and just a browser field. I wonder if we could change the test to use the label name instead of the field name though… 🤔

app/controllers/maintenance_tasks/tasks_controller.rb
adrianna-chang-shopify: Hum, interesting. Let's leave it for now, I can remove it in the CLI PR if we choose to go that route and we feel it's redundant. Alternatively, we could keep the use of the Strong Parameters API here and just rescue `Task::NotFoundError` and return `{}`. I'm suspecting that we may just want to check the keys in `validate_task_arguments` though, since we're already rescuing on `Task::NotFoundError` there. But I'll defer to the CLI PR.

app/models/maintenance_tasks/runner.rb
adrianna-chang-shopify: Yeah, I'll open a ticket to explore this, I think there were other things that came up that were making us think about splitting these up?

test/dummy/db/schema.rb
adrianna-chang-shopify: Yeah my bad, didn't even see this! Thanks for flagging 👍

app/views/maintenance_tasks/tasks/show.html.erb
adrianna-chang-shopify: Yeah, forgot to circle back on this. I'll add Bulma's `textarea` class and see if that works for FF.

db/migrate/20210517131953_add_arguments_to_maintenance_tasks_runs.rb
adrianna-chang-shopify: Good point, we probably want a validation so we can surface in the UI and fail early rather than just not serializing the data. Do we just want to support the maximum size for a `text` column (65,535 bytes => 30000 char length)?

test/system/maintenance_tasks/runs_test.rb
adrianna-chang-shopify: Yep, we can change this to use the label's `for` (the textarea's `id` instead of `name`)

adrianna-chang-shopify:
@etiennebarrie latest round of changes in 😁 Can you let me know:
- How the textarea looks in Firefox for you now with the added Bulma class? (we can also make it larger / smaller, see https://bulma.io/documentation/form/textarea/#sizes)
- Validating the length of arguments - see https://github.com/Shopify/maintenance_tasks/pull/413#discussion_r640954788

db/migrate/20210517131953_add_arguments_to_maintenance_tasks_runs.rb
etiennebarrie: We may be able to pull that from the column itself, in case users have tweaked the migration to support their own use case.

PR #414, opened 2021-05-17
Introduce value objects for complex parameter types

adrianna-chang-shopify:
We can offer users some custom parameter type objects to handle simple input validation + appropriate casting for more complex parameter types (the two we've talked about so far are arrays of strings & arrays of integers).

[`test/dummy/app/tasks/maintenance/params_task.rb`](https://github.com/Shopify/maintenance_tasks/pull/414/files#diff-80f167327e73bc8914ccd55812a0a82770f9497ea32e7081216ea99cbf53d2ef) demonstrates how a Task author would use one of these types in their Task. This way, users don't have to copy-pasta a regex across Tasks, and they also get typecasting for free without having to cast their user input themselves in the Task.

The regexes right now are pretty simple. 
For array of integers: We accept a comma-delimited series of integers. We allow zero or one whitespace at the start or end of the string, before and after commas, and before and after the integers. (We could make it accept any amount of whitespace here).
 For array of strings: We accept a comma-delimited series of strings (alphanumeric characters). I've just applied the same whitespace pattern matching as the integer array, although these will be included in the typecast result (we could discard additional whitespace at the start / end of the strings we parse? Not sure how fancy we want to get here)

Feedback welcome! 🤗  I would like to get an initial version of this ship and iterate as necessary though without getting too invested in the details 😉


app/controllers/maintenance_tasks/tasks_controller.rb
```diff
@@ -33,11 +33,11 @@ def run
       redirect_to(task_path(task))
     rescue ActiveRecord::RecordInvalid => error
       redirect_to(task_path(error.record.task_name), alert: error.message)
-    rescue ActiveRecord::ValueTooLong => error
-      task_name = params.fetch(:id)
-      redirect_to(task_path(task_name), alert: error.message)
     rescue Runner::EnqueuingError => error
       redirect_to(task_path(error.run.task_name), alert: error.message)
+    rescue ActiveRecord::ValueTooLong, Parameters::TypeError => error
```
adrianna-chang-shopify, line 38: Moving errors where we don't have access to the `task_name` any way other than via the original `params` to the bottom here. I think we can clean up some of this later on (ie. just use the id from the params for all of them), but not in this PR 😁

adrianna-chang-shopify:
cc @Shopify/rails @Shopify/core-stewardship

app/models/maintenance_tasks/parameters/integer_array_type.rb
```diff
@@ -0,0 +1,31 @@
...
+        input.split(",").map(&:to_i)
+      end
+
+      private
+
+      def validate_input(input)
+        unless /\A(\s?\d+\s?(,\s?\d+\s?)*\s?)\z/.match?(input)
+          error_message = "#{self.class} expects a " \
+            "comma-delimited string of integers. Input received: #{input}"
+          raise TypeError, error_message
```
etiennebarrie, line 26: Typically in Active Record, types don't raise too much, they're not really meant to do validation as far as I understand:
```ruby
run = MaintenanceTasks::Run.new
run.tick_total = "a"
# => "a"
run.tick_total
# => 0
run.tick_total_before_type_cast
# => "a"
```

app/controllers/maintenance_tasks/tasks_controller.rb
etiennebarrie: There should probably be something similar in Runner to support the CLI as well.

app/models/maintenance_tasks/parameters/type_error.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+module MaintenanceTasks
+  module Parameters
+    # Error to raise when parameter ActiveModel::Value::Type objects fail
+    # validation.
+    class TypeError < StandardError; end
```
etiennebarrie, line 6: This means types implemented by users would need to use this to support error handling. They probably don't want to couple their types with the gem though.

etiennebarrie (commented):
I understand we want to make it real easy to deal with array of ids, but I think we should try to keep true to the roles each part of the framework need to play (types vs validations, etc.).

There is something in ActiveModel:Type::Integer that raises but it's closer to correctness than validation: https://github.com/rails/rails/blob/v6.1.3.2/activemodel/lib/active_model/type/integer.rb#L49
The DB would raise otherwise anyway.

I think I need to think about it more, I'm still not entirely sure what the solution would look like.

adrianna-chang-shopify:
@etiennebarrie I can get on board with your perspective 😄  My inclination was to treat input data with a lot of caution, being pretty strict about what was expected and raising if it didn't conform to those expectations. You're right about Active Record types though - they're super loose in terms of how they coerce data, and they mostly just resort to `nil` if the data can't be coerced properly.

Let's go with just performing the coercion for users, and returning nil if the input can't be parsed properly into the intended value type - it can be up to users to define further validations on their Tasks if they need more safeguards in place.

adrianna-chang-shopify:
For now, I've opted to introduce the value types without any of the validations, and I've added a note in the README:
> If you use one of these types for a parameter in your Task, your input will becoerced appropriately. **You may still want write validations for your input, as type coercion is fairly lax. For example, if you input "xyz" for an`IntegerArrayType` parameter, it will be cast to `[0]`.**

We may want to offer custom validations eventually (as you'd suggested), but I'm not sure what makes the most sense right now and so I'd prefer to defer until we have some feedback from users on what sort of custom validators would bring value.

adrianna-chang-shopify:
We've agreed to put this on pause until we have more info. I've opened https://github.com/Shopify/maintenance_tasks/issues/432 in the meantime to keep track of this feature.

gmcgibbon:
@adrianna-chang-shopify can this be closed?

adrianna-chang-shopify:
Closing, I think we've determined that there isn't enough of a demand for a feature like this, and we weren't quite happy with the API in this PR.

PR #425, opened 2021-05-31
Yield the run in `Runner.run`

### LINKED ISSUE #405: Refactor `MaintenanceTasks::Runner` to return either the `Task` or the `Run`
For Core, we have a Run per pod. The `Runner` currently returns a Task, so the only way to get the run is to perform an additional query using the Task name to find the run. We can avoid this by modifying [`Runner#run` ](https://github.com/Shopify/maintenance_tasks/blob/main/app/models/maintenance_tasks/runner.rb#L46-L53 )to return a value that can be overriden downstream to return something else (ie. we'll hook into a method and modify `Runner.run` to return the run in Core, while still returning a Task otherwise).

etiennebarrie:
When wrapping this gem for Shopify/shopify, we want to get the newly created or resumed `Run` from `Runner.run`.

This yields the current run only if a block is passed to `Runner.run`, without documenting that feature since `Run` is a private API that most apps shouldn't tinker with.

Fixes #405

### DIFF
--- a/app/models/maintenance_tasks/runner.rb
+++ b/app/models/maintenance_tasks/runner.rb
@@ -54,6 +54,7 @@ module MaintenanceTasks
 
       run.enqueued!
       enqueue(run)
+      yield run if block_given?
       Task.named(name)
     end
 
--- a/test/models/maintenance_tasks/runner_test.rb
+++ b/test/models/maintenance_tasks/runner_test.rb
@@ -32,6 +32,20 @@ module MaintenanceTasks
       end
     end
 
+    # Yielding the run is undocumented and not supported in the Runner's API
+    test "#run yields the newly created Run when there is no active Run" do
+      run = Run.create!(task_name: @name, status: :paused)
+
+      @runner.run(name: @name) { |yielded| @run = yielded }
+      assert_equal run, @run
+    end
+
+    # Yielding the run is undocumented and not supported in the Runner's API
+    test "#run yields the existing active Run" do
+      @runner.run(name: @name) { |run| @run = run }
+      assert_equal Run.last, @run
+    end
+
     test "#run raises validation error if no Task for given name" do
       assert_no_difference -> { Run.where(task_name: "Invalid").count } do
         assert_no_enqueued_jobs do


app/models/maintenance_tasks/runner.rb
```diff
@@ -54,11 +54,15 @@ def run(name:, csv_file: nil, arguments: {})
 
       run.enqueued!
       enqueue(run)
-      Task.named(name)
+      returns(run)
     end
 
     private
 
+    def returns(run)
```
gmcgibbon, line 62: If we want this to be overridable / documented, we should probably make this public right?

etiennebarrie: I don't really want to make it public. Just make it easier to return a Run in Shopify/shopify, without having to re-write the full method, or introducing a flag/API.

But I think this is not enough though because overriding `returns` as it is would break the controller and CLI. I have to find another way, maybe I need to add a flag to `Runner.run` after all…

gmcgibbon: It may also be that this is the wrong time to upstream. Especially if

> `Run` is a private API that most apps shouldn't tinker with.

We may be able to make a better decision when we have two instances of needing to return a run, if that exists in our use-case.

adrianna-chang-shopify: Agree that since `Run` is private API it doesn't make sense to change the method to return the `Run` in the general case.

> as it is would break the controller and CLI.

Is this really a problem? The only thing talking to the gem's `Runner` in Core will be our CPT that handles enqueuing - we'll have a top-level runner (and separate CLI / controller entrypoints) so I don't think modifying this return value will cause issues unless I'm missing something?

adrianna-chang-shopify: Ahh nice!! Yep, even better without having to monkey patch 🙌

test/models/maintenance_tasks/runner_test.rb
```diff
@@ -32,6 +32,18 @@ class RunnerTest < ActiveSupport::TestCase
       end
     end
 
+    test "#run yields the Run when there is no active Run (undocumented)" do
+      run = Run.create!(task_name: @name, status: :paused)
+
+      @runner.run(name: @name) { |yielded| @run = yielded }
```
adrianna-chang-shopify, line 38:
```suggestion
      @runner.run(name: @name) { |run| @run = run }
```
Nit

test/models/maintenance_tasks/runner_test.rb
```diff
@@ -32,6 +32,18 @@ class RunnerTest < ActiveSupport::TestCase
       end
     end
 
+    test "#run yields the Run when there is no active Run (undocumented)" do
```
adrianna-chang-shopify, line 35:
```suggestion
    test "#run yields the newly created Run when there is no active Run" do
```

test/models/maintenance_tasks/runner_test.rb
```diff
@@ -32,6 +32,18 @@ class RunnerTest < ActiveSupport::TestCase
...
+      run = Run.create!(task_name: @name, status: :paused)
+
+      @runner.run(name: @name) { |yielded| @run = yielded }
+      assert_equal run, @run
+    end
+
+    test "#run yields the existing active Run (undocumented)" do
```
adrianna-chang-shopify, line 42:
```suggestion
    test "#run yields the existing active Run" do
```

test/models/maintenance_tasks/runner_test.rb
adrianna-chang-shopify, line 35: Can we add a comment above each of these two tests instead of putting `undocumented` in the test name to make things clearer? Maybe:

`# Yielding the run is undocumented and not supported in the Runner's API`

adrianna-chang-shopify (approved):

test/models/maintenance_tasks/runner_test.rb
etiennebarrie: I can't really do that one because Rubocop complains.

adrianna-chang-shopify: Right, forgot we were using `run` already above. Ignore me!

etiennebarrie:
This one shouldn't make it to the release notes. ;)

app/models/maintenance_tasks/runner.rb
gmcgibbon: Yeah, agree yield was a good choice 👍

PR #426, opened 2021-05-31
Added bin prefix to rails command in post-install message  + Use heredoc in post command

andynguyenshopify:
Rails encourages the use of binstubs when invoking rails.

### DIFF
--- a/maintenance_tasks.gemspec
+++ b/maintenance_tasks.gemspec
@@ -20,7 +20,7 @@ Gem::Specification.new do |spec|
   spec.executables = ["maintenance_tasks"]
 
   spec.post_install_message = "Thank you for installing Maintenance Tasks "\
-    "#{spec.version}. To complete, please run:\n\nrails generate "\
+    "#{spec.version}. To complete, please run:\n\nbin/rails generate "\
     "maintenance_tasks:install"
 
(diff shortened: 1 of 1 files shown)


maintenance_tasks.gemspec
```diff
@@ -19,9 +19,11 @@ Gem::Specification.new do |spec|
   spec.bindir = "exe"
   spec.executables = ["maintenance_tasks"]
 
-  spec.post_install_message = "Thank you for installing Maintenance Tasks "\
-    "#{spec.version}. To complete, please run:\n\nrails generate "\
-    "maintenance_tasks:install"
+  spec.post_install_message = <<~MESSAGE.strip
```
rafaelfranca, line 22: Can we remove this post install message? It show every time the gem is installed, even if it is not in the first time and the generator was already installed, so it is more confusing than helpful.

rafaelfranca (approved):

adrianna-chang-shopify (approved):

gmcgibbon (approved):

gmcgibbon:
> There are few uses of rails in the README too if we really care about this.

I think reflecting rails' standard is a good idea. I always use `bin/rails` out of habit, and I know we accept PRs on rails related to doc updates using the binstub, so this project should be no different.

PR #427, opened 2021-06-01
Remove lockfiles for Rails edge testing

etiennebarrie:
Right now we don't really update the version of Rails edge meaning we wouldn't know if something upstream was breaking the gem.

By removing the lockfiles, bundler will get new versions of Rails automatically. Also doing this for 6.0 which allows to remove manual steps when dependabot updates a gem.


rafaelfranca:
Are the lockfiles being used in CI? While the lockfiles should be in the repository, they should be deleted on CI. If you do that you don't need a new workflow. https://bundler.io/blog/2018/01/17/making-gem-development-a-little-better.html

adrianna-chang-shopify:
We test across Rails 6.0 / 6.1 (just `Gemfile.lock`) / edge (which as Étienne pointed out is just the edge when we first committed the lockfile so not that useful) in CI using `BUNDLE_GEMFILE` and specifying a matrix:
https://github.com/Shopify/maintenance_tasks/blob/cb7fc8a99401d8e556ed201b5e986b39712c46c4/.github/workflows/ci.yml#L10-L15


> While the lockfiles should be in the repository, they should be deleted on CI. If you do that you don't need a new workflow

This wouldn't test against the edge of Rails though would it, only the latest stable release? We want to test edge so we can catch incompatibilities in Core ahead of shipping them in the gem.

rafaelfranca:
> This wouldn't test against the edge of Rails though would it, only the latest stable release? We want to test edge so we can catch incompatibilities in Core ahead of shipping them in the gem.

If you remove the Gemfile lock before running the tests, it would test against the Rails version specified in the Gemfiles.

adrianna-chang-shopify (approved):
TIL! Neat 🚀

etiennebarrie:
> Using rails 7.0.0.alpha from https://github.com/rails/rails.git (at main@ee8134d)

https://github.com/Shopify/maintenance_tasks/runs/2721520258#step:4:105

---

So that works, but maybe we should just not have the lockfiles for the alternative Gemfiles, just the main one (for easy setup for new contributors and general reproducibility for archeology/git bisect). That would solve this issue and also remove the need to update all the lockfiles when dependabot updates a gem 🤔

adrianna-chang-shopify:
> That would solve this issue and also remove the need to update all the lockfiles when dependabot updates a gem

Ah, that's brilliant! Yeah let's do that - just keep the main Gemfile.lock, relying on dependabot to help us keep that up to date, and then remove the others and have CI test against the latest versions of deps. Having to keep all our lockfiles up to date manually is really a pain, so this would be great.

adrianna-chang-shopify (approved):
👌

etiennebarrie:
One tiny issue is that without a lockfile, we don't know exactly which version of rails is used, since ruby/setup-ruby does a `bundle lock` then `bundle install`:

>   Using rails 7.0.0.alpha from https://github.com/rails/rails.git

https://github.com/Shopify/maintenance_tasks/runs/2722058837?check_suite_focus=true#step:3:90

Fixing that and shipping.

PR #430, opened 2021-06-02
Return a Duration instead of a Time for the ETA

### LINKED ISSUE #419: Run#time_to_completion without Time.now included in the calculation.
Replace [`Run#estimated_completion_time`](https://github.com/Shopify/maintenance_tasks/blob/main/app/models/maintenance_tasks/run.rb#L158-L165) with `Run#time_to_completion`, which should be an absolute amount of time rather than the calculation based on the current time. This will ensure downstream compatibility with time estimations in Shopify/shopify.

etiennebarrie:
Fixes #419

### DIFF
--- a/test/models/maintenance_tasks/run_test.rb
+++ b/test/models/maintenance_tasks/run_test.rb
@@ -158,16 +158,16 @@ module MaintenanceTasks
       end
     end
 
-    test "#estimated_completion_time returns nil if the run is completed" do
+    test "#time_to_completion returns nil if the run is completed" do
       run = Run.new(
         task_name: "Maintenance::UpdatePostsTask",
         status: :succeeded
       )
 
-      assert_nil run.estimated_completion_time
+      assert_nil run.time_to_completion
     end
 
-    test "#estimated_completion_time returns nil if tick_count is 0" do
+    test "#time_to_completion returns nil if tick_count is 0" do
       run = Run.new(
         task_name: "Maintenance::UpdatePostsTask",
         status: :running,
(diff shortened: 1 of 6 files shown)


test/models/maintenance_tasks/run_test.rb
```diff
@@ -158,51 +158,47 @@ class RunTest < ActiveSupport::TestCase
...
 
-      assert_nil run.estimated_completion_time
+      assert_nil run.time_to_completion
     end
 
-    test "#estimated_completion_time returns estimated completion time based on average time elapsed per tick" do
-      started_at = Time.utc(2020, 1, 9, 9, 41, 44)
-      travel_to started_at + 9.seconds
-
+    test "#time_to_completion returns estimated completion time based on average time elapsed per tick" do
```
adrianna-chang-shopify, line 191:
```suggestion
    test "#time_to_completion returns estimated duration until completion based on average time elapsed per tick" do
```
Nit - just to clarify from the previous wording

test/models/maintenance_tasks/run_test.rb
```diff
@@ -158,51 +158,47 @@ class RunTest < ActiveSupport::TestCase
...
       run = Run.new(
         task_name: "Maintenance::UpdatePostsTask",
-        started_at: started_at,
+        started_at: nil,
```
adrianna-chang-shopify, line 194: Can we remove `started_at` altogether since it's not necessary to compute the time to completion?

adrianna-chang-shopify (approved):

PR #434, opened 2021-06-09
Remove status transition of paused => cancelling

adrianna-chang-shopify:
This transition was introduced in https://github.com/Shopify/maintenance_tasks/pull/166, but I have no idea why 😅 

With our logic in `Run#cancel`:
https://github.com/Shopify/maintenance_tasks/blob/64ad4a55bdc97f0c4b30ec2f5cb0c64bfb55fbfa/app/models/maintenance_tasks/run.rb#L180-L185

The only possible transition for a `paused` run if a user presses "Cancel" is to move it to `cancelled`.


adrianna-chang-shopify:
RuboCop [has got a bug](https://github.com/rubocop/rubocop/issues/9861) so CI is failing

Issue #441, opened 2021-06-29
Offer upgrade command separate from install generator


adrianna-chang-shopify:
Look into providing separate upgrade command (rather than `bin/rails generate maintenance_tasks:install`) that only copies relevant migrations to consumers who are using a db dump file or have rolled up / deleted migrations.

This will likely require patching `ActiveRecord` to use the original datestamps when copying over migrations so that we can query the schema versions appropriately in the schema migration table (right now, new timestamps are generated when migrations are copied to the app).

This won't solve issues for our existing users (since we won't know what timestamps the migrations were originally generated with), and is somewhat niche to large Rails apps in Shopify (as many don't rollup migrations), so it might not be worth the effort to solve. Let's think on it.

adrianna-chang-shopify:
Closing this. The note in the README seems to be sufficient.

PR #442, opened 2021-06-29
Add note on handling rolled-up migrations to the README

adrianna-chang-shopify:
This comes up a fair amount at Shopify, since many large Rails apps roll up their db migrations. It's reasonable to assume that external applications using this gem may be doing a similar thing.

We discussed trying to be smarter about figuring out which migrations an app already ran and only copying over necessary ones, which spawned [this issue](https://github.com/Shopify/maintenance_tasks/issues/441), but it's a pretty involved solution for the size of the problem.

A quick fix that makes the user experience better is laid out in this PR. If an app has rolled up past MT migrations, they are advised to run the migration copying subcommand, remove any unnecessary migrations (by checking the release notes to see which ones have been added since the last upgrade) and then manually running `bin/rails db:migrate`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -548,6 +548,16 @@ $ bin/rails generate maintenance_tasks:install
 
 This ensures that new migrations are installed and run as well.
 
+**What if I've deleted my previous Maintenance Task migrations?**
+
+The install command will attempt to reinstall these old migrations and migrating
+the database will cause problems. Use `bin/rails generate maintenance_tasks:install:migrations`
+to copy the gem's migrations to your `db/migrate` folder. Check the release
+notes to see if any new migrations were added since your last gem upgrade.
+Ensure that these are kept, but remove any migrations that already ran.
+
+Run the migrations using `bin/rails db:migrate`.
+
 ## Contributing
 
 Would you like to report an issue or contribute with code? We accept issues and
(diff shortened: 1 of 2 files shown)


test/dummy/config/application.rb
```diff
@@ -33,5 +33,9 @@ class Application < Rails::Application
     # Application configuration can go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded after loading
     # the framework and any gems in your application.
+
+    if Rails::VERSION::STRING.match?(/^7.0/)
```
etiennebarrie, line 37: I'm not even sure we need the condition, I think `config` being an `OrderedOptions` would allow writing to it always.

README.md
```diff
@@ -548,6 +548,16 @@ $ bin/rails generate maintenance_tasks:install
 
 This ensures that new migrations are installed and run as well.
 
+**What if I've rolled up previous Maintenance Task migrations?**
```
etiennebarrie, line 551: I'm not sure about the wording. At Shopify we don't actually roll them up so much as delete them, and rely on the schema. Maybe just:
```suggestion
**What if I've deleted my previous Maintenance Task migrations?**
```

etiennebarrie (approved):

test/dummy/config/application.rb
adrianna-chang-shopify: [CI complained](https://github.com/Shopify/maintenance_tasks/runs/2945274445?check_suite_focus=true) when setting up the db on previous Rails versions:

Screen Shot 2021-06-30 at 10 33 08 AM

etiennebarrie: Oh ok it's not using ActiveSupport::Configurable. 👍

Issue #448, opened 2021-07-13
Expose a reusable constant for stuck task timeout


andynguyenshopify:
As discussed in https://github.com/Shopify/shopify/pull/302746#pullrequestreview-704222553. We need a way to synchronize changes to the stuck task timer in both core and the gem. At the moment both are using 5 minutes for a stuck timeout but there could be consequences if these changed.

I.e if we changed the timer to one min in Core, we're marking a cross pod run as cancelled, but the run as cancelling in the gem, because we call Run#cancel and our timer is still 5 mins here.

Once this is closed here: We should also open a separate PR in core to use this constant as well.

PR #450, opened 2021-07-14
Add constant for stuck task timeouts

andynguyenshopify:
Closes: #448 
We need a way to synchronize changes to the stuck task timer if an application decides to extend the gem. At the moment if both applications are using 5 minutes for a stuck timeout but there could be consequences if these changed. (I.e in the gem but not the application based off it.

I.e if we changed the timer to one min and another application has the old timeout (5 minutes).

### DIFF
--- a/app/models/maintenance_tasks/run.rb
+++ b/app/models/maintenance_tasks/run.rb
@@ -27,6 +27,7 @@ module MaintenanceTasks
     ]
     COMPLETED_STATUSES = [:succeeded, :errored, :cancelled]
     COMPLETED_RUNS_LIMIT = 10
+    STUCK_TASK_TIMEOUT = 5.minutes
 
     enum status: STATUSES.to_h { |status| [status, status.to_s] }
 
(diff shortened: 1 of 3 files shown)


rafaelfranca:
How do we plan to synchronize changes to a constant? The value of a constant should not be changed, because, if it does, it is not a constant anymore.

andynguyenshopify:
> How do we plan to synchronize changes to a constant? The value of a constant should not be changed, because, if it does, it is not a constant anymore.

I think there is a misunderstanding. The constant right now is 5 minutes, and it won't get changed unless a dev changes it in the source code. If we ever change this in the source code in the future (i.e we change this in the gem). Then any application that uses a hard-coded 5 minutes for stuck tasks will be broken. This will allow the said application to use Run::StuckTaskTimeout to avoid this issue.

rafaelfranca (approved):

adrianna-chang-shopify (approved):

adrianna-chang-shopify:
Just need to rebase to fix CI 😄

PR #451, opened 2021-07-14
Persist `job_id` to Run before job is enqueued

adrianna-chang-shopify:
Compatibility with the blackhole system in Shopify/shopify requires us to have the `job_id` available on the `MaintenanceTasks::Run` as soon as it's enqueued, not once it's about to start performing.

This PR simply changes the deserialization of the `MaintenanceTasks::Run` + the setting of `job_id` on the run to happen in a `before_enqueue` block rather than in the `before_perform` block.

### DIFF
--- a/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
+++ b/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
@@ -8,6 +8,7 @@ module MaintenanceTasks
     include JobIteration::Iteration
 
     included do
+      before_enqueue(:before_enqueue)
       before_perform(:before_perform)
 
       on_start(:on_start)
@@ -87,13 +88,17 @@ module MaintenanceTasks
       raise error
     end
 
+    def before_enqueue
+      @run = arguments.first
+      @run.update!(job_id: job_id)
+    end
+
     def before_perform
       @run = arguments.first
       @task = @run.task
       if @task.respond_to?(:csv_content=)
         @task.csv_content = @run.csv_file.download
       end
-      @run.job_id = job_id
 
       @run.running! unless @run.stopping?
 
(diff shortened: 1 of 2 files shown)


app/jobs/concerns/maintenance_tasks/task_job_concern.rb
```diff
@@ -87,13 +88,17 @@ def task_iteration(input)
...
     end
 
-    def before_perform
+    def before_enqueue
       @run = arguments.first
+      @run.update!(job_id: job_id)
+    end
+
+    def before_perform
+      @run ||= arguments.first
```
rafaelfranca, line 97: when would `@run` be present already?

etiennebarrie (commented):
It's a bit more work now since it does an INSERT then an UPDATE.

Maybe we could instantiate the job and get its job_id before we persist the Run, but that requires changes to Runner. WDYT?

etiennebarrie:
> we wouldn't want to instantiate the job

Instantiating it is fine, what is not is enqueuing it. Instead of `MaintenanceTasks.job.constantize.perform_later(run)`, we should be able to do
```ruby
job = MaintenanceTasks.job.constantize.new(run)
run.job_id = job.job_id
job.enqueue
```

… except we also need to persist before enqueuing, or serializing will fail because of global id I think.

adrianna-chang-shopify:
Ah, yeah instantiating ≠ enqueuing 😄  Indeed though, I think we're stuck because of the need to serialize the GID. Are you okay with shipping this with the extra `UPDATE`?

rafaelfranca (approved):

etiennebarrie:
It's ok to have shipped like that but

> I think we're stuck because of the need to serialize the GID

I meant we had to rework the Runner but I think that was doable. I'll give it a try.

adrianna-chang-shopify:
Yeah, I wanted to ship this to unblock the work on blackholing Task Jobs, but feel free to fix this up 😄  I'd like to get in a PR for moving the yield up anyways prior to cutting a new version for use in Core

PR #453, opened 2021-07-19
Allow backtrace cleaner to be configured

adrianna-chang-shopify:
Rather than defaulting to `Rails.backtrace_cleaner` for cleaning the backtrace before persisting it in `Run#persist_error`, it would be nice if this could be configured. In Shopify/shopify, I need to add some silencers to remove lines related to our job infra / instrumentation from the backtrace, so I'd like to specify a new backtrace cleaner to be used by the gem for cleaning up Task backtraces. 

I think it makes sense to expose this as a config option -- users might have job-related lines that they want to include in traces as part of their default `Rails.backtrace_cleaner`, but want to exclude from traces the gem shows.


lib/maintenance_tasks.rb
```diff
@@ -52,6 +52,16 @@ module MaintenanceTasks
...
 
+  # @!attribute backtrace_cleaner
+  #   @scope class
+  #
+  #   The Active Support backtrace cleaner that will be used to clean the
+  #   backtrace of a Task that errors.
+  #
+  #   @return [ActiveSupport::BacktraceCleaner, nil] the backtrace cleaner to
+  #     use when cleaning a Run's backtrace.
+  mattr_accessor :backtrace_cleaner
```
adrianna-chang-shopify, line 63: Tried to use `default: Rails.backtrace_cleaner` here to avoid having to check for it in `#persist_error`, but unfortunately `Rails.root` is not set at this point, so when the `Rails.backtrace_cleaner` gets set (and memoized) here, it's `@root` is wrong (nil instead of the dummy app) and this throws off the traces.

gmcgibbon (commented):
You'll want to set the value in rails config and then add defaults in an `initializer` block. See https://github.com/Shopify/app_profiler/blob/78cb8b579a2bbf3034c6941f9adb865f5ef3d283/lib/app_profiler/railtie.rb#L10 (for example).

adrianna-chang-shopify:
Thanks @gmcgibbon, I was hoping there was a better way to do that! ❤️

lib/maintenance_tasks/engine.rb
```diff
@@ -11,6 +11,10 @@ class Engine < ::Rails::Engine
       eager_load! unless Rails.autoloaders.zeitwerk_enabled?
     end
 
+    initializer "maintenance_tasks.configs" do
+      MaintenanceTasks.backtrace_cleaner = Rails.backtrace_cleaner
```
gmcgibbon, line 15:
```suggestion
      MaintenanceTasks.backtrace_cleaner ||= Rails.backtrace_cleaner
```

You could move configuration to a rails config option (eg. `Rails.application.config.maintenance_tasks.*`) in the engine, but this should work fine. Given there's already config options that don't follow this pattern, I think it is fine as-is.

gmcgibbon (approved):

PR #457, opened 2021-07-26
Bump version to 1.5.0

adrianna-chang-shopify:
I'm proposing we release a new version now so we can start tackling some of the changes in Shopify/shopify 😄 

We can cut a `v.1.5.1` later to address fixing `job_id` to be set on insert instead of update, as well as to allow collections that support objects responding to `to_enum` to finish up CSV support.

Drafted release notes: https://github.com/Shopify/maintenance_tasks/releases/edit/untagged-a14428f6d5f7d26e76bc

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    maintenance_tasks (1.4.0)
+    maintenance_tasks (1.5.0)
       actionpack (>= 6.0)
       activejob (>= 6.0)
       activerecord (>= 6.0)
--- a/README.md
+++ b/README.md
@@ -591,6 +591,7 @@ are merged.
 Once a release is ready, follow these steps:
 
 * Update `spec.version` in `maintenance_tasks.gemspec`.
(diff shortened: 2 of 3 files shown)


andynguyenshopify (approved):
Looks good! I think the update notes are missing https://github.com/Shopify/maintenance_tasks/issues/448 (I added it to this link https://github.com/Shopify/maintenance_tasks/releases/edit/untagged-e70d7910e5096a546840).

adrianna-chang-shopify:
Ohai @sambostock ! My understanding of pre-releases is that they're generally meant to be used if a release is unstable / might pose issues in production. In the case of 
> We can cut a v.1.5.1 later to address fixing job_id to be set on insert instead of update, as well as to allow collections that support objects responding to to_enum to finish up CSV support.

The change to `job_id` isn't really a bug - it's an optimization that will save us a write - and the change to supporting collections (also heyo I think we might finally ship custom enumerators 🙈 ) is a new feature entirely (required only for supporting CSV in Shopify/shopify, with no impact on the usability of CSV in the gem). IMO I don't think this warrants a pre-release.

PR #458, opened 2021-07-26
Support to_enum objects as collections

gmcgibbon:
Support collection objects that respond to `to_enum`.

### DIFF
--- a/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
+++ b/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
@@ -57,11 +57,15 @@ module MaintenanceTasks
       when CSV
         JobIteration::CsvEnumerator.new(collection).rows(cursor: cursor)
       else
-        raise ArgumentError, <<~MSG.squish
-          #{@task.class.name}#collection must be either an
-          Active Record Relation, ActiveRecord::Batches::BatchEnumerator,
-          Array, or CSV.
-        MSG
+        if collection.respond_to?(:to_enumerable)
+          collection.to_enumerable(cursor: cursor)
+        else
+          raise ArgumentError, <<~MSG.squish
+            #{@task.class.name}#collection must be either an
+            Active Record Relation, ActiveRecord::Batches::BatchEnumerator,
+            Array, CSV, or object that responds to #to_enumerable.
+          MSG
+        end
       end
 
       @task.throttle_conditions.reduce(collection_enum) do |enum, condition|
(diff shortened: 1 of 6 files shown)


app/jobs/concerns/maintenance_tasks/task_job_concern.rb
```diff
@@ -57,11 +57,18 @@ def build_enumerator(_run, cursor:)
       when CSV
         JobIteration::CsvEnumerator.new(collection).rows(cursor: cursor)
       else
-        raise ArgumentError, <<~MSG.squish
-          #{@task.class.name}#collection must be either an
-          Active Record Relation, ActiveRecord::Batches::BatchEnumerator,
-          Array, or CSV.
-        MSG
+        if collection.respond_to?(:to_enum)
```
gmcgibbon, line 60: This conflicts with `to_enum` in the standard library. Do we want to make a different method to avoid needing to rescue from type errors?

app/jobs/concerns/maintenance_tasks/task_job_concern.rb
```diff
@@ -57,11 +57,15 @@ def build_enumerator(_run, cursor:)
       when CSV
         JobIteration::CsvEnumerator.new(collection).rows(cursor: cursor)
       else
-        raise ArgumentError, <<~MSG.squish
-          #{@task.class.name}#collection must be either an
-          Active Record Relation, ActiveRecord::Batches::BatchEnumerator,
-          Array, or CSV.
-        MSG
+        if collection.respond_to?(:to_enumerable)
+          collection.to_enumerable(cursor: cursor)
```
gmcgibbon, line 61: I think I have to warp this in something from job-iteration. What am I missing?

adrianna-chang-shopify:
Étienne had mentioned he wasn't on board with the API [here](https://github.com/Shopify/shopify/pull/304347#issuecomment-887445883) - truth be told there's kind of no winning with introducing the ability to build a collection that has knowledge about enumerators / cursors into the gem. 😅  Right now the thing the gem does well is keeping things simple with its `#collection` API.

For V1, in the interest of shipping CSV as a feature and not getting blocked, it's probably best to keep the changes to support custom enums inside Core only and override `build_enumerator` to handle a `BackgroundQueue::GCSCsvEnumerator` collection. I'm concerned we'll be going around in circles again if we try to nail an API for the gem.

Sorry for getting things sidetracked here Gannon!

gmcgibbon:
Ah. My mistake. Thanks for clarifying 🙇

PR #460, opened 2021-07-27
Style Task arguments section better

adrianna-chang-shopify:
Some slight improvements to the Task show page for Tasks with arguments / CSVs:
### Added horizontal rules to split Task arguments / CSV download section for the latest run from the "Run" section of the page:

**Before**
Screen Shot 2021-07-27 at 4 13 20 PM
**After**
Screen Shot 2021-07-27 at 4 12 58 PM

### Wrapped argument form labels in `<var>` tags:

**Before**
Screen Shot 2021-07-27 at 4 16 37 PM

**After**
Screen Shot 2021-07-27 at 4 21 00 PM

### Changed args to be displayed in a table instead of a single `<pre>` tag. If the arg is a single line, we wrap it in `<code>`, if the argument spans multiple lines we wrap it in `<pre>`:

**Before**
Screen Shot 2021-07-27 at 4 18 16 PM

**After**
Screen Shot 2021-07-27 at 4 19 09 PM
----------------------
Screen Shot 2021-07-27 at 4 19 44 PM
----------------------
Screen Shot 2021-07-27 at 4 20 34 PM
(Long content is scrollable)

### DIFF
--- a/app/views/maintenance_tasks/runs/_info.html.erb
+++ b/app/views/maintenance_tasks/runs/_info.html.erb
@@ -13,6 +13,7 @@
   <div class="block">
     <%= link_to("Download CSV", csv_file_download_path(run)) %>
   </div>
+  <hr/>
 <% end %>
 <% if run.arguments.present? %>
   <div class="block">
@@ -23,4 +24,5 @@
       <% end %>
     </pre>
   </div>
+  <hr/>
 <% end %>
--- a/app/views/maintenance_tasks/runs/_info.html.erb
+++ b/app/views/maintenance_tasks/runs/_info.html.erb
@@ -24,6 +24,7 @@
           <tr>
             <td class="is-family-monospace"><%= key %></td>
             <td>
+              <% next if value.nil? || value.empty? %>
               <% if value.include?("\n") %>
                 <pre><%= value %><pre>
               <% else %>
--- a/app/views/maintenance_tasks/runs/_info.html.erb
+++ b/app/views/maintenance_tasks/runs/_info.html.erb
@@ -22,7 +22,7 @@
       <tbody>
         <% run.arguments.each do |key, value| %>
           <tr>
-            <td><var><%= key %></var></td>
+            <td class="is-family-monospace"><%= key %></td>
             <td>
               <% if value.include?("\n") %>
                 <pre><%= value %><pre>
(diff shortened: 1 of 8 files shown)


gmcgibbon (approved):
TIL `<var>` tags exist 🤔

app/views/maintenance_tasks/runs/_info.html.erb
```diff
@@ -13,14 +13,27 @@
...
-        <%= "#{key}: #{value}" %>
-      <% end %>
-    </pre>
+<div class="table-container">
+  <h6 class="title is-6">Arguments:</h6>
+    <table class="table">
+      <tbody>
+        <% run.arguments.each do |key, value| %>
+          <tr>
+            <td><var><%= key %></var></td>
```
etiennebarrie, line 25: Not a fan of the `<var>`. What I really want is monospace, so that argument names look like code, but `<code>` is a bit too intense for the label and in the table. I pushed a commit to use a class to add styling instead.

app/views/maintenance_tasks/runs/_info.html.erb
```diff
@@ -13,14 +13,27 @@
...
+            <td>
+              <% if value.include?("\n") %>
+                <pre><%= value %><pre>
+              <% else %>
+                <code><%= value %><code>
```
etiennebarrie, line 30: I didn't like that empty values were just a grey blob, so I took care of that, also handling the nil case which should not happen but we never know 🤷

etiennebarrie (commented):
Added some commits to use a monospaced font and remove empty values, let me know what you think.

**Before**

image

---

**After** 
image

app/views/maintenance_tasks/runs/_info.html.erb
adrianna-chang-shopify: Nice! Yeah sorry, forgot to mention that I went with `<var>` because `<tt> is actually deprecated. The styling looks nice this way 👍

adrianna-chang-shopify:
@etiennebarrie made one last small change to avoid adding the horizontal rules in the "previous run" cards and the "task" cards on the index page. I ended up inlining some of the code into `show.html.erb` rather than trying to continue to make `runs/_info.html.erb` work for both the latest run on the show page, as well as the cards on the index page / previous runs on the show page. Let me know if you think this is good to 🚢 !

etiennebarrie (approved):

PR #461, opened 2021-07-28
Remove n+1 for Active Storage attachments

etiennebarrie:
On the index page, for each available task we run one query to get the ActiveStorage::Attachment of the last run of that Task, then if there's one, another query to get the ActiveStorage::Blob. So more like:
* 1 query for Run records
* n queries for ActiveStorage::Attachment records
* m queries for ActiveStorage::Blob records (where m < n)

This reduces to 3 queries, one for each table.


adrianna-chang-shopify (approved):

adrianna-chang-shopify:
Going to 🚢  this so we can include it in the 1.5.0 release.

PR #465, opened 2021-08-03
Change CSV collection implementation to use a strategy pattern.

rafaelfranca:
We want to add more ways to build collection, so using a strategy
pattern makes sense since we don't need to be including different
modules with different API and we can just swap the strategy at
runtime.

### DIFF
--- a/app/models/maintenance_tasks/csv_collection_builder.rb
+++ b/app/models/maintenance_tasks/csv_collection_builder.rb
@@ -3,8 +3,7 @@
 require "csv"
 
 module MaintenanceTasks
-  # Module that is included into Task classes by Task.csv_collection for
-  # processing CSV files.
+  # Strategy for building a Task that processes CSV files.
   #
   # @api private
   class CsvCollectionBuilder
@@ -25,6 +24,7 @@ module MaintenanceTasks
       task.csv_content.count("\n") - 1
     end
 
+    # Return that the Task processes CSV content.
     def has_csv_content?
       true
     end
--- a/app/models/maintenance_tasks/csv_collection_builder.rb
+++ b/app/models/maintenance_tasks/csv_collection_builder.rb
@@ -7,18 +7,13 @@ module MaintenanceTasks
   # processing CSV files.
   #
   # @api private
-  module CsvCollection
-    # The contents of a CSV file to be processed by a Task.
-    #
-    # @return [String] the content of the CSV file to process.
-    attr_accessor :csv_content
-
+  class CsvCollectionBuilder
     # Defines the collection to be iterated over, based on the provided CSV.
     #
     # @return [CSV] the CSV object constructed from the specified CSV content,
     #   with headers.
-    def collection
-      CSV.new(csv_content, headers: true)
+    def collection(task)
+      CSV.new(task.csv_content, headers: true)
     end
 
     # The number of rows to be processed. Excludes the header row from the count
@@ -26,8 +21,12 @@ module MaintenanceTasks
     # an approximation based on the number of new lines.
     #
     # @return [Integer] the approximate number of rows to process.
-    def count
-      csv_content.count("\n") - 1
+    def count(task)
+      task.csv_content.count("\n") - 1
+    end
+
+    def has_csv_content?
+      true
     end
   end
 end
--- a/app/tasks/maintenance_tasks/task.rb
+++ b/app/tasks/maintenance_tasks/task.rb
@@ -7,6 +7,20 @@ module MaintenanceTasks
     include ActiveModel::AttributeAssignment
     include ActiveModel::Validations
 
+    class NullCollectionBuilder
+      def collection(task)
+        raise NoMethodError, "#{task.class.name} must implement `collection`."
+      end
+
+      def count(task)
+        :no_count
+      end
+
+      def has_csv_content?
+        false
+      end
+    end
+
     class NotFoundError < NameError; end
 
     # The throttle conditions for a given Task. This is provided as an array of
@@ -15,6 +29,8 @@ module MaintenanceTasks
     #
     # @api private
     class_attribute :throttle_conditions, default: []
+
+    class_attribute :collection_builder_strategy, default: NullCollectionBuilder.new
 
     class << self
       # Finds a Task with the given name.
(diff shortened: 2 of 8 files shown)


app/tasks/maintenance_tasks/task.rb
```diff
@@ -103,13 +123,35 @@ def load_constants
...
+    def csv_content
+      raise NoMethodError unless has_csv_content?
+
+      @csv_content
+    end
+
+    # Set the contents of a CSV file to be processed by a Task.
+    #
+    # @param csv_content [String] the content of the CSV file to process.
+    def csv_content=(csv_content)
```
rafaelfranca, line 138: This can be made generic using method missing to allow the strategy to define which attributes can be set. Of course that would be meta programming heavy, so harder to understand since we need to set instance variables that don't exist using `instance_variable_set`.

app/tasks/maintenance_tasks/task.rb
```diff
@@ -103,13 +141,36 @@ def load_constants
...
+      raise NoMethodError unless has_csv_content?
+
+      @csv_content = csv_content
+    end
+
+    # Returns whether the Task handles CSV.
+    #
+    # @return [Boolean] whether the Task handles CSV.
+    def has_csv_content?
+      self.class.has_csv_content?
```
gmcgibbon, line 166: You can `delegate :has_csv_content?, to: :class` here.

app/models/maintenance_tasks/csv_collection_builder.rb
```diff
@@ -3,31 +3,30 @@
...
+  # Strategy for building a Task that processes CSV files.
   #
   # @api private
-  module CsvCollection
-    # The contents of a CSV file to be processed by a Task.
-    #
-    # @return [String] the content of the CSV file to process.
-    attr_accessor :csv_content
-
+  class CsvCollectionBuilder
```
gmcgibbon, line 9: Are builders supposed to be private API? There's no tests for them either.

app/tasks/maintenance_tasks/task.rb
```diff
@@ -7,6 +7,34 @@ class Task
...
+      # expected at the start of the run. Return +nil+ if the amount is
+      # undefined, or counting would be prohibitive for your database.
+      #
+      # @return [Integer, nil]
+      def count(task)
+        :no_count
+      end
+
+      # Return that the Task does not process CSV content.
+      def has_csv_content?
```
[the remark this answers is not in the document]
rafaelfranca: True. Same with the method missing for the attributes. I didn't do it, but if we need to we can.

app/tasks/maintenance_tasks/task.rb
rafaelfranca: Yes, but I particularly think `delegate` is overused and prefer the clear method. It is easier to refactoring and to search in the codebase.

rafaelfranca:
To move forwards, let's try to implement our other strategy using this PR as basis to see which API we need for that other use case.

app/models/maintenance_tasks/csv_collection_builder.rb
```diff
@@ -3,31 +3,30 @@
...
     #
     # @return [Integer] the approximate number of rows to process.
-    def count
-      csv_content.count("\n") - 1
+    def count(task)
+      task.csv_content.count("\n") - 1
+    end
+
+    # Return that the Task processes CSV content.
+    def has_csv_content?
```
etiennebarrie, line 28: I think we should have another name for this since we could have CSV content in another way (like we will do in Shopify/shopify for example). This hook method would make more sense as `needs_csv_upload?` or something. But I see why you called it that way, since we have a `csv_content` attribute. Maybe it's just the comment that needs to be changed: "Returns that the Task processes the CSV content passed to the Run"

---

I think looking back, what the gem provides and what the conditionals we've been using (`task_class < CsvCollection` and `task.respond_to?(:csv_content=)` mean in a more generic way is:

* a `input[type=file]` in the view
* an attribute on the Task instance that holds the content† of the file that was uploaded for this run

That way we would not even be prescribing CSV, it could be JSON or whatever, as long as the task has a collection strategy providing items to be processed. It would be more involved to fix this though

† ideally it would be a stream, an IO of some sort, that way we wouldn't have to download the whole file for a single job execution, since with a big file there would be multiple job executions on multiple machines, downloading the whole file each time would be a bit wasteful, but we don't have proper streaming in Active Storage (there is [Service#download_chunk,](https://api.rubyonrails.org/classes/ActiveStorage/Service.html#method-i-download_chunk) but it doesn't give an IO, [Blob#open](https://api.rubyonrails.org/classes/ActiveStorage/Blob.html#method-i-open) does but it downloads everything.

app/tasks/maintenance_tasks/task.rb
```diff
@@ -16,6 +16,9 @@ class NotFoundError < NameError; end
     # @api private
     class_attribute :throttle_conditions, default: []
 
+    class_attribute :collection_builder_strategy,
```
etiennebarrie, line 19: I'm not sure of the name, since it also provides the `count` method.

app/tasks/maintenance_tasks/task.rb
```diff
@@ -16,6 +16,9 @@ class NotFoundError < NameError; end
...
+      default: NullCollectionBuilder.new
```
etiennebarrie, line 20: I think we may want a different instance per Task instance instead of one per class. That way we could handle memoization of `collection` when we want to provide an implicit `count`.

Also maybe the strategy should take the task instance when it's being instantiated, instead of for each method?

etiennebarrie (commented):
I think this approach makes sense. Besides the changes we need in Shopify/shopify, we could use the approach to handle the implicit `count` change we wanted to make but that was causing compatibility issues. Instead of changing `count` to:
```ruby
def count
  if MaintenanceTasks.implicit_count?
    collection.count
  else
    :no_count
  end
end
```

… we could use a different strategy based on the configuration.

rafaelfranca:
> But what are we trying to solve here exactly? The strategy pattern would help if we couldn't change the superclass, but had different behaviours based on data. But here we know at coding time which behaviour we need, so we could swap the class instead. And I don't think we need changes here to be able to provide other subclasses in Shopify/shopify?

We are trying to avoid having to include module at run time (it is technically load time, but it still a run time) to define behavior and define an API for extension. All the problems you can solve with inheritance you can solve with the state or strategy patterns. The difference is that one use inheritance, the other composition. With composition it is easier to swap behavior without resorting to meta-programming.

As the API for extension, with the inheritance approach, we needed to change the `Task` class and the `Job` class to add new behavior. With this, we only need to change the a value in the `Task` class, with no need to modify any of its methods.

It doesn't make easier to implement what #307 or #326 wants to do, because the task still doesn't know about how to build the enumerator, which is the main goal of those PRs.

gmcgibbon:
I don't think the design is 100% decided yet, but merging to unblock some other work. We won't cut a release until the implementation is more concrete. Please feel free to fix forward!

PR #466, opened 2021-08-05
Change the name of the CSV file before uploading

etiennebarrie:
In #318 we wanted to change the name of CSV files, but the added complexity wasn't worth it because we were adding a controller to provide the file name.

This changes the name before uploading, that way files also have useful names in the Active Storage service, not only when downloading them with the UI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,8 +113,7 @@ My Title,Hello World!
 ```
 
 The files uploaded to your Active Storage service provider will be renamed first
-to include a timestamp (ISO8601 without dashes or colons) and the Task name
-(with colons changed into dashes).
+to include an ISO8601 timestamp and the Task name in snake case format.
 
 ### Processing Batch Collections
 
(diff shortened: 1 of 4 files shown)


gmcgibbon (approved):
To be clear, the filename would only be a part of the metadata of the file since Active Storage stores files by key. Regardless, I think this is a good change.

adrianna-chang-shopify:
Agreed, not sure the value of having them named within Active Storage, but even just for downloading, it's nice to have a more comprehensive name.

Slightly nitpicking of me -- I think I'd prefer the format to use underscores instead. Ie. uploading to `Maintenance::CsvTask` would produce a filename in the form `20210805T120542Z_maintenance_import_posts_task.csv`. Do you have an opinion here @gmcgibbon ? 😄

adrianna-chang-shopify (approved):
Going to ship this, we can revisit the naming convention if anyone objects at a later point 😄

Issue #462, opened 2021-07-29
CSV task fails if uuid is used for active_storage_attachments.record_id


shouichi:
Background

We use uuid for primary keys, thus `active_storage_attachments.record_id` is uuid. On the other hand, `maintenance_tasks_runs.id` is `bigint`. Therefore, when a CSV task tries to upload a CSV as an `ActiveStorage::Attachment`, the attachment fails to reference the CSV task's run record (`MaintenanceTasks::Run`) because of the type inconsistency.

Possible solutions

- Document that the generated migration file must be manually edited when uuid is used.
- Respect `g.orm :active_record, primary_key_type: :uuid` when generating the migration file.

Thanks in advance.

shouichi:
Unfortunately simply changing `maintenance_tasks_runs.id` to uuid did not work because of the following code.

https://github.com/Shopify/maintenance_tasks/blob/e64cdf0c265f67dd3a487c5fb0db8dc9df5464d4/app/models/maintenance_tasks/task_data.rb#L44

shouichi:
The project only uses sqlite for testing. It seems that the first step is using both sqlite and postgres for testing. Is this the right direction @etiennebarrie ?

etiennebarrie:
Yes, if it's necessary. I wonder if we could just test UUIDs with sqlite?

shouichi:
Mmm, maybe we could use blob as a primary key and store uuid. I'll take a look.

shouichi:
Changed primary keys to uuid, ran tests, and found that the current code depends on monotonically increasing ID more than I thought. For example, `Run.last` or `Run.where("id < ?", @cursor)`. It would be better to change incrementally. The first step should be to stop depending on monotonically increasing IDs. Does this make sense?

etiennebarrie:
I'm not sure how which approach we should take:
* make all code dependencies on monotonically-increasing IDs abstract so we can change the behavior when using UUIDs, but keep the behavior as it is f

shouichi:
Thanks for the advice.

> If we can avoid a migration, then the first solution is great: we don't need two code paths for UUIDs. But if we need more work for existing users, I think the second solution would be preferable.

Honestly, I'm not sure if we can avoid migration or not.

> In these situations, I tend to do a spike so that I can see all the moving parts, even if I end up shipping change incrementally. I would favor having a branch where we can see everything we will need, even things aren't perfect or tests fail.

[...]

PR #468, opened 2021-08-11
Run tests against sqlite and postgres

shouichi:
This is the first step to tackle https://github.com/Shopify/maintenance_tasks/issues/462.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,13 +5,15 @@ on: [push, pull_request]
 jobs:
   build:
     runs-on: ubuntu-latest
-    name: Ruby ${{ matrix.ruby }} / ${{ matrix.gemfile }}
+    name: Ruby ${{ matrix.ruby }} / ${{ matrix.gemfile }} / ${{ matrix.database }}
     strategy:
       matrix:
         gemfile: [Gemfile, gemfiles/rails_6_0.gemfile, gemfiles/rails_edge.gemfile]
         ruby: [2.7, 3.0]
+        database: [sqlite, postgres]
     env:
       BUNDLE_GEMFILE: ${{ matrix.gemfile }}
+      DB: ${{ matrix.database }}
     steps:
     - name: Check out code
       uses: actions/checkout@v2.3.4
@@ -23,6 +25,12 @@ jobs:
     - name: Rails version
       if: ${{ matrix.gemfile == 'gemfiles/rails_edge.gemfile' }}
       run: bundle info rails | head -1
+    - name: Start Postgres
+      if: ${{ matrix.database == 'postgres' }}
+      run: |
+        sudo sed -i s/md5/trust/g /etc/postgresql/*/main/pg_hba.conf
+        sudo systemctl start postgresql.service
+        echo "DATABASE_URL=postgresql://postgres:@localhost:5432/maintenance_tasks_test" >> $GITHUB_ENV
     - name: Set up database
       run: RAILS_ENV=test bundle exec rails db:setup
     - name: Ruby Tests
--- a/test/lib/generators/maintenance_tasks/install_generator_test.rb
+++ b/test/lib/generators/maintenance_tasks/install_generator_test.rb
@@ -12,10 +12,12 @@ module MaintenanceTasks
     setup do
       skip "This test is too slow" if ENV["SKIP_SLOW"].present?
       setup_sample_app
+      setup_database_url
     end
 
     teardown do
       FileUtils.rm_rf(SAMPLE_APP_PATH)
+      teardown_database_url
     end
 
     test "generator mounts engine and runs migrations" do
@@ -44,6 +46,21 @@ module MaintenanceTasks
 
       Dir.chdir(SAMPLE_APP_PATH) do
         FileUtils.rm_r("db")
+      end
+    end
+
+    def setup_database_url
+      if ENV["DATABASE_URL"]
+        @previous_database_url = ENV["DATABASE_URL"]
+        ENV["DATABASE_URL"] += "_#{Process.pid}"
+        %x(bin/rails db:create)
+      end
+    end
+
+    def teardown_database_url
+      if defined?(@previous_database_url)
+        %x(bin/rails db:drop)
+        ENV["DATABASE_URL"] = @previous_database_url
       end
     end
   end
--- a/Gemfile
+++ b/Gemfile
@@ -6,6 +6,7 @@ gemspec
 
 gem "capybara"
 gem "mocha"
+gem "pg"
 gem "pry-byebug"
 gem "puma"
 gem "rails", defined?(@rails_gem_requirement) && @rails_gem_requirement
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -116,6 +116,7 @@ GEM
     parallel (1.20.1)
     parser (3.0.2.0)
       ast (~> 2.4.1)
+    pg (1.2.3)
     pry (0.13.1)
       coderay (~> 1.1)
       method_source (~> 1.0)
@@ -210,6 +211,7 @@ DEPENDENCIES
   capybara
   maintenance_tasks!
   mocha
+  pg
   pry-byebug
   puma
   rails
--- a/test/models/maintenance_tasks/run_test.rb
+++ b/test/models/maintenance_tasks/run_test.rb
@@ -321,7 +321,9 @@ module MaintenanceTasks
     def count_uncached_queries(&block)
       count = 0
 
-      query_cb = ->(*, payload) { count += 1 unless payload[:cached] }
+      query_cb = ->(*, payload) do
+        count += 1 if !payload[:cached] && payload[:sql] != "SHOW search_path"
+      end
       ActiveSupport::Notifications.subscribed(query_cb,
         "sql.active_record",
         &block)


test/lib/generators/maintenance_tasks/install_generator_test.rb
```diff
@@ -9,13 +9,20 @@ class InstallGeneratorTest < Rails::Generators::TestCase
     destination SAMPLE_APP_PATH
     setup :prepare_destination
 
+    original_adapter = ENV["DB"]
+
     setup do
+      # Force the sample app to use sqlite. This is because both the dummy app
+      # and the sample use the same postgres database, ended up migration
+      # failure.
+      ENV["DB"] = "sqlite"
```
shouichi, line 18: This is the only tricky part that I failed to find a cleaner solution.

shouichi (commented):
This PR essentially switches database adapters on database.yml and others are just adjustments except for the part I made a comment.

adrianna-chang-shopify:
Hey @Shopify/rails, could I get thoughts / feedback on doing the above ^ to handle testing the install generator properly with PostgreSQL?

test/dummy/config/database.yml
```diff
@@ -1,9 +1,6 @@
-# SQLite. Versions 3.8.0 and up are supported.
-#   gem install sqlite3
-#
-#   Ensure the SQLite 3 gem is defined in your Gemfile
-#   gem 'sqlite3'
-#
+<% adapter = ENV.fetch("DB") { "sqlite" } %>
+<% case adapter
```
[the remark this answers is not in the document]
shouichi: Thanks for the suggestion. The `adapter` variable is used for raising an error when it's unknown, i.e.,

```
<% else %>
	<% raise "unknown adapter: #{adapter"} %>
<% end %>
```

I believe it's a good practice to raise an unknown adapter. Do you still want me to change?

test/dummy/config/sample_app_postgres_database.yml
```diff
@@ -0,0 +1,14 @@
+# Database settings for postgres must be in sync with
```
shouichi, line 1: It turns out that rails 6.0 does not allow us to put yaml files in the fixtures directory https://github.com/rails/rails/pull/36884. It's a PR that @gmcgibbon sent, what a coincidence!

shouichi (commented):
Hi, guys. Thanks for the help! I applied most of the suggestions and it __mostly__ works.

The last thing that needs to be fixed is `MaintenanceTasks::RunTest#test_#reload_status_does_not_use_query_cache` which probabilistically fails when postgres.

shouichi:
Found the cause of `MaintenanceTasks::RunTest#test_#reload_status_does_not_use_query_cache` random failure.

When the test is executed first, active record first executes `SHOW search_path` to find out which postgres schemas to search. The solution I came up with is to run a query (e.g., `Run.first`) in setup and it worked. Is there any better solution?

test/dummy/config/database.yml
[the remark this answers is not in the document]
adrianna-chang-shopify: You can still store it in a variable (`<% case (adapter = ENV.fetch("DB")) %>`), but I agree with Gannon that it's more straightforward to set the default in the test helper instead.

adrianna-chang-shopify:
@shouichi you can filter that query out from the `count_uncached_queries` method:
```ruby
    def count_uncached_queries(&block)
      count = 0

      query_cb = ->(*, payload) do
        unless payload[:cached] || payload[:sql].match?("SHOW search_path")
          count += 1
        end
      end
      ActiveSupport::Notifications.subscribed(query_cb,
        "sql.active_record",
        &block)

      count
    end
```

.github/workflows/ci.yml
```diff
@@ -5,13 +5,28 @@ on: [push, pull_request]
 jobs:
   build:
     runs-on: ubuntu-latest
-    name: Ruby ${{ matrix.ruby }} / ${{ matrix.gemfile }}
+    name: Ruby ${{ matrix.ruby }} / ${{ matrix.gemfile }} / ${{ matrix.database }}
     strategy:
+      fail-fast: false
```
adrianna-chang-shopify, line 10: Why this change?

shouichi:
Fixed all the problems, please take a look. Feel free to ping me if I need to rebase or squash.

test/dummy/config/database.yml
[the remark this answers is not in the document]
adrianna-chang-shopify: Let's still set this default in the `test_helper` 😄

test/dummy/config/database.yml
```diff
@@ -13,13 +10,36 @@ development:
...
 test:
   <<: *default
   database: db/test.sqlite3
 
 production:
   <<: *default
   database: db/production.sqlite3
+<% when "postgres" %>
+# Database settings for postgres must be in sync with
+# test/dummy/config/sample_app_postgres_database.yml.
```
adrianna-chang-shopify, line 22: Let's ✂️  this comment

test/dummy/config/sample_app_postgres_database.yml
```diff
@@ -0,0 +1,14 @@
...
+# test/dummy/config/database.yml.
```
adrianna-chang-shopify, line 2: Let's ✂️  this comment as well

adrianna-chang-shopify (approved):
Great work, thanks @shouichi ! A couple small comments. Let's also squash these commits before merging.

.github/workflows/ci.yml
```diff
@@ -5,13 +5,27 @@ on: [push, pull_request]
...
     strategy:
       matrix:
         gemfile: [Gemfile, gemfiles/rails_6_0.gemfile, gemfiles/rails_edge.gemfile]
         ruby: [2.7, 3.0]
+        database: [sqlite, postgres]
     env:
       BUNDLE_GEMFILE: ${{ matrix.gemfile }}
+      DB: ${{ matrix.database }}
+    services:
+      postgres:
```
etiennebarrie, line 18: Is there a way to have a service run conditionally? Better yet, the environment that runs CI already has postgres, so we shouldn't need an external service like this:
https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#databases

etiennebarrie (commented):
> could I get thoughts / feedback on doing the above ^ to handle testing the install generator properly with PostgreSQL?

How about using `DATABASE_URL` which overrides `config/database.yml` entirely, just for the postgres cases in the matrix?

adrianna-chang-shopify:
> How about using DATABASE_URL which overrides config/database.yml entirely, just for the postgres cases in the matrix?

I think we'll still run into the same problem -- if we use a db with the same name (`maintenance_tasks_test`), Postgres will complain that the `maintenance_tasks_runs` table exists already in the install generator test. We could use a `DATABASE_URL` instead of the additional db config file, but I believe we'd need to have a different database. We could use psql to create one in CI, but this makes things not-so-nice for local testing. 

Also looked into starting PostgreSQL directly on the Ubuntu VM, but couldn't quite get it -- not sure how to provide a password and / or specify no password. I tried
```bash
  sudo systemctl start postgresql.service
  psql -h localhost -U postgres -w --no-password 'create database maintenance_tasks_test;'
---
  sudo systemctl start postgresql.service
  PG_PASSWORD=postgres psql -h localhost -U postgres 'create database maintenance_tasks_test;'
---
  sudo systemctl start postgresql.service
  sudo -u postgres psql postgres -c 'create database maintenance_tasks_test;'
```

But no luck with any of them. I think it's acceptable to start postgres in another service. If you'd like to invest some more time into it @etiennebarrie let me know -- otherwise I'll ship this tmrw afternoon.

.github/workflows/ci.yml
shouichi: > Is there a way to have a service run conditionally?

As far as I can tell, there's no way to start a service conditionally. https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idservices

The github guide (https://docs.github.com/en/actions/guides/creating-postgresql-service-containers) uses docker for postgres, so it shouldn't be that bad. Also, using containers gives us reproducibility and composability which may amortize the cost of spinning up extra service when sqlite testing.

shouichi:
Thanks for the feedback. Speaking of the local development, I plan to open a follow-up PR to introduce docker based development environment. This way we don't need to set up postgres locally. Does this make sense?

adrianna-chang-shopify:
Agree that `DATABASE_URL` is a more correct solution to this problem 👍  So to clarify @shouichi, let's do the following before shipping:
- Add a step to CI to export the `DATABASE_URL` if the db is postgres, as Étienne did in 3b8f99edfc1558b8eed3437245c67c877ccfb219 (although to me it's more readable to have it as `echo "DATABASE_URL=postgresql://localhost/maintenance_tasks_test" >> $GITHUB_ENV`)
- Remove the changes to `config/database.yml`
- Change the install generator to use the `DATABASE_URL` instead, as proposed in 631c89df519b7ace188f2350fa4a981e4b2c9863

Thanks for your patience! 🙏

shouichi:
@adrianna-chang-shopify Thanks for the clarification. The `DATABASE_URL` solution is much cleaner, it's great! Changed based on the suggestions, rebased, and squashed. Please take a look.

PR #471, opened 2021-08-19
Include idempotency hint in generated task template.

NatMorcos:
The README makes it clear that `#process` needs to be idempotent. There are potentially major consequences of not adhering to this recommendation. We should remind developers inline while they are writing the `#process` method. 

The generated task file includes help comments already. I've extended them to include the idempotency hint.


adrianna-chang-shopify (approved):
Thanks for the contribution @NatMorcos ! ❤️

PR #474, opened 2021-08-24
Params test docs

gmcgibbon:
Adds task with parameters testing docs. Also a few general improvements that I can revert if we don't want them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ module Maintenance
     end
 
     def process(post)
-      post.update!(content: 'New content!')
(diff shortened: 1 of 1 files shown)


.ruby-version
```diff
@@ -1 +1 @@
-2.7.2
+3.0.2
```
adrianna-chang-shopify, line 1: 🙏

adrianna-chang-shopify (approved):
These look great! Thanks for fixing all the inconsistent styling in the README 😅

PR #483, opened 2021-09-09
Delimit ticks in progress text

kddnewton:
Had trouble reading the tick counts otherwise.

Also I created this in the GitHub editor and it was pretty slick but I also have no idea if it's right, so working on that.

### DIFF
--- a/app/models/maintenance_tasks/progress.rb
+++ b/app/models/maintenance_tasks/progress.rb
@@ -51,10 +51,10 @@ module MaintenanceTasks
       count = @run.tick_count
       total = @run.tick_total
 
-      count_delim = number_with_delimiter(count)
+      count_delim = number_to_delimited(count)
       return "Processed #{count_delim} #{"item".pluralize(count)}." unless total?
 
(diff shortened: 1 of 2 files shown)


app/models/maintenance_tasks/progress.rb
```diff
@@ -50,14 +50,21 @@ def max
     def text
       count = @run.tick_count
       total = @run.tick_total
-      if !total?
-        "Processed #{pluralize(count, "item")}."
-      elsif over_total?
-        "Processed #{pluralize(count, "item")} (expected #{total})."
```
etiennebarrie, line 56: Purely stylistic: I think I preferred the old way, with a bit of duplication but a straight if/elsif/else. If we really cared about duplication we could also de-duplicate the call to `"item".pluralize(count)`. I don't really like the local variables.

kddnewton: No worries I can change it!

etiennebarrie (approved):
Thanks!

Issue #488, opened 2021-09-21
Tasks with `csv_collection` tries to connect to the database when loaded


rafaelfranca:
It is caused by `ActiveStorage::Attachment.table_exists?` https://github.com/Shopify/maintenance_tasks/commit/a030de8e800675964f4453a9811a43ac8a01afa8.

This is a problem because it will try to connect to the database when booting, which might fail, if for example the database was not created yet, or the database is unhealthy.

etiennebarrie:
Good catch, I feel like we could easily guard that check against `Rails.env.production?` since it's only really helpful in development/test. I'm not a fan of doing this, but the alternative of checking whether the schema cache is used and primed seems pretty bad too. We could also just remove this but it was causing some support load.

I started looking more at whether we could test this but we can't because it seems it's not the only thing that prevents booting without a connection.

rafaelfranca:
I found this problem in the test environment, so I don't think checking environment will help. Can't we move this check to the execution of the task instead to the definition?

PR #491, opened 2021-09-28
Raise Active Storage error in CSV collection method

gmcgibbon:
Closes https://github.com/Shopify/maintenance_tasks/issues/488

Checks for Active Storage during `csv_file` call instead of during evaluation time of a `csv_collection` task. We don't appear to test this.

### DIFF
--- a/app/models/maintenance_tasks/task.rb
+++ b/app/models/maintenance_tasks/task.rb
@@ -53,10 +53,11 @@ module MaintenanceTasks
       # An input to upload a CSV will be added in the form to start a Run. The
       # collection and count method are implemented.
       def csv_collection
-        if !defined?(ActiveStorage) || !ActiveStorage::Attachment.table_exists?
+        unless defined?(ActiveStorage)
           raise NotImplementedError, "Active Storage needs to be installed\n"\
             "To resolve this issue run: bin/rails active_storage:install"
         end
+
         self.collection_builder_strategy = CsvCollectionBuilder.new
       end
 


adrianna-chang-shopify (approved):

nvasilevski (approved):

gmcgibbon:
> Also the way it is, it prevents running without Active Storage

Right... We can't have it talk to the DB at class eval time, so we could instead only do `defined?` in `csv_collection`. Does that make sense? That or we just remove the error entirely. Thoughts?

etiennebarrie:
> Also the way it is, it prevents running without Active Storage

Wait that was wrong, it was perfectly fine to run a non-CSV task without Active Storage.

adrianna-chang-shopify:
Indeed, I forgot that we can't raise inside `csv_file` because we use it (and expect it to just return early) in the views:
https://github.com/Shopify/maintenance_tasks/blob/9bbb30c1dd07bd5e9ec46d06ed5a82baa9bb7dc4/app/views/maintenance_tasks/runs/_csv.html.erb#L1

so we'd no longer allow AS to be optional.

I think it makes sense to only check `defined?` in `csv_collection` like you've done, Gannon. With the addition of https://github.com/Shopify/maintenance_tasks/pull/391, if we removed the check altogether from `csv_collection`, users end up seeing `undefined method `attach' for nil:NilClass` inside `Runner.run` because `Run#csv_file` just returns nil if AS isn't defined.

etiennebarrie (approved):

PR #494, opened 2021-09-30
Truncate the error message when message is too long

### LINKED ISSUE #493: Exception when error message is too long
Repro steps:

```sh-console
$ bundle add mysql2
$ export DATABASE_URL=mysql2://root@localhost/maintenance_tasks_development # or similar
$ brew services start mysql # or similar
$ bundle exec rails db:setup
```

Apply this patch
```diff
diff --git i/test/dummy/app/tasks/maintenance/error_task.rb w/test/dummy/app/tasks/maintenance/error_task.rb
index 730cea7..9f5d2a9 100644
--- i/test/dummy/app/tasks/maintenance/error_task.rb
+++ w/test/dummy/app/tasks/maintenance/error_task.rb
@@ -6,7 +6,7 @@ def collection
     end
 
     def process(input)
-      raise ArgumentError, "Something went wrong" if input == 2
+      raise ArgumentError, "Something went wrong#{"!"*255}" if input == 2
     end
   end
 end
```

And run the ErrorTask. The run can't be persisted as errored, the job stops, the error handler is not called in v1.5.0 but is called on main (because of #484 moving the callback to an `ensure` section). The run is stuck after that.

SQLite doesn't care about the size, postgresql is unlimited by default for Active Record string columns it seems, mysql uses 255 as the limit.

---

We should probably simply truncate, using the limit information provided by `columns_hash`, same with `error_class` and `backtrace` I guess.

cc @joelpinheiro

joelpinheiro:
Closes #493 to prevent [errors](https://app.bugsnag.com/shopify/arrive-server/errors/61fd2678bb89340008de6f8c?filters[search]=ActiveRecord%3A%3AValueTooLong) like:
```
ActiveRecord::ValueTooLongMaintenanceTaskJob@default
Mysql2::Error: Data too long for column 'error_message' at row 1
```

How I have 🎩'd it:
```
$ bundle add mysql2
$ export DATABASE_URL=mysql2://root@localhost/maintenance_tasks_development
$ brew services start mysql
$ bundle exec rails db:setup
$ bin/rails app:db:environment:set RAILS_ENV=test
$ bundle exec rails test
```

🎉

### DIFF
--- a/test/models/maintenance_tasks/run_test.rb
+++ b/test/models/maintenance_tasks/run_test.rb
@@ -164,6 +164,33 @@ module MaintenanceTasks
       run.persist_error(error)
     end
 
+    test "#persist_error does not raise on longer error class names" do
+      run = Run.create!(task_name: "Maintenance::ErrorTask")
+      error = ArgumentError.new("Something went wrong")
+      class_name = "SomeVeryLongErrorClassName #{"." * 20000}"
+      error.class.stubs(name: class_name)
+      error.set_backtrace(["lib/foo.rb:42:in `bar'"])
+
+      assert_nothing_raised do
+        run.persist_error(error)
+      end
+      limit = MaintenanceTasks::Run.column_for_attribute(:error_message).limit
+      assert_equal class_name.first(255), run.error_class if limit
+    end
+
+    test "#persist_error does not raise on longer errors messages" do
+      run = Run.create!(task_name: "Maintenance::ErrorTask")
+      error_name = "SomeVeryLongErrorMessage #{"." * 20000}"
+      error = ArgumentError.new(error_name)
+      error.set_backtrace(["lib/foo.rb:42:in `bar'"])
+
+      assert_nothing_raised do
+        run.persist_error(error)
+      end
+      limit = MaintenanceTasks::Run.column_for_attribute(:error_class).limit
+      assert_equal error_name.first(limit), run.error_message if limit
+    end
+
     test "#reload_status reloads status and lock version, and clears dirty tracking" do
       run = Run.create!(task_name: "Maintenance::UpdatePostsTask")
       original_lock_version = run.lock_version
--- a/app/models/maintenance_tasks/run.rb
+++ b/app/models/maintenance_tasks/run.rb
@@ -131,8 +131,8 @@ module MaintenanceTasks
       self.started_at ||= Time.now
       update!(
         status: :errored,
-        error_class: error.class.to_s,
-        error_message: error.message,
+        error_class: truncate(:error_class, error.class.name),
+        error_message: truncate(:error_message, error.message),
         backtrace: MaintenanceTasks.backtrace_cleaner.clean(error.backtrace),
         ended_at: Time.now,
       )
@@ -421,6 +421,12 @@ module MaintenanceTasks
         MSG
         errors.add(:base, error_message)
       end
+    end
+
+    def truncate(attribute_name, value)
+      limit = MaintenanceTasks::Run.column_for_attribute(attribute_name).limit
+      return value unless limit
+      value&.first(limit)
     end
   end
 end


test/jobs/maintenance_tasks/task_job_test.rb
```diff
@@ -131,7 +131,7 @@ class TaskJobTest < ActiveJob::TestCase
 
       run.expects(:persist_error).with do |exception|
         assert_kind_of ArgumentError, exception
-        assert_equal "Something went wrong", exception.message
+        assert_equal "Something went wrong #{"!" * 255}", exception.message
```
tgwizard, line 134: This message is longer than 255 characters, so what does the assertion check here?

joelpinheiro: It asserts that even if the error is longer than 255 chars the error will be correctly persisted.

test/models/maintenance_tasks/run_test.rb
```diff
@@ -164,6 +164,25 @@ class RunTest < ActiveSupport::TestCase
...
+      end
+    end
+
+    test "#persist_error does not raise on longer errors messages" do
+      run = Run.create!(task_name: "Maintenance::ErrorTask")
+      error = ArgumentError.new("SomeVeryLongErrorMessage #{"." * 20000}")
+      error.set_backtrace(["lib/foo.rb:42:in `bar'"])
+      assert_nothing_raised do
+        run.persist_error(error)
+      end
```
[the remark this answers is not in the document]
joelpinheiro: That would depend on the database in use and this test is database agnostic. Currently, CI is running with `sqlite` therefore there's no limit. I ran this test using `mysql` and without the `truncate(attribute_name, value)` it would fail.

@etiennebarrie will bring `mysql` into the CI and then we can run the tests for each supported database.

Gemfile
```diff
@@ -6,6 +6,7 @@ gemspec
 
 gem "capybara"
 gem "mocha"
+gem "mysql2", "~> 0.5.3"
```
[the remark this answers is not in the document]
joelpinheiro: We don't need it. Removed! 👍

test/models/maintenance_tasks/run_test.rb
```diff
@@ -164,6 +164,33 @@ class RunTest < ActiveSupport::TestCase
       run.persist_error(error)
     end
 
+    test "#persist_error does not raise on longer error class names" do
+      run = Run.create!(task_name: "Maintenance::ErrorTask")
+      error = ArgumentError.new("Something went wrong")
+      class_name = "SomeVeryLongErrorClassName #{"." * 20000}"
+      error.stubs(class: class_name)
```
tgwizard, line 171: (This is weird - `class` should return a Class object, not a string. Can we instead create a new error class with a name that's > 255 character long?)

tgwizard (approved):
I think this kinda looks good, but I think we could just hard code the limit to 255 and let that be, no need to handel different DBs differently etc.

Please wait for a maintenance-task-team person to review as well.

etiennebarrie:
> I think this kinda looks good, but I think we could just hard code the limit to 255 and let that be, no need to handel different DBs differently etc.

I see your point but really since the engine ships in an embedding app, we don't really control what people do to their database schema, even if it looks like we do. I think it's ok to give users more flexibility here.
In the tests we can hardcode 255 because we know using a `string` column and MySQL you get `varchar(255)`.

Plus ActiveRecord gives us enough information to solve that in a flexible manner that lets people using PostgreSQL enjoy their (mostly) unbounded `character varying` columns.

.gitignore
```diff
@@ -13,3 +13,4 @@ test/dummy/log/*.log
 test/dummy/public/assets
 test/dummy/storage/
 test/dummy/tmp/
+.idea
```
etiennebarrie, line 16: Put this in your own `~/.config/git/ignore` https://git-scm.com/docs/gitignore

PR #495, opened 2021-09-30
Update Rails to 6.1.4.1

etiennebarrie:
Because of our version requirement on `rails` to be able to change the Rails version depending on the Gemfile, Dependabot is not able to generate update PRs for Rails.
https://github.com/Shopify/maintenance_tasks/blob/1a8d85cafac4ab815afc446ae01202716bc7bc48/Gemfile#L12

```
updater | ERROR <job_215152768> Error processing rails (NoMethodError)
updater | ERROR <job_215152768> undefined method `begin' for #<Parser::Source::Map::Operator:0x000055581953a450>
updater | ERROR <job_215152768> /home/dependabot/dependabot-updater/vendor/ruby/2.7.0/gems/dependabot-bundler-0.162.1/lib/dependabot/bundler/file_updater/requirement_replacer.rb:214:in `range_for'
updater | ERROR <job_215152768> /home/dependabot/dependabot-updater/vendor/ruby/2.7.0/gems/dependabot-bundler-0.162.1/lib/dependabot/bundler/file_updater/requirement_replacer.rb:109:in `on_send'
updater | ERROR <job_215152768> /home/dependabot/dependabot-updater/vendor/ruby/2.7.0/gems/ast-2.4.2/lib/ast/processor/mixin.rb:259:in `process'
updater | ERROR <job_215152768> /home/dependabot/dependabot-updater/vendor/ruby/2.7.0/gems/ast-2.4.2/lib/ast/processor/mixin.rb:276:in `block in process_all'
updater | ERROR <job_215152768> /home/dependabot/dependabot-updater/vendor/ruby/2.7.0/gems/ast-2.4.2/lib/ast/processor/mixin.rb:275:in `map'
updater | ERROR <job_215152768> /home/dependabot/dependabot-updater/vendor/ruby/2.7.0/gems/ast-2.4.2/lib/ast/processor/mixin.rb:275:in `process_all'
updater | ERROR <job_215152768> /home/dependabot/dependabot-updater/vendor/ruby/2.7.0/gems/parser-3.0.2.0/lib/parser/ast/processor.rb:11:in `process_regular_node'
updater | ERROR <job_215152768> /home/dependabot/dependabot-updater/vendor/ruby/2.7.0/gems/ast-2.4.2/lib/ast/processor/mixin.rb:259:in `process'
[...]

PR #496, opened 2021-10-01
Simple rails requirement that Dependabot can update

etiennebarrie:
Replaces #495

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,7 +9,13 @@ gem "mocha"
 gem "pg"
 gem "pry-byebug"
 gem "puma"
-gem "rails", defined?(@rails_gem_requirement) && @rails_gem_requirement
+if defined?(@rails_gem_requirement) && @rails_gem_requirement
+  # causes Dependabot to ignore the next line and update the next gem "rails"
+  rails = "rails"
+  gem rails, @rails_gem_requirement
+else
+  gem "rails"
+end
 gem "rubocop"
 gem "rubocop-shopify"
 gem "selenium-webdriver"


Gemfile
```diff
@@ -9,7 +9,12 @@ gem "mocha"
 gem "pg"
 gem "pry-byebug"
 gem "puma"
-gem "rails", defined?(@rails_gem_requirement) && @rails_gem_requirement
+if defined?(@rails_gem_requirement) && @rails_gem_requirement
+  rails = "rails" # confuses Dependabot
```
adrianna-chang-shopify, line 13:
```suggestion
  rails = "rails" # allows Dependabot to update rails properly
```
Nit: I interpret "confuses" as this causing issues for Dependabot, when really we're "tricking" Dependabot into doing the right thing

adrianna-chang-shopify (commented):
Can you explain to me a bit more about how this works? Does parsing `rails = "rails"` result in the AST parser skipping the next line? 

Were you able to test this locally?

etiennebarrie:
> Can you explain to me a bit more about how this works? Does parsing `rails = "rails"` result in the AST parser skipping the next line?

Yes, so usually Dependabot doesn't bother updating the Gemfile if it can avoid to (like there's no requirement) but it seems like our code triggers that anyway, based on the backtrace I shared in #495. In any case, once you're in the RequirementReplacer class, it will [look for a call to "gem"](https://github.com/dependabot/dependabot-core/blob/main/bundler/lib/dependabot/bundler/file_updater/requirement_replacer.rb#L124) whose [first parameter is a string of the name of the dependency](https://github.com/dependabot/dependabot-core/blob/18785ae20d06c4c86e794a8c6542aa0c110afc20/bundler/lib/dependabot/bundler/file_updater/requirement_replacer.rb#L133) it's trying to update. Using a local variable is enough to have it not consider this "gem" line at all.

To run it I basically tweaked RequirementReplacer until I could run it on my file with the rails dependency and version 6.1.4.1. I had the same error and then I made these changes and it updated fine (adding the requirement to the Gemfile as a side-effect, I think it's fine if it has to since it's able to update it later, but I'm not even sure it will do it once shipped.

Gemfile
etiennebarrie: Yeah we're tricking it into doing the right thing by confusing it on this line.

PR #499, opened 2021-10-01
Make Job class configurable per Task

adrianna-chang-shopify:
## Motivation

To support tasks in the global ("master") context in Shopify/shopify, we need to use a separate job class that disables all of the automatic pod selection logic. This is done at load time, so I don't think there's a way to handle this other than using distinct job classes to handle non-podded vs podded tasks.

One possible way to accomplish this is to inject the job class into the runner, just as we're doing with the `run_model`. See https://github.com/Shopify/maintenance_tasks/pull/498/commits/1cce03d2cac084dd2a7a7c8342d5007b495bcb50

However, we'd [previously discussed](https://github.com/Shopify/maintenance_tasks/pull/96#pullrequestreview-517821649) allowing job classes to be configured on a per-task basis, so I figured I'd put this up and see how we felt about it. IMO, while allowing the job to be injected into the Runner serves us for Core, it isn't that useful for generic gem users. A class-level method on `Task` seems to be a more appropriate solution in terms of bringing value to _all_ users of the gem (not just for our own use case in Core), and as much as possible I think we should try to keep code out of the gem that's _just_ for Core.

The counter-argument is that this gives users too much flexibility. But I think if we keep the app-level config it's okay.

## Solution

Offer a `with_job_class` (needs a better name?) macro that can be set on a Task class, in order to configure the job class to use for that Task. We can keep the app-level config -- this value will only take precedence if it's set.

Thoughts?

## To Do
- [ ] Add docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -604,6 +604,29 @@ The Job class **must inherit** from `MaintenanceTasks::TaskJob`.
 Note that `retry_on` is not supported for custom Job classes, so failed jobs
 cannot be retried.
 
+#### Customizing the Job for a single Task
+
+A Job class can also be specified on an individual Task, overriding the global
+configuration, using `with_job_class`:
+
+```ruby
+# app/jobs/high_priority_job.rb
+
+class HighPriorityJob < MaintenanceTasks::TaskJob
+  queue_as :high_priority
+end
+
+# app/tasks/maintenance/really_important_task.rb
+
+module Maintenance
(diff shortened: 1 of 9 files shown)


dirceu (approved):

paarthmadan (approved):
Nice!

nvasilevski (approved):

PR #515, opened 2021-11-02
Fix race condition when moving to running

etiennebarrie:
If the move to cancelling happens after the enqueued/interrupted run was loaded in the job, but before it's moved to running, it's possible that a cancelling run returns to running.

This fixes it by only moving the run to running if it wasn't stopping, doing it in the SQL query rather than in Ruby code.

We had solved the same race condition in #178 by using optimistic locking, this does an "optimistic update" and checks whether it went through.


adrianna-chang-shopify (approved):

etiennebarrie:
Just added an assertion for the dirty handling.

PR #522, opened 2021-11-08
Ignore Selenium warning we don't control

etiennebarrie:
Since it's due to Rails' system tests and fixed there for Rails 7, we can ignore the warning for previous versions.

### DIFF
--- a/test/application_system_test_case.rb
+++ b/test/application_system_test_case.rb
@@ -6,6 +6,15 @@ require "action_dispatch/system_testing/server"
 
 ActionDispatch::SystemTesting::Server.silence_puma = true
 
+if Rails::VERSION::MAJOR < 7
+  # Necessary so that Capybara::Selenium::DeprecationSuppressor is prepended in
+  # Selenium::WebDriver::Logger before it is instantiated in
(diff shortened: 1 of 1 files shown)


test/application_system_test_case.rb
```diff
@@ -6,6 +6,15 @@
 
 ActionDispatch::SystemTesting::Server.silence_puma = true
 
+if Rails::VERSION::MAJOR < 7
+  # Necessary so that Capybara::Selenium::DeprecationSuppressor is prepended in
+  # Selenium::WebDriver::Logger before it is instantiated in
+  # Selenium::WebDriver.logger to prevent an uninitialized instance variable
+  # warning.
```
adrianna-chang-shopify, line 13: Is it the `@suppress_for_capybara` ivar that's uninitialized if we don't load selenium & prepend `Capybara::Selenium::DeprecationSuppressor`? 🤔  I don't see any warnings, even if I comment out the `load_selenium` line, and wondering what I'm missing 😄

adrianna-chang-shopify: Ah, version thing, got it 👍  Thanks!

adrianna-chang-shopify (approved):

PR #525, opened 2021-11-12
Update Bugsnag.notify error handler example

tjwp:
`Bugsnag::Report#add_tab` is deprecated, so this updates the README example to use the recommended `#add_metadata` method instead.

I've also included a line marks the error report as `unhandled`. This is more opinionated and I'd be happy to remove it from the example if it is controversial. By default `Bugsnag.notify` reports errors as `handled`. My perspective is that if an error makes it to this handler is should be treated as `unhandled` in Bugsnag, similar to an uncaught error from a controller or a job that reported by the default Bugsnag integration.


gmcgibbon (approved):

etiennebarrie (commented):
I think unhandled makes sense in the sense that's it's not the user who caught an exception and reported it, a bit like described here:
> unhandled exceptions in your Rails app will be automatically reported to your Bugsnag dashboard
> https://docs.bugsnag.com/platforms/ruby/rails/#reporting-unhandled-exceptions

But on the other hand, from the documentation on the unhandled attribute:
> Changing the unhandled flag for an event will affect how it contributes to your application’s stability score. 
> https://docs.bugsnag.com/platforms/ruby/rails/reporting-handled-errors/#unhandled


> # Errors that impact stability
> Any session in which a user experiences an “unhandled” error is considered a crashing session and will impact the stability score. 
> https://docs.bugsnag.com/product/stability/#errors-that-impact-stability

I'm not so sure overall. If you care about the stability score, maybe you don't care that your maintenance task crashed as it wouldn't have impacted your users.

Also it seems technically if we want to be similar to what bugsnag does automatically, we should be using `Bugsnag.notify(error, true)`, which seems to not change the severity reason https://github.com/bugsnag/bugsnag-ruby/blob/94da895ad148b9cd132099a6a0be5710a5f03e35/lib/bugsnag.rb#L140-L142

Maybe let's just ignore this for now, and individual apps can figure out what they want from their bug reporter.

Issue #528, opened 2021-11-15
Classic autoloader support


gmcgibbon:
Classic autoload reloading seems to be broken in the dummy app (and apps that use the classic autoloader). You can reproduce the error with the following:

- `cd test/dummy`
- `CLASSIC_AUTOLOADER=1 bin/rails c`
- `reload!`

```
Reloading...
~/src/github.com/Shopify/maintenance_tasks/app/models/maintenance_tasks/task.rb:62:in `csv_collection': uninitialized constant CsvCollectionBuilder (NameError)
```

Adding a `require_dependency` to `app/models/maintenance_tasks/task.rb` fixes the problem, but our support for non-Zeitwerk applications is unclear. Should we support them, or not? If not, I suggest we remove the ENV var to boot in classic mode. Also, keep in mind Rails 7 will not include the classic autoloader as Zeitwerk has been the default for quite awhile now.

adrianna-chang-shopify:
> our support for non-Zeitwerk applications is unclear

We currently ensure that we're compatible with the classic autoloader by eager loading the app and requiring all task files in the engine:

https://github.com/Shopify/maintenance_tasks/blob/05f03c4467422934120963ce4afcbd19086c5fc0/lib/maintenance_tasks/engine.rb#L11-L13

https://github.com/Shopify/maintenance_tasks/blob/05f03c4467422934120963ce4afcbd19086c5fc0/lib/maintenance_tasks/engine.rb#L21-L26

So I'd say that we have tried to support it up until this point 😄  That said, I'm not sure it's worth doing. As you said, Zeitwerk is the default in Rails 6 apps, and classic autoloading won't be supported soon, so I think it's fair game to enforce Zeitwerk. It also lets us get rid of the above code. I'd be OK to disable classic mode and call it out in the docs.

etiennebarrie:
I think it was just a matter of qualifying `CsvCollectionBuilder` (=> `MaintenanceTasks::CsvCollectionBuilder`).

I couldn't write a test to cause the exception though (using `Rails.application.reloader.reload!` or even including the `Rails::ConsoleMethods` and calling `reload!` wasn't working from a rake task for some reason).

I'm conflicted because this particular means of testing autoloading has been failing forever basically (since v1.1.1), but in general autoloading from the UI works, and you can run your task. I mean it's pretty bad that we broke `reload!`, but you could run your tasks with the classic autoloader, now the UI is entirely broken.
[...]

PR #530, opened 2021-11-16
Revert "Drop support for classic autoloading"

gmcgibbon:
Reverts Shopify/maintenance_tasks#529

We can revert this revert once we've properly deprecated classic autoloading.


adrianna-chang-shopify (approved):

simonlevasseur:
What's the context on this revert?

gmcgibbon:
> What's the context on this revert?

See https://github.com/Shopify/maintenance_tasks/issues/528

Issue #485, opened 2021-09-17
Task parameter defaults are ignored


etiennebarrie:
Since we have form inputs for all the task parameters, parameter defaults are overridden by form values, even when the form input is empty.

A possible workaround is to open developer tools and remove the input or disable it. 😅 

It works with the command-line runner.

In terms of fix, I think we can't know if an attribute has a default because it's stored in `_default_attributes` and it looks like a pretty internal API. We could populate the fields with their default values but that seems a bit fishy to call the default, serialize it to a string, put in the HTML, get it back through HTTP and deserialize it back to an object. I mean that could work.

Maybe instead we could just remove attributes from the hash where the value is just an empty string, and let Active Model do the default only once before enqueuing.

nvasilevski:
>We could populate the fields with their default values but that seems a bit fishy to call the default, serialize it to a string, put in the HTML, get it back through HTTP and deserialize it back to an object.

For some reason I'm leaning towards this option. I'm terrible at UX but I feel like this seems to be a very user-friendly option, to actually see default values on the form
For example having `attribute :content, :string, default: "default content"` attribute we would see it as:
![image](https://user-images.githubusercontent.com/5512772/142268969-a4212c49-4b3b-4eef-a30a-e9db924c4145.png)

The problem that `_default_attributes` indeed seems to be a private API. Wonder if we could consider exposing default values publicly in Rails 🤔 
Should we try getting default values from the object, for example:
```ruby
irb(main):005:0> Maintenance::ParamsTask.new.attributes
=> {"post_ids"=>nil, "content"=>"default content"}
```

It seems to be more or less common way of rendering form for a non-persisted object in the scope of the `new` action for a scaffold-generated Rails model 🤔

etiennebarrie:
Yes using a new object and accessing the value is the only API I know to get to the defaults. Serializing/deserializing to a HTML form inputs is probably OK most of the time but will be a bit weird with booleans/time etc. Maybe we could get a bit fancy on the inputs depending on the type.

nvasilevski:
>Maybe we could get a bit fancy on the inputs depending on the type.

Yep, definitely 👍 
I think it would be reasonable to go ahead and render default values in the existing `text_area` fields and consider rendering different input types per attribute type as a separate task

PR #532, opened 2021-11-18
Render task default attributes on the form

nvasilevski:
Addresses - https://github.com/Shopify/maintenance_tasks/issues/485

Currently blank attribute values passed from the form override all default values

The solution is far from being perfect but seems to be the most reasonable from UX point of view.
It seems helpful to render default values on the form so anyone can be aware of them just by looking on the UI 

Example:
image

### Solution details

Using an instance of a task class seems to be the most appropriate and most publicly available API at the moment to get default values

### DIFF
--- a/app/views/maintenance_tasks/tasks/show.html.erb
+++ b/app/views/maintenance_tasks/tasks/show.html.erb
@@ -31,14 +31,15 @@
           <%= form.file_field :csv_file %>
         </div>
       <% end %>
-      <% if @task.parameter_names.any? %>
+      <% parameter_names = @task.parameter_names %>
+      <% if parameter_names.any? %>
         <div class="block">
-          <%= form.fields_for :task_arguments do |ff| %>
-            <% @task.parameter_names.each do |parameter| %>
+          <%= form.fields_for :task_arguments, task_object_for(@task) do |ff| %>
+            <% parameter_names.each do |parameter_name| %>
               <div class="field">
-                <%= ff.label parameter, parameter, class: "label is-family-monospace" %>
+                <%= ff.label parameter_name, parameter_name, class: "label is-family-monospace" %>
                 <div class="control">
-                  <%= ff.text_area parameter, class: "textarea" %>
+                  <%= ff.text_area parameter_name, class: "textarea" %>
                 </div>
               </div>
             <% end %>
--- a/test/models/maintenance_tasks/task_data_test.rb
+++ b/test/models/maintenance_tasks/task_data_test.rb
@@ -149,12 +149,13 @@ module MaintenanceTasks
     end
 
     test "#parameter_names returns list of parameter names for Tasks supporting parameters" do
-      assert_equal ["post_ids"],
+      assert_equal ["post_ids", "content", "integer_attr"],
         TaskData.new("Maintenance::ParamsTask").parameter_names
     end
 
     test "#parameter_names returns empty list for deleted Tasks" do
-      assert_empty TaskData.new("Maintenance::DoesNotExist").parameter_names
+      names = TaskData.new("Maintenance::DoesNotExist").parameter_names
+      assert_equal [], names
     end
   end
 end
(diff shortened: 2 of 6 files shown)


app/models/maintenance_tasks/task_data.rb
```diff
@@ -133,17 +133,22 @@ def csv_task?
...
       else
-        Task.named(name).attribute_names
+        task_object.attributes
       end
     end
 
     private
 
+    def task_object
+      @task_object ||= Task.named(name).new
```
nvasilevski, line 149: I was considering memoizing `Task.named(name)` lookup as well, in a separate PR, but might be premature

gmcgibbon: If we assign `parameters = @task.parameters`, I don't think memoization is needed.

app/views/maintenance_tasks/tasks/show.html.erb
```diff
@@ -31,14 +31,15 @@
           <%= form.file_field :csv_file %>
         </div>
       <% end %>
-      <% if @task.parameter_names.any? %>
+      <% parameters = @task.parameters %>
+      <% if parameters.any? %>
         <div class="block">
           <%= form.fields_for :task_arguments do |ff| %>
```
etiennebarrie, line 37: We can pass an object to `fields_for`, and then we don't need to pass the value for the form helpers. Then we don't need to change `parameter_names` into `parameters`.

nvasilevski: Definitely an option! The only reason why I haven't implemented it because I was hesitant of making `task_object` publicly accessible 

>Then we don't need to change parameter_names into parameters.

I wonder if with `task_object` being publicly accessible we need the `TaskData#parameter_names` at all? We could just get it from `@task_object.attribute_names` 

^ While I was writing this I realized that it would be better to keep and use `TaskData#parameter_names` method as "parameters" is a maintenance task concept and "attributes" is an implementation detail so it would be better to hide it a little bit

I'll update the form 👍

app/models/maintenance_tasks/task_data.rb
nvasilevski: I end up keeping the memoization as `task_object` method became public and we are using it twice (inside `parameter_names` and for `fields_for`)

adrianna-chang-shopify: I don't think we need to create a new task instance for `#parameter_names` though since we can just do `Task.named(name).attribute_names` 😄 

I'd prefer not to expose this new method, to leave `#parameter_names` as is, and to only init the Task object in the view:
```erb
<%= form.fields_for :task_arguments, Task.named(@task).new do |ff| %>
```

nvasilevski: Done!
`task_object = MaintenanceTasks::Task.named(@task.name).new` seems like a little too much code for a view but I feel even more uncomfortable exposing task object and documenting it so let's try initializing it in the view 👍

test/models/maintenance_tasks/task_data_test.rb
```diff
@@ -149,12 +149,13 @@ class TaskDataTest < ActiveSupport::TestCase
...
     test "#parameter_names returns list of parameter names for Tasks supporting parameters" do
-      assert_equal ["post_ids"],
+      assert_equal ["post_ids", "content", "integer_attr"],
         TaskData.new("Maintenance::ParamsTask").parameter_names
     end
 
     test "#parameter_names returns empty list for deleted Tasks" do
-      assert_empty TaskData.new("Maintenance::DoesNotExist").parameter_names
+      names = TaskData.new("Maintenance::DoesNotExist").parameter_names
+      assert_equal [], names # to make it's not any empty object
```
adrianna-chang-shopify, line 158: Nit: I think it's clear without the comment! 😄 
```suggestion
      assert_equal [], names
```

adrianna-chang-shopify (approved):
🚀

gmcgibbon (approved):

PR #537, opened 2021-11-22
Move Task instantiation to TaskData

etiennebarrie:
Follow up to #532

I feel like the presenter is a good place to deal with this, TaskData takes the place of a Task class in views (we even call its instance `@task`), so since we want `Task.new` to instantiate a task and get its attributes for defaults, we can use instead `TaskData#new`.

Also added an edge case about deleted Tasks. It's not strictly necessary since `parameter_names` is empty when the task is deleted, and we don't show the form and instantiate the Task in that situation, but it's more for completeness.

### DIFF
--- a/app/helpers/maintenance_tasks/tasks_helper.rb
+++ b/app/helpers/maintenance_tasks/tasks_helper.rb
@@ -100,11 +100,5 @@ module MaintenanceTasks
         only_path: true
       )
     end
-
-    # @param task_data [TaskData] an instance of the TaskData.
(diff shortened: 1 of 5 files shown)


nvasilevski (approved):
Looks good to me 👍 
I don't have any strong preference on this and reasoning behind this change makes sense

adrianna-chang-shopify (approved):
Yeah, I forgot that `TaskData` was a presentation object, so it does indeed make sense for the logic to live there over a helper 😄

PR #542, opened 2021-12-01
Fix broken .yardopts

ryanwilsonperkin:
Same as the fix in https://github.com/Shopify/job-iteration/pull/160
https://github.com/Shopify/maintenance_tasks/pull/540#issuecomment-984006384


adrianna-chang-shopify (approved):
Thanks!

PR #544, opened 2021-12-02
Add .devcontainer config

rafaelfranca:
This config can be used with GitHub Codespaces and
VSCode Remote - Containers, so it make easier to contributors to get a
working development environment.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,37 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
+// https://github.com/microsoft/vscode-dev-containers/tree/v0.205.2/containers/ruby
+{
+	"name": "Ruby",
+	"build": {
+		"dockerfile": "Dockerfile",
+		"args": {
+			// Update 'VARIANT' to pick a Ruby version: 3, 3.0, 2, 2.7, 2.6
+			// Append -bullseye or -buster to pin to an OS version.
+			// Use -bullseye variants on local on arm64/Apple Silicon.
+			"VARIANT": "3-bullseye",
+			// Options
+			"NODE_VERSION": "lts/*"
+		}
+	},
+
+	// Set *default* container specific settings.json values on container create.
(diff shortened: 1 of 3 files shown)


nvasilevski (approved):
👍 
Wonder if there is anything worth adding to README about these capabilities? 
Or perhaps reading about codespaces/ vscode containers in general is enough to get started 🤔

sambostock (commented):
Does this handle the installation of PostgreSQL? That's something I ran into locally when I tried to `bundle install`, because the `pg` gem needed some part of it to compile the native extensions.

rafaelfranca:
> Does this handle the installation of PostgreSQL? That's something I ran into locally when I tried to `bundle install`, because the `pg` gem needed some part of it to compile the native extensions.

The development headers (`libpq`) are in the base image. If we want the postgresql server I need to add.

gmcgibbon:
I can't seem to get codespaces working on a fork, but using this as a local development container works well. Maybe we need the main repo to have this config first? In order to run tests, we need chrome for the webdrivers gem because we have system tests.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,37 @@
...
+	// Add the IDs of extensions you want installed when the container is created.
+	"extensions": [
+		"rebornix.Ruby"
+	],
+
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
+	// "forwardPorts": [],
+
+	// Use 'postCreateCommand' to run commands after the container is created.
+	"postCreateCommand": "bundle install",
```
etiennebarrie, line 29: This is where we could use a bin/setup that also sets up the database.

etiennebarrie (commented):
I tried a bit but can't get chromedriver to work.

nvasilevski:
Just a quick update: `pg` is not a development requirement anymore, though we still need to look into being able to run system tests

gmcgibbon:
@rafaelfranca are you still interested in working on this?

PR #546, opened 2021-12-06
Ignore tasks defined outside of Task module

etiennebarrie:
This changes the gem to only consider Tasks (MaintenanceTasks::Task subclasses) defined in the `MaintenanceTasks.tasks_module` (`Maintenance` by default).

Previously it was possible to define a Task anywhere:

* It wouldn't show in the list by default in development (the constant would not have been loaded).
* It would show in production though.
* In any case it was possible to access the task URL directly and start running.
* Running would work normally.

I think the first point in particular is not ideal, and we have no way of finding all the subclasses without knowing where to look for them when eager loading is not on.

With these changes:
* Accessing the URL directly doesn't work.
* Running will cause a validation error (using Runner or the command line).

This is a breaking change but it's also a use case that wasn't supported so I think it's OK to do this.

### DIFF
--- a/app/models/maintenance_tasks/task.rb
+++ b/app/models/maintenance_tasks/task.rb
@@ -37,6 +37,13 @@ module MaintenanceTasks
         unless task.is_a?(Class) && task < Task
           raise NotFoundError.new("#{name} is not a Task.", name)
         end
+        unless task.module_parent_name == MaintenanceTasks.tasks_module
+          raise NotFoundError.new(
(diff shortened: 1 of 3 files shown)


nvasilevski (approved):
👍

adrianna-chang-shopify (approved):

etiennebarrie:
Actually we want to support multiple modules instead of enforcing one.

PR #547, opened 2021-12-06
Allow backoff to be specified via a proc

adrianna-chang-shopify:
Allow backoff to be provided via a callable object, rather than requiring a hardcoded `ActiveSupport::Duration`. This allows the backoff to be computed when the enumerator is being built, rather than being assigned when the task code is loaded.

This will facilitate removing some of the custom throttle enumerator usage in `Shopify/shopify`, allowing us to define the throttle conditions directly on a base task class.


app/jobs/concerns/maintenance_tasks/task_job_concern.rb
```diff
@@ -65,7 +65,13 @@ def build_enumerator(_run, cursor:)
...
 
       @task.throttle_conditions.reduce(collection_enum) do |enum, condition|
-        enumerator_builder.build_throttle_enumerator(enum, **condition)
+        throttle_on, backoff = condition.values
+        backoff = backoff.call if backoff.respond_to?(:call)
+        enumerator_builder.build_throttle_enumerator(
+          enum,
+          throttle_on: throttle_on,
+          backoff: backoff
+        )
```
nvasilevski, line 74: Just a nitpick but perhaps it would be better to avoid relying on a particular values order just in case we may accidentally or intentionally change the structure here
https://github.com/Shopify/maintenance_tasks/blob/3a07296714c25043072b2c54702dfa16c38febd2/app/models/maintenance_tasks/task.rb#L109

```suggestion
        backoff = condition[:backoff]
        backoff = backoff.call if backoff.respond_to?(:call)
        enumerator_builder.build_throttle_enumerator(
          enum,
          throttle_on: condition[:throttle_on],
          backoff: backoff
        )
```

nvasilevski (approved):

app/jobs/concerns/maintenance_tasks/task_job_concern.rb
adrianna-chang-shopify: Good point, done! ❤️

etiennebarrie (approved):
Just a tiny proposition to tweak the place we check for `call`.

app/models/maintenance_tasks/task.rb
```diff
@@ -99,9 +99,11 @@ def count
...
-      # @param backoff [ActiveSupport::Duration] optionally, a custom backoff
+      # @param backoff [ActiveSupport::Duration, #call] a custom backoff
       #   can be specified. This is the time to wait before retrying the Task.
-      #   If no value is specified, it defaults to 30 seconds.
+      #   If no value is specified, it defaults to 30 seconds. Alternatively,
+      #   an object responding to call can be used. It must return an
+      #   ActiveSupport::Duration.
       # @yieldreturn [Boolean] where the throttle condition is being met,
       #   indicating that the Task should throttle.
       def throttle_on(backoff: 30.seconds, &condition)
```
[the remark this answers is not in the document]
adrianna-chang-shopify: Sure, can do!

PR #549, opened 2021-12-14
Optimistic locking + fix data races

adrianna-chang-shopify:
Take 2 of https://github.com/Shopify/maintenance_tasks/pull/178

## Problem
Data races are a problem in the gem because:
1) We update the status of a `MaintenanceTasks::Run` object frequently as it progresses through the lifecycle of a Task, and
2) A `run` is updated by a running `TaskJob`, but also by the controller when pausing / resuming / cancelling a Task. This can create race conditions between writes. The primary scenario is that the `@run` in the job gets "stale" if a write happens between when we read the status value for an in-memory run object (where this value determines what status we should write), and when we actually perform the write.

Why is this a problem?

- It can lead to tasks not getting cancelled properly. If an interrupt happens at the same time as the task is cancelled, the `interrupted` state may end up getting persisted and the task won't stop.
- It can lead to tasks getting into a stuck state if the task is cancelled or paused as it succeeds -- if the run moves to succeeded just as the user cancels, the run can get updated to `cancelling`, but the job will be finished, so the run will stay stuck.
- In general, the nature of the gem is highly conducive to races, so we should address this properly so that unexpected state transitions / status changes "not going through" don't keep happening.

## Solution
This PR uses [`ActiveRecord::Locking::Optimistic`](https://api.rubyonrails.org/classes/ActiveRecord/Locking/Optimistic.html) to handle data races. Any scenarios we have seen or suspect might cause a data race are handled by:
1) Rescuing the `ActiveRecord::StaleObjectError`
2) Reloading the status on the `run` record
3) Retrying the status transition

Any data races that are missed will result in an unhandled `ActiveRecord::StaleObjectError` being raised, and will let us know there is a race condition to be handled.

## Summary of changes
- Adds `lock_version` column to `maintenance_tasks_runs` table, which will automatically set us up to use optimistic locking on this model
- Ensure we refresh lock version when reloading status so that any time we're fetching an updated status, we're also taking into account the latest lock version for that status
- Ensure we also refresh the lock version after persisting progress in `on_shutdown` -- `update_counters` [handles optimistic locking](https://api.rubyonrails.org/classes/ActiveRecord/Locking/Optimistic/ClassMethods.html#method-i-update_counters) in the db, but won't update the lock version for our in-memory `@run`. We don't need to do this when persisting progress (`@ticker.tick`) in `#each_iteration`, because we call `@run.reload_status` immediately after, which will reload the lock version.
- Converts all of the status transition methods on the `Run` to rescue, reload, and retry stale object errors. For `#running`, I decided to leave our swap-and-replace implementation in the case where optimistic locking is not being used, to preserve backwards compatibility for users who may not want to update their dbs right away. I also extracted `#start` to its own method, and it handles optimistic locking.
- Handling data races between interrupt / pause / cancel are a bit trickier due to the fact that assigning the status to the object in memory, and persisting the object to the db, are done in separate statements (and in separate callbacks!)

I'll go into the commits in a bit more detail:

[Address race conditions between interrupts and pause / cancel operation](https://github.com/Shopify/maintenance_tasks/pull/549/commits/d4800bb4d6df86df716b3da7e1262a7644935b6a):
- Handles race conditions between the run being "interrupted" after a pause / cancel, and the job not actually stopping
- AFTER we set the status on the run in `shutdown`, we check the db lock version against the run's current lock version. If the values are different, we manually raise a stale object error (this would replicate the behaviour we would see if we were _writing_ to the db with these status updates instead of just assigning the attribute)
- We will then rescue, reload, and retry as usual

[Handle data races between run succeeding and being paused / cancelled](https://github.com/Shopify/maintenance_tasks/pull/549/commits/baf3ab73bb9e640d3241677b5a1e1e95b97eade1)
- Handles races where we pause / cancel a run in between the status being set to `:succeeded` and the run actually being saved in the `after_perform` callback
- Basically, we just use the same rescue, reload and retry flow in `after_perform`, WITH the caveat of needing to save the status at the start so that we don't lose it when we reload

[ Early return when cancelling / pausing run if run is completed](https://github.com/Shopify/maintenance_tasks/pull/549/commits/e0e9287243d13470978de3e44483b37ba163857b):
- Handles races where we pause / cancel a run immediately after the status has been updated as `:succeeded` in the db (ie. user presses "cancel" right as the task finishes - the `@run` in the controller is not yet `succeeded`, so we will call `#cancel` on the run despite the fact that its status has since changed to `succeeded` in the db)
- The early return is necessary, because after we rescue / reload / retry to get the completed status, we'll want to return from the method rather than trying to perform an invalid status transition.

## Feedback
Does this make sense? Are there any flows I'm missing?
Do the tests make sense? Race conditions are pretty tricky to test. I opted to hijack Mocha's `#with` method and use a block to try and simulate some of the races, and then followed the tests we wrote to ensure `#running` handled races properly and used a callback on `CustomTaskJob` to execute a block. If anyone has better ideas for how to write these, I'm all ears.

### DIFF
--- a/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
+++ b/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
@@ -122,8 +122,7 @@ module MaintenanceTasks
     end
 
     def on_complete
-      @run.status = :succeeded
-      @run.ended_at = Time.now
+      @run.complete
     end
 
     # We are reopening a private part of Job Iteration's API here, so we should
--- a/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
+++ b/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
@@ -12,8 +12,8 @@ module MaintenanceTasks
       before_perform(:before_perform)
 
       on_start(:on_start)
-      on_complete(:on_complete)
       on_shutdown(:on_shutdown)
+      on_complete(:on_complete)
 
       after_perform(:after_perform)
 
@@ -115,23 +115,15 @@ module MaintenanceTasks
       @run.start(count)
     end
 
+    def on_shutdown
+      @run.job_shutdown
+      @run.cursor = cursor_position
+      @ticker.persist
+    end
+
     def on_complete
       @run.status = :succeeded
       @run.ended_at = Time.now
-    end
-
-    def on_shutdown
-      if @run.cancelling?
-        @run.status = :cancelled
-        @run.ended_at = Time.now
-      elsif @run.pausing?
-        @run.status = :paused
-      else
-        @run.status = :interrupted
-      end
-
-      @run.cursor = cursor_position
-      @ticker.persist
     end
 
     # We are reopening a private part of Job Iteration's API here, so we should
--- a/app/models/maintenance_tasks/run.rb
+++ b/app/models/maintenance_tasks/run.rb
@@ -205,18 +205,31 @@ module MaintenanceTasks
       seconds_to_finished.seconds
     end
 
-    # Mark a Run as running.
+    # Marks a Run as running.
     #
     # If the run is stopping already, it will not transition to running.
+    # Rescues and retries status transition if an ActiveRecord::StaleObjectError
+    # is encountered.
     def running
-      return if stopping?
-      updated = self.class.where(id: id).where.not(status: STOPPING_STATUSES)
-        .update_all(status: :running, updated_at: Time.now) > 0
-      if updated
-        self.status = :running
-        clear_attribute_changes([:status])
+      if locking_enabled?
+        begin
+          running! unless stopping?
+        rescue ActiveRecord::StaleObjectError
+          reload_status
+          retry
+        end
       else
-        reload_status
+        # Preserve swap-and-replace solution for data races until users
+        # run migration to upgrade to optimistic locking solution
+        return if stopping?
+        updated = self.class.where(id: id).where.not(status: STOPPING_STATUSES)
+          .update_all(status: :running, updated_at: Time.now) > 0
+        if updated
+          self.status = :running
+          clear_attribute_changes([:status])
+        else
+          reload_status
+        end
       end
     end
 
--- a/app/models/maintenance_tasks/run.rb
+++ b/app/models/maintenance_tasks/run.rb
@@ -249,6 +249,9 @@ module MaintenanceTasks
     def start(count)
       update!(started_at: Time.now, tick_total: count)
       run_task_callbacks(:start)
+    rescue ActiveRecord::StaleObjectError
+      reload_status
+      retry
     end
 
     def job_shutdown
--- a/app/models/maintenance_tasks/run.rb
+++ b/app/models/maintenance_tasks/run.rb
@@ -214,6 +214,17 @@ module MaintenanceTasks
       run_task_callbacks(:start)
     end
 
+    def job_shutdown
+      if cancelling?
+        self.status = :cancelled
+        self.ended_at = Time.now
+      elsif pausing?
+        self.status = :paused
+      else
+        self.status = :interrupted
+      end
+    end
+
     # Cancels a Run.
     #
     # If the Run is paused, it will transition directly to cancelled, since the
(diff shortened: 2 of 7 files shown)


app/jobs/concerns/maintenance_tasks/task_job_concern.rb
```diff
@@ -135,8 +135,20 @@ def on_shutdown
         @task.run_callbacks(:interrupt)
       end
 
+      db_lock_version = @run.class.where(id: @run.id).pluck(:lock_version).first
```
etiennebarrie, line 138: This doesn't work without the migration, which I think is a goal of this PR.

app/jobs/concerns/maintenance_tasks/task_job_concern.rb
```diff
@@ -135,8 +135,20 @@ def on_shutdown
...
+      if db_lock_version != @run.lock_version
+        raise ActiveRecord::StaleObjectError, @run
+      end
+
       @run.cursor = cursor_position
       @ticker.persist
+
+      # Refresh lock version here since we've written to the run by
+      # persisting progress
+      refresh_run_lock_version
```
etiennebarrie, line 148: Can we update the lock_version in memory in `Run#persist_progress` instead, allowing the next update to succeed?

app/jobs/concerns/maintenance_tasks/task_job_concern.rb
```diff
@@ -135,8 +135,20 @@ def on_shutdown
         @task.run_callbacks(:interrupt)
       end
 
+      db_lock_version = @run.class.where(id: @run.id).pluck(:lock_version).first
+      if db_lock_version != @run.lock_version
+        raise ActiveRecord::StaleObjectError, @run
+      end
```
etiennebarrie, line 141: I don't think we should read the lock version, because we're always going to end up with a potential race condition between that read and the write in `after_perform`.

app/jobs/concerns/maintenance_tasks/task_job_concern.rb
```diff
@@ -135,8 +135,20 @@ def on_shutdown
...
+
       @run.cursor = cursor_position
       @ticker.persist
+
+      # Refresh lock version here since we've written to the run by
+      # persisting progress
+      refresh_run_lock_version
+    rescue ActiveRecord::StaleObjectError
+      @run.reload_status
+      retry
```
etiennebarrie, line 151: This may end up calling the callbacks more than once.

dirceu (approved):
These changes make sense to me and I think it would be helpful for us to test them more thoroughly in production before cutting a release.

Thank you for the excellent PR description and commit history!

app/jobs/concerns/maintenance_tasks/task_job_concern.rb
```diff
@@ -157,10 +169,15 @@ def reenqueue_iteration_job(should_ignore: true)
...
 
     def after_perform
+      new_status = @run.status
       @run.save!
       if defined?(@reenqueue_iteration_job) && @reenqueue_iteration_job
         reenqueue_iteration_job(should_ignore: false)
       end
+    rescue ActiveRecord::StaleObjectError
+      @run.reload_status
+      @run.status = new_status
```
etiennebarrie, line 179: It feels a bit heavy-handed to force the new status here, is this to handle the running -> cancelling -> succeeded case which doesn't seem handled like the other ones in `on_complete`?
Like I mentioned I think there's still a race between on_shutdown and after_perform, so I think we should move the logic here. The hard part is that we want the transition logic from on_shutdown and on_complete to run, without the callbacks

etiennebarrie (commented):
Style note: we kept the bang methods from ActiveRecord::Enum intact, and defined methods without the trailing `!` (`running`, `cancel`).

We should do the same here instead of overriding the Enum methods: anyone familiar with Enum would get confused by `run.pausing!` not causing the run to have a status of `pausing` in the end.

nvasilevski (approved):
👍 
I think `job_shutdown` change is a good candidate for a separate PR  if you'll ever want to consider making this PR smaller
But if no changes required to this PR I think it's totally fine to ship it as part of this PR 👍

etiennebarrie (approved):

PR #552, opened 2021-12-17
Move callback handling to Run

etiennebarrie:
As the [job test in the first commit](https://github.com/Shopify/maintenance_tasks/commit/7aff74ee00eaa21956a9e517f66bcca9f1adea1f#diff-ee1840045a31fb5c53ac6edf68b90c38162d26d58cd8833148345505fd90e15d) shows, we're calling the `interrupt` callback just before the `complete` callback.

Mainly, we also want to move the callback handling to Run so that #549 is easier to implement, where we will be able to better handle persisting the transition and handling stale object errors, while also only triggering the correct callback.


adrianna-chang-shopify (approved):

dirceu (approved):

paarthmadan (approved):
LGTM, makes more sense!

PR #556, opened 2022-01-03
Ruby 3.1

etiennebarrie:
This enables CI for 3.1.0 and fixes what needed to be fixed.

I haven't set it as the default yet (`.ruby-version`) since there are some deprecations warnings for DidYouMean.

PR #565, opened 2022-01-30
Set default Ruby to 3.1.0

larouxn:
Simply setting the default development Ruby to 3.1.0 now that we support Ruby 3.1 as of https://github.com/Shopify/maintenance_tasks/pull/556.

Follow up to https://github.com/Shopify/maintenance_tasks/pull/550


adrianna-chang-shopify:
There are deprecation warnings for `DidYouMean::SPELL_CHECKERS` with Ruby 3.1 and a bundler version < 2.3.3. Looks like [upgrading to bundler 2.3.4](https://github.com/rubygems/rubygems/issues/5234#issuecomment-1002544751) should resolve this 😄

adrianna-chang-shopify (approved):

PR #566, opened 2022-01-30
Bump Rubocop target to Ruby 3.1

larouxn:
Noticed we still have the target Ruby for Rubocop set to 2.7. We can set this to 3.1 now that we support it, as of https://github.com/Shopify/maintenance_tasks/pull/556. The one gotcha here is if we don't add the following config we'll hit 22 offenses telling use to omit the values from the key value pairs, the new Ruby 3.1 feature (see bottom). If we _want_ to make these changes, we can omit the config change and autocorrect the offenses.

```yml
Style/HashSyntax:
  Enabled: false
```

Screen Shot 2022-01-30 at 13 14 18

*See "Other Notable New Features: Language" in the [release notes](https://www.ruby-lang.org/en/news/2021/12/25/ruby-3-1-0-released/).


larouxn:
> My preference would be to stick to using explicit hash literal values

This is also my preference, hence the config change, but I'm also open to making the changes if Rails peeps prefer.

casperisfine:
We can't allow that new hash syntax yet anyway as sorbet doesn't support it.

Issue #568, opened 2022-02-03
Support for collection-less one-off tasks


promulo:
Currently, tasks are supposed to execute over elements of a given collection. However, there is the use case of a task that doesn't really need a collection and just executes something once (e.g. triggering an async job, interacting with an external API). This behavior can be achieved in the gem by, for example, creating a collection with a single element which is just a workaround.

I am creating this issue in order to kick-off a discussion about supporting tasks that don't rely on collections "natively". Happy to collaborate on this if needed.

adrianna-chang-shopify:
Makes sense to me to support this. We've seen a number of folks in Core ask for this, and our suggestion to iterate over a collection of `[1]` is not the nicest API.

This should be fairly easy to build using Job Iteration's `build_once_enumerator`:
https://github.com/Shopify/job-iteration/blob/b459d6f696c04e9e93a40fba97ee5e12f3b2f579/lib/job-iteration/enumerator_builder.rb#L39-L41

The tricker piece is going to be the API from the task side. We could use a class-level method like we did for CSV tasks:

```ruby
  class ProcessOnceTask < MaintenanceTasks::Task
    no_collection   # could also be process_once, etc...

    def process
      # Do something
    end
  end
```

Or we could just allow users to omit a collection and interpret that as a task that only calls process once:
 ```ruby
  class ProcessOnceTask < MaintenanceTasks::Task
    def process
      # Do something
    end
  end
```
(I think here we risk being unclear if a user accidentally omits a collection).

cc @Shopify/rails

nvasilevski:
Makes sense for me as well 👍 

Personally I would prefer having a macro similar to the `csv_collection` one we have for csv tasks

PR #570, opened 2022-02-03
Support Tasks that perform a single iteration

adrianna-chang-shopify:
Closes: https://github.com/Shopify/maintenance_tasks/issues/568

Sometimes, a Task may simply need to perform an action without iterating over multiple elements in a collection. For example, a Task might need to kick off another job, interact with a third-party API, or perform a single operation. This adds support for "collection-less" tasks, which need only to define a `#process` method.

I'm proposing an API similar to what we've done for `csv_collection` tasks, where we use a `no_collection` macro in the task. Under the hood, we can use a `NoCollectionBuilder` to handle defining `#collection` and `#count` (with collection simply being `:no_collection` for use when building the enumerator), and in the job we can use Job Iteration's `#build_once_enumerator` to handle this.

I've added a generator option for this as a distinct boolean option. We could change the task generator to take a `type` option (ie. `--type=csv`, `--type=no-collection`) but this is technically an API change, and I'm not sure it's necessary to generalize this yet (maybe it will make more sense if we have _three_ options at some point in the future 😉 ). Instead, I've opted to raise if a user supplies multiple option flags.

## Feedback

Thoughts on the generator flag / macro name? Is `process_once` more suitable?

### DIFF
--- a/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
+++ b/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
@@ -36,6 +36,8 @@ module MaintenanceTasks
       @enumerator = nil
 
       collection_enum = case collection
+      when :no_collection
+        enumerator_builder.build_once_enumerator(cursor: nil)
       when ActiveRecord::Relation
         enumerator_builder.active_record_on_records(collection, cursor: cursor)
       when ActiveRecord::Batches::BatchEnumerator
--- a/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
+++ b/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
@@ -91,7 +91,11 @@ module MaintenanceTasks
     end
 
     def task_iteration(input)
-      @task.process(input)
+      if @task.no_collection?
+        @task.process
+      else
+        @task.process(input)
+      end
     rescue => error
       @errored_element = input
       raise error
--- a/lib/generators/maintenance_tasks/task_generator.rb
+++ b/lib/generators/maintenance_tasks/task_generator.rb
@@ -30,7 +30,7 @@ module MaintenanceTasks
       )
       if options[:csv]
         template("csv_task.rb", template_file)
-      elsif options[:no_collection]
+      elsif no_collection?
         template("no_collection_task.rb", template_file)
       else
         template("task.rb", template_file)
@@ -84,6 +84,10 @@ module MaintenanceTasks
 
     def test_framework
       Rails.application.config.generators.options[:rails][:test_framework]
+    end
+
+    def no_collection?
+      options[:no_collection]
     end
   end
 end
--- a/lib/generators/maintenance_tasks/task_generator.rb
+++ b/lib/generators/maintenance_tasks/task_generator.rb
@@ -12,10 +12,17 @@ module MaintenanceTasks
     class_option :csv, type: :boolean, default: false,
       desc: "Generate a CSV Task."
 
+    class_option :no_collection, type: :boolean, default: false,
+      desc: "Generate a collection-less Task."
+
     check_class_collision suffix: "Task"
 
     # Creates the Task file.
     def create_task_file
+      if options[:csv] && options[:no_collection]
+        raise "Multiple Task type options provided. Please use either "\
+          "--csv or --no-collection."
+      end
       template_file = File.join(
         "app/tasks/#{tasks_module_file_path}",
         class_path,
@@ -23,6 +30,8 @@ module MaintenanceTasks
       )
       if options[:csv]
         template("csv_task.rb", template_file)
+      elsif options[:no_collection]
+        template("no_collection_task.rb", template_file)
       else
         template("task.rb", template_file)
       end
--- a/README.md
+++ b/README.md
@@ -183,7 +183,7 @@ module Maintenance
   class NoCollectionTask < MaintenanceTasks::Task
     no_collection
 
-    def process(_)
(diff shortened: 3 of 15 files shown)


lib/generators/maintenance_tasks/task_generator.rb
```diff
@@ -12,17 +12,26 @@ class TaskGenerator < Rails::Generators::NamedBase
     class_option :csv, type: :boolean, default: false,
       desc: "Generate a CSV Task."
 
+    class_option :no_collection, type: :boolean, default: false,
```
gmcgibbon, line 15: To me, no_collection doesn't really imply a single run. Maybe `once`, `single_run` or `one_off` would make more sense?

app/jobs/concerns/maintenance_tasks/task_job_concern.rb
```diff
@@ -36,6 +36,8 @@ def build_enumerator(_run, cursor:)
       @enumerator = nil
 
       collection_enum = case collection
+      when :no_collection
+        enumerator_builder.build_once_enumerator(cursor: nil)
```
gmcgibbon, line 40: https://github.com/Shopify/job-iteration/blob/53767c94295af5655971c3b33397da692227519f/lib/job-iteration/enumerator_builder.rb#L40

I believe that means that `1` will be passed to the enumerator, if anyone else was curious.

gmcgibbon (approved):

app/jobs/concerns/maintenance_tasks/task_job_concern.rb
adrianna-chang-shopify: > I believe that means that 1 will be passed to the enumerator

Yes! I was also a bit intrigued when I saw this. Indeed, that means that `1` is passed to the user's `process` method, but I don't think it's a big deal either 😄

lib/generators/maintenance_tasks/task_generator.rb
gmcgibbon: I mean keeping
```rb
def collection(_task)
  :no_collection
end
```
and `when :no_collection` but renaming the class option, builder class, and class macro method to `single_run`. `process_once` or `one_off` seem fine to me as well, no strong opinion on which is best.

adrianna-chang-shopify:
> I wish we could solve the issue that process takes a parameter though.

Yeah agreed @etiennebarrie 😄 I thought about doing something like this:

```diff
diff --git a/app/jobs/concerns/maintenance_tasks/task_job_concern.rb b/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
index 59c09d0..f885984 100644
--- a/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
+++ b/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
@@ -91,7 +91,11 @@ module MaintenanceTasks
     end
 
     def task_iteration(input)
-      @task.process(input)
+      if @task.no_collection_task
+        @task.process
+      else
+        @task.process(input)
+      end
     rescue => error
       @errored_element = input
       raise error
diff --git a/app/models/maintenance_tasks/task.rb b/app/models/maintenance_tasks/task.rb
index ce33e83..384ac31 100644
--- a/app/models/maintenance_tasks/task.rb
+++ b/app/models/maintenance_tasks/task.rb
@@ -21,6 +21,8 @@ module MaintenanceTasks
     class_attribute :collection_builder_strategy,
       default: NullCollectionBuilder.new
 
+    class_attribute :no_collection_task, default: false
+
     define_callbacks :start, :complete, :error, :cancel, :pause, :interrupt
 
     class << self
@@ -68,6 +70,7 @@ module MaintenanceTasks
       def no_collection
         self.collection_builder_strategy =
           MaintenanceTasks::NoCollectionBuilder.new
+        self.no_collection_task = true
       end
 
       # Returns whether the Task handles CSV.
diff --git a/test/dummy/app/tasks/maintenance/no_collection_task.rb b/test/dummy/app/tasks/maintenance/no_collection_task.rb
index 67f2bb3..ef4dcbb 100644
--- a/test/dummy/app/tasks/maintenance/no_collection_task.rb
+++ b/test/dummy/app/tasks/maintenance/no_collection_task.rb
@@ -4,7 +4,7 @@ module Maintenance
   class NoCollectionTask < MaintenanceTasks::Task
     no_collection
 
-    def process(_)
+    def process
       Rails.logger.debug("#process method was called")
     end
   end
```

But wasn't sure if the added complexity was worth it given the generator gives you a template to follow. It's really not that much code though 😄  And I do wonder about the generated test suit:
```
class <%= class_name %>TaskTest < ActiveSupport::TestCase
  # test "#process performs a task iteration" do
  #   <%= tasks_module %>::<%= class_name %>Task.process(element)
  # end
end
```
I suppose this is potentially confusing? WDYT about going with the suggestion above ^ and providing a new template for collection-less tasks?

nvasilevski (approved):
I like the feature 👍 
Personally I feel it's reasonable that we are going with a macro approach to be consistent with `csv_collection`. 
Naming is hard so I don't have a strong preference, `no_collection` seems fine to me :) 
I agree that ideally we would want to have an argument-less `process` method, but I think it's not a blocker for the feature in general? I was thinking that perhaps we could have a separate PR for tweaking the `process` method to evaluate reasonability and possible approaches. 
It's just this particular PR looks nice and clean to me and it makes a decent UX improvement in terms of moving from:
```ruby
def collection
 [1]
end

def process(_)
end
```
to

```ruby
no_collection

def process(_)
end
```

PR #583, opened 2022-02-25
Don't re-enqueue stopped throttled jobs

etiennebarrie:
Currently throttling relies on the Enumerator enqueuing a job, which is problematic if we don't want to re-enqueue because the run is stopping (pausing or cancelling). This is shown in the new test which fails on main. The run is properly cancelled, but there's an enqueued job which then fails because it tries to move the run from cancelled to running.

This adds a conditional in the `after_perform` callback when we enqueue the job, to not do so when the run is stopped (at this point we've transitioned from _stopping_ to _stopped_.

This only works if we also change job-iteration to use the same `reenqueue_iteration_job` method for throttling, which is done in https://github.com/Shopify/job-iteration/pull/190.

If that PR gets merged, I'll release a new version and update this PR to require that fixed version.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -5,7 +5,7 @@ PATH
       actionpack (>= 6.0)
       activejob (>= 6.0)
       activerecord (>= 6.0)
-      job-iteration (~> 1.1)
+      job-iteration (~> 1.3.6)
       railties (>= 6.0)
 
 GEM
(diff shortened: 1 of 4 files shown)


adrianna-chang-shopify (approved):

adrianna-chang-shopify:
@etiennebarrie can we cut a new version of job-iteration and get this 🚢 ?

PR #586, opened 2022-02-28
fix: approximate elapsed time calculation helper

guemidiborhane:


test/helpers/maintenance_tasks/tasks_helper_test.rb
```diff
@@ -55,6 +55,7 @@ class TasksHelperTest < ActionView::TestCase
     end
 
     test "#time_running_in_words reports the approximate time running of the given Run" do
+      @run.started_at = Time.now
```
etiennebarrie, line 58: If I run this test without the app change, it passes, so this doesn't show the issue.

app/helpers/maintenance_tasks/tasks_helper.rb
```diff
@@ -70,7 +70,8 @@ def status_tag(status)
     #
     # @return [String] the description of the time running attribute.
     def time_running_in_words(run)
-      distance_of_time_in_words(0, run.time_running, include_seconds: true)
+      time_running = run.started_at + run.time_running.seconds
+      distance_of_time_in_words(run.started_at, time_running, include_seconds: true)
```
etiennebarrie, line 74: Ok so we're trying to get a real time here, we don't even need `run.started_at`, we could use `Time.now` as the start, but we're not seeing any issues with the way this is done with `Time.at(0)`.
I assume something must be happening with activesupport, which version are you using?

guemidiborhane: Hey sorry for the late response, the version we're using is `6.1.4.6` for both actionview and activesupport.

gmcgibbon:
Since it seems the like fault of a third party gem, I'm going to close this. Feel free to re-open it if you experience issues. Thank you for your contributions!

Issue #601, opened 2022-03-31
Batch CSV tasks


adrianna-chang-shopify:
Users may want to iterate over batches of data that aren't Active Record Relations. Job Iteration has a [`CsvEnumerator#batches`](https://github.com/Shopify/job-iteration/blob/main/lib/job-iteration/csv_enumerator.rb#L41-L47) API that we could leverage to support CSV batches. This should be pretty straightforward to implement. We could tweak the existing API:
```ruby
class BatchCsvTask < MaintenanceTasks::Task
  csv_collection(in_batches: 50)

  def process(batch_of_rows); end
end
```
with the default batch size being 1 to preserve existing behaviour, or we could expose a new macro:
```ruby
class BatchCsvTask < MaintenanceTasks::Task
  batch_csv_collection(batch_size: 50)

  def process(batch_of_rows); end
end
```

Any opinions on this @Shopify/rails ?

PR #604, opened 2022-04-04
Support processing CSVs in batches

adrianna-chang-shopify:
Closes: https://github.com/Shopify/maintenance_tasks/issues/601

This PR adds support for batch-processing CSV rows in a task by extending the existing `csv_collection` macro to take an `in_batches` option . Arrays of `CSV::Row` objects are then passed as input to the task's `#process` method.

Job Iteration already has [support for iterating over CSVs in batches](https://github.com/Shopify/job-iteration/blob/2e29ba910b7b97bfbe318a97e33918638dfd2d36/lib/job-iteration/csv_enumerator.rb#L39-L47), so our `#build_enumerator` method simply needs to hit job-iteration's `#batches` API to achieve this.

🎩 

- `bin/rails s`
- Navigate to http://127.0.0.1:3000/maintenance_tasks/tasks/Maintenance::BatchImportPostsTask
- Upload `test/fixtures/files/sample.csv`
- Run the task

### DIFF
--- a/app/models/maintenance_tasks/batch_csv_collection_builder.rb
+++ b/app/models/maintenance_tasks/batch_csv_collection_builder.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+require "csv"
+
+module MaintenanceTasks
+  # Strategy for building a Task that processes CSV files in batches.
+  #
+  # @api private
+  class BatchCsvCollectionBuilder
+    BatchCsv = Struct.new(:csv, :batch_size, keyword_init: true)
+
+    # Initialize a BatchCsvCollectionBuilder with a batch size.
+    #
+    # @param batch_size [Integer] the number of CSV rows in a batch.
+    def initialize(batch_size)
+      @batch_size = batch_size
+    end
+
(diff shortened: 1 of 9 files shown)


app/models/maintenance_tasks/batch_csv_collection_builder.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+require "csv"
+
+module MaintenanceTasks
+  # Strategy for building a Task that processes CSV files in batches.
+  #
+  # @api private
+  class BatchCsvCollectionBuilder
```
gmcgibbon, line 9: You _could_ inherit `CsvCollectionBuilder`, to repeat less code here.

gmcgibbon (approved):

app/models/maintenance_tasks/batch_csv_collection_builder.rb
adrianna-chang-shopify: Good point, thanks Gannon!

etiennebarrie (approved):

PR #607, opened 2022-04-07
CSS updates

etiennebarrie:


etiennebarrie:
No other reason than that's what was documented at https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity.

PR #610, opened 2022-04-12
Remove prefixed $ from shell command examples

larouxn:
As the prefixed `$` makes it impossible to copy and paste example shell commands, I was thinking it might be beneficial to remove them from the README. What do y'all think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,8 +9,8 @@ A Rails engine for queuing and managing maintenance tasks.
 To install the gem and run the install generator, execute:
 
 ```bash
-$ bundle add maintenance_tasks
-$ bin/rails generate maintenance_tasks:install
+bundle add maintenance_tasks
+bin/rails generate maintenance_tasks:install
 ```
 
 The generator creates and runs a migration to add the necessary table to your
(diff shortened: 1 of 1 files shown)


etiennebarrie:
I don't know because while it makes copying more easy with triple-click selection, it also makes it less clear that these are commands to type at the shell prompt… Also it's the default from bundler generated gems for instance, so I don't think people are too confused:
https://github.com/rubygems/rubygems/blob/085094b60caa7a9ac6771a7fc6b452fc79911bfb/bundler/lib/bundler/templates/newgem/README.md.tt#L11
https://github.com/rails/rails/#getting-started

larouxn:
I suppose it _is_ a bit less clear they are shell commands but I feel it's still somewhat clear given the code block formatting and usage of terms like `bin/rails`. As for Bundler's default, a bit brazen, but I feel Bundler should update their docs/generation as well then. Ever since GitHub added this copy button, having breaking `$` prefixed has become a bit of an issue.
Screen Shot 2022-04-13 at 7 30 28

re: triple click, this is the current experience. Can't do much with `$ gem install rails`. 🤷‍♂️ 
Screen Shot 2022-04-13 at 7 35 35

I don't think this change makes things more clear. Perhaps 20% less clear but 100% more convenient. 😅

etiennebarrie (approved):
Yes that's what I meant by triple-clicking, this PR makes it much more useful. The copy button is similar in that regard. Ok I'm convinced!

While we're at it, can we change all the `bash` into `sh-session`?

larouxn:
> While we're at it, can we change all the `bash` into `sh-session`?

Sure thing. Done!

Issue #569, opened 2022-02-03
Use Turbo stream instead of a meta tag to refresh page


elfassy:
The [current solution](https://github.com/Shopify/maintenance_tasks/blob/main/app/views/layouts/maintenance_tasks/application.html.erb#L35) for refreshing the page was good as a MVP, but it comes with some problems:
- hard to search / scroll as the page always moves back to the top (#510)
- hard to copy paste text or links

Suggestion:
Use rails' turbo https://turbo.hotwired.dev/handbook/streams

TopRoupi:
my suggestion would be just to add a refresh button in the nav bar and stop with the auto refresh every 3 seconds.

PR #612, opened 2022-04-13
Refresh by replacing document body instead of using header

adrianna-chang-shopify:
Co-authored-by: Étienne Barrié <etienne.barrie@shopify.com>

Closes: https://github.com/Shopify/maintenance_tasks/issues/510, https://github.com/Shopify/maintenance_tasks/issues/569

This is an alternative approach to page refreshing, that should solve some of the complaints folks had with the meta tag usage (losing search abilities, making it difficult to copy/paste text, etc.)

We don't want to jump straight to Turbo for a couple of reasons:
- We'd need to make it an explicit dependency / configure it properly for support in pre Rails 7 apps
-  The gem doesn't use any assets currently / doesn't force embedding apps to use the asset pipeline, so it's tricky make the JS component for Turbo work
- Turbo includes a lot of stuff we don't need to achieve the desired behaviour (page refreshing)

@etiennebarrie and I came up with a fairly simple solution to replace using the `http-equiv="refresh"` attribute for refreshing, but that sticks with plain JS. We essentially just swap out the document body in an inline script that uses `window.setInterval()` to run repeatedly. The script's SHA is included in the allowlist for the gem's CSP.  This preserves the ability to do things like Cmd+F to search for tasks.

### DIFF
--- a/app/controllers/maintenance_tasks/application_controller.rb
+++ b/app/controllers/maintenance_tasks/application_controller.rb
@@ -15,7 +15,7 @@ module MaintenanceTasks
       )
       policy.script_src(
         # page refresh script
-        "'sha256-2tgF+1+25h/mbN7WzbRy0lIuf3qoUBtqt/ipsHJXYfg='",
+        "'sha256-kOtH9RgO29UNB3GHLZpYttLMkK0yn5uWA9jDK1WLXlQ='",
       )
       policy.frame_ancestors(:self)
     end
--- a/app/controllers/maintenance_tasks/application_controller.rb
+++ b/app/controllers/maintenance_tasks/application_controller.rb
@@ -14,6 +14,10 @@ module MaintenanceTasks
         "'sha384-" \
           "MJ6MP7wjBI68eI1gLBSKU86HYHooeOnBdlmsV/RfzritEYu5Xaa1vROm+3dkbeZt'",
       )
+      policy.script_src(
+        # page refresh script
+        "'sha256-fADwJN++FN9LuhJ4XRCiIoJ7KPwmDYx2qiBgm20D8rA='",
+      )
       policy.frame_ancestors(:self)
     end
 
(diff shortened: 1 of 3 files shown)


nvasilevski (approved):

gmcgibbon (approved):

adrianna-chang-shopify:
@etiennebarrie I realized that one issue with this approach is that the body will continue to be refreshed after the task finishes because `setInterval` is never cleared. If the user manually refreshes or navigates away from the page it will stop, but if they stay on the page the function will continue to run. Do you think that's problematic?

etiennebarrie:
Some key points:

> losing search abilities

We're still reseting the browser search, but at least we're not scrolling up.

> making it difficult to copy/paste text

Your selection is also lost when the refresh occurs. I wonder what people copy though. Even with the current hard reload, if you've right clicked on a link, you can still copy it after the page reloaded.

I think this helps a lot simply by not scrolling around too much, but it's still not ideal. I wonder how hard would it be to only refresh the progress bars, and the description about the status, without touching the rest. I guess we'd also need to move the tasks at their right spot (they might have been completed or started).

---

I tested on Safari and noticed the style was broken (the stylesheet was never linked). We can't rely on the `Sec-Fetch-User` header we were using, so I sent the "Rails" one so that we can use `Request::xhr?`.

---

> the body will continue to be refreshed after the task finishes

I changed the approach a little bit: we're always including the script, but only refreshing if the body doesn't have a `data-refresh` attribute.

adrianna-chang-shopify:
Definitely still not perfect, but I agree that the scrolling was the main issue. The browser search resetting is not too bad if you've already navigated to your task on the index page, and losing your selection also matters less if the page doesn't move because you can just reselect. 

> I guess we'd also need to move the tasks at their right spot (they might have been completed or started).

I think this is the trickiest part. Even if we were to rely on something like Turbo, if tasks are moving from one section to another (for example new => active => completed), it would refresh the contents of those sections and you'd lose search /  text selection anyways.

Nice fix for the refreshing! 👌

PR #620, opened 2022-05-05
Drop support for classic autoloading

gmcgibbon:
In https://github.com/Shopify/maintenance_tasks/commit/c11faffc00218cb703e804444a83ced2d0e21cd8 we accidentally broke classic autoloading by referencing `CsvCollectionBuilder`. This made me remember that we should have removed support for this it several releases ago anyway. 

Yes, we could instead make it clearer that we need to reference `MaintenanceTasks::CsvCollectionBuilder` in a fully qualified way, but I think since classic autoloading is deprecated both in Rails and the gem, we can just remove it now.

Closes https://github.com/Shopify/maintenance_tasks/issues/528

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,13 @@ take a look at the [Active Job documentation][active-job-docs].
 [async-adapter]: https://api.rubyonrails.org/classes/ActiveJob/QueueAdapters/AsyncAdapter.html
 [active-job-docs]: https://guides.rubyonrails.org/active_job_basics.html#setting-the-backend
 
+
+### Autoloading
+
+The Maintenance Tasks framework does not support autoloading in `:classic` mode.
+Please ensure your application is using [Zeitwerk](https://github.com/fxn/zeitwerk) to load your code.
+For more information, please consult the [Rails guides on autoloading and reloading constants](https://guides.rubyonrails.org/autoloading_and_reloading_constants.html).
+
 ## Usage
 
 ### Creating a Task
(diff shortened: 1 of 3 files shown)


adrianna-chang-shopify (approved):
Thanks for following up on this Gannon 🙏  Seems fair to drop support at this point.

etiennebarrie:
Wouldn't this be a breaking change though? I'm not ready to ship a 2.0 over this…

adrianna-chang-shopify:
I do feel like we could get away with dropping support for a deprecated feature in a minor version release, but you're right that that's technically not good release behaviour. My concern is that we'll never feel ready to release a major version, and we'll have to keep reminding ourselves that these deprecations still exist and may continue to accidentally ship changes that are incompatible with the classic autoloader.

gmcgibbon:
If we want to ship this with some other breaking changes, I'm alright with that. Maybe we can use milestones to coordinate releases in that case? If we don't have anything blocking the release, I think we should just do it. This has been warned about since 1.7.0 in https://github.com/Shopify/maintenance_tasks/pull/531/commits/2fe9ab933f4027c088d140abd64b541f4d05743f.

adrianna-chang-shopify:
Using a milestone to prep for a 2.0 release sounds like a great idea to me @gmcgibbon 👍  There are a couple of other deprecations we could remove:
- The old `error_handler` API
https://github.com/Shopify/maintenance_tasks/blob/812d8e56beec57665c06f7b2a9abf9c4cb301241/lib/maintenance_tasks.rb#L76-L79
- `Runner.new`
https://github.com/Shopify/maintenance_tasks/blob/812d8e56beec57665c06f7b2a9abf9c4cb301241/app/models/maintenance_tasks/runner.rb#L9-L14

gmcgibbon:
I can work on the other deprecations today. If there's any issues or PRs open that we should add to the milestone, please do so.

adrianna-chang-shopify:
My vote would be to ship v1.10 with everything up to the deprecation removals, and then ship all the breaking changes together in 2.0.

PR #623, opened 2022-05-06
Drop support for error_handler lambdas with unsupported arities

### LINKED ISSUE #621: Remove error handler deprecation
https://github.com/Shopify/maintenance_tasks/blob/812d8e56beec57665c06f7b2a9abf9c4cb301241/lib/maintenance_tasks.rb#L76-L79

gmcgibbon:
Closes #621


adrianna-chang-shopify (approved):

lib/maintenance_tasks.rb
```diff
@@ -64,25 +64,7 @@ module MaintenanceTasks
   mattr_accessor :backtrace_cleaner
 
   # @private
```
adrianna-chang-shopify, line 66: I can't remember why this ended up with a `@private` tag. I think it should be `@!attribute` to match the others?

gmcgibbon: Below you can see the `@!attribute` docs, so I just moved the accessor to it. I assume the method override was marked as private because you didn't want repeated docs.

PR #625, opened 2022-05-06
Change the cursor while the page is refreshing

etiennebarrie:
Follow-up to #612 

This makes it a bit more visible that the page is being refreshed. Otherwise there's no way to know.


gmcgibbon (approved):
We could alternatively place one of these at the top of the page: https://bulma.io/documentation/elements/progress/#indeterminate, but this seems fine to me.

adrianna-chang-shopify (approved):
🎩  looks 👍

PR #628, opened 2022-05-09
Use helper for HTML interpolation

### LINKED ISSUE #627: Better HTML doesn't like the new refresh
Just saw the following error from Shopify's [better-html gem](https://github.com/Shopify/better-html) after deploying [v1.10.0](https://github.com/Shopify/maintenance_tasks/releases/tag/v1.10.0).

```ruby
ActionView::Template::Errormaintenance_tasks/tasks#index
Do not interpolate String in a tag. Instead of <body <%= "data-refresh" if defined?(@refresh) && @refresh %>> please try <body <%= html_attributes(attr: value) %>>.
```

Looks to be related to https://github.com/Shopify/maintenance_tasks/pull/612, specifically [this line](https://github.com/Shopify/maintenance_tasks/pull/612/files#diff-dd4a5002cedc1952acb99e96c8b3e23084db2a1fe6cb41682049142b1976bacaR53). Not sure if we want to do anything but I thought I'd bring it up.

etiennebarrie:
Fixes #627 

cc @larouxn

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,6 +4,7 @@ source "https://rubygems.org"
 
 gemspec
 
+gem "better_html"
 gem "capybara"
 gem "mocha"
 gem "net-smtp" # mail is missing a dependency on net-smtp https://github.com/mikel/mail/pull/1439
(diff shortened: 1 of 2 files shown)


promulo (approved):

etiennebarrie:
Rebased on top of an older version of main to be able to have a separate branch for v1.x.

PR #651, opened 2022-06-15
Fail on invalid CSV content type

### LINKED ISSUE #443: Fail early when incorrect file is used for csv tasks
We had a user attempt to use a `tsv` file instead of a `csv` for our maintenance task.

The task continued to attempt to process the file and fail.

We should fail earlier and indicate that the input is incorrect.

george-ma:
Co-authored-by: @promulo 

Closes #443 

## What are you trying to accomplish?
Fails early when the content type is not a "text/csv"
image

## What should reviewers focus on?
We are setting the CLI `content_type` to be "text/csv" because unlike on the controller, the CLI csv_file is a hash.

### Before you deploy
- [X] I [tophatted](https://development.shopify.io/engineering/developing_at_Shopify/write-code/tophatting) or tested this change. (tested on my own sandbox repo)
- [x] This PR is [safe to rollback](https://development.shopify.io/engineering/developersToolbox/ship/safe_to_merge#Signs_your_PR_is_safe_to_rollback).

### DIFF
--- a/app/models/maintenance_tasks/run.rb
+++ b/app/models/maintenance_tasks/run.rb
@@ -40,6 +40,7 @@ module MaintenanceTasks
       Task.available_tasks.map(&:to_s)
     } }
     validate :csv_attachment_presence, on: :create
+    validate :csv_content_type, on: :create
     validate :validate_task_arguments, on: :create
 
     attr_readonly :task_name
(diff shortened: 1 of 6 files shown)


app/models/maintenance_tasks/run.rb
```diff
@@ -342,6 +343,18 @@ def csv_attachment_presence
       nil
     end
 
+    # Performs validation on the content type of the :csv_file attachment.
+    # A Run for a Task that uses CsvCollection must have a present :csv_file
+    # and a content type of "text/csv" to be valid. The appropriate error is
+    # added if the Run does not meet the above criteria.
+    def csv_content_type
+      if csv_file.present? && csv_file.content_type != "text/csv"
+        errors.add(:csv_file, "must have a content_type of text/csv")
```
etiennebarrie, line 352:
```suggestion
        errors.add(:csv_file, "must be a CSV")
```

etiennebarrie (commented):
I think there's one more thing we could do, is to add the `accept` option to the `input[type=file]` so that the browser helps with filtering:
image

adrianna-chang-shopify:
Late to the party, but thanks for shipping this guys! ❤️

PR #654, opened 2022-06-17
Add typical workflow to README

sambostock:
Currently, it is easy to overlook the section mentioning running tasks via the web UI. By adding a typical workflow, and some section headings, we make this clearer.


sambostock:
I've addressed the punctuation comments. I'm going to merge leaving the database section for now to get the improved docs in; we can always iterate and remove it later if we so choose.

PR #667, opened 2022-07-11
Add support for multiple runs per task

promulo:
Currently, only one task run can be active (running or paused) at a given time. This PR intends to add support for multiple runs for a given task to the gem.

The approach here was to introduce as least disruption as possible and implement minimal changes to both backend and UI, still adhering to current patterns. The dashboard and individual task pages were tweaked a bit in order to allow displaying more than one run for the same task. This means we should rely less on the concept of _last run_ and more on _active runs_.

I haven't experienced issues when manually testing this. Multiple task runs can be triggered using the usual _Run_ button and individual runs can be paused and resumed accordingly.

This is how the "new" index page looks like with these changes (the list of "New Tasks" and "Completed Tasks" look the same):

image

The above screenshot shows a real example of two runs of the same task being active at the same time.

As for the individual task page, this is the current state:

image

That is an example of a task with multiple active runs listed on its own page.

I will keep pushing changes to this branch but wanted to have some early feedback. The proposed UI might not be what we want. I'm also considering splitting this in separate backend/frontend PRs if that favors iteration. 

Thanks!

### DIFF
--- a/app/models/maintenance_tasks/task_data.rb
+++ b/app/models/maintenance_tasks/task_data.rb
@@ -24,7 +24,8 @@ module MaintenanceTasks
       #   a Run.
       def find(name)
         task_data = new(name)
-        task_data.last_run || Task.named(name)
+        task_data.active_runs.load
+        task_data.has_any_run? || Task.named(name)
         task_data
       end
 
@@ -32,40 +33,55 @@ module MaintenanceTasks
       # available Tasks.
       #
       # Tasks are sorted by category, and within a category, by Task name.
-      # Determining a Task's category require its latest Run record.
-      # To optimize calls to the database, a single query is done to get the
-      # last Run for each Task, and Task Data instances are initialized with
-      # these last_run values.
+      # Determining a Task's category requires their latest Run records.
+      # Two queries are done to get the currently active and completed Run
+      # records, and Task Data instances are initialized with these last_run
+      # values.
       #
       # @return [Array<TaskData>] the list of Task Data.
       def available_tasks
-        task_names = Task.available_tasks.map(&:name)
-        available_task_runs = Run.where(task_name: task_names)
-        last_runs = Run.with_attached_csv.where(
-          id: available_task_runs.select("MAX(id) as id").group(:task_name)
-        )
+        tasks = []
 
+        task_names = Task.available_tasks.map(&:name)
+
+        active_runs = Run.with_attached_csv.active.where(task_name: task_names)
+        active_runs.each do |run|
+          tasks << TaskData.new(run.task_name, run)
+          task_names.delete(run.task_name)
+        end
+
+        completed_runs = Run.completed.where(task_name: task_names)
+        last_runs = Run.with_attached_csv.where(
+          id: completed_runs.select("MAX(id) as id").group(:task_name)
+        )
         task_names.map do |task_name|
           last_run = last_runs.find { |run| run.task_name == task_name }
-          TaskData.new(task_name, last_run)
-        end.sort_by!(&:name)
+          tasks << TaskData.new(task_name, last_run)
+        end
+
+        # We add an additional sorting key (status) to avoid possible
+        # inconsistencies across database adapters when a Task has
+        # multiple active Runs.
+        tasks.sort_by! { |task| [task.name, task.status] }
       end
     end
 
-    # Initializes a Task Data with a name and optionally a last_run.
+    # Initializes a Task Data with a name and optionally a related run.
     #
     # @param name [String] the name of the Task subclass.
-    # @param last_run [MaintenanceTasks::Run] optionally, a Run record to
+    # @param related_run [MaintenanceTasks::Run] optionally, a Run record to
     #   set for the Task.
-    def initialize(name, last_run = :none_passed)
+    def initialize(name, related_run = nil)
       @name = name
-      @last_run = last_run unless last_run == :none_passed
+      @related_run = related_run
     end
 
     # @return [String] the name of the Task.
     attr_reader :name
+    attr_reader :related_run
 
     alias_method :to_s, :name
+    alias_method :last_run, :related_run
 
     # The Task's source code.
     #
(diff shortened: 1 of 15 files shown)


test/dummy/db/schema.rb
```diff
@@ -53,18 +52,19 @@
...
+    t.index ["task_name", "status"], name: "index_maintenance_tasks_runs_on_task_name_and_status"
   end
 
   create_table "posts", force: :cascade do |t|
     t.string "title"
     t.string "content"
-    t.datetime "created_at", precision: 6, null: false
-    t.datetime "updated_at", precision: 6, null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
```
adrianna-chang-shopify, line 67: Same with these 😄

promulo:
Thanks a lot @adrianna-chang-shopify, those are all fair points. I'll update the PR soon.

adrianna-chang-shopify:
> After discussing with @etiennebarrie we thought it would be a good idea to already drop the label from the title in the #show page and rely only on individual runs' status labels.

Yep, I'm OK with this too! 👍

test/dummy/db/schema.rb
```diff
@@ -56,8 +55,9 @@
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
     t.text "arguments"
-    t.index ["task_name", "created_at"], name: "index_maintenance_tasks_runs_on_task_name_and_created_at"
     t.integer "lock_version", default: 0, null: false
+    t.index ["task_name", "created_at"], name: "index_maintenance_tasks_runs_on_task_name_and_created_at"
```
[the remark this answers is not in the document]
etiennebarrie: Oh great yeah sorry I saw the pagination query and missed the fact that `TaskData#runs` was setting the order. So we use it but with this PR it becomes no longer used. We should have a similar ordered one since we now order by status.

test/dummy/db/schema.rb
```diff
@@ -56,8 +55,9 @@
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
     t.text "arguments"
-    t.index ["task_name", "created_at"], name: "index_maintenance_tasks_runs_on_task_name_and_created_at"
     t.integer "lock_version", default: 0, null: false
+    t.index ["task_name", "created_at"], name: "index_maintenance_tasks_runs_on_task_name_and_created_at", order: { created_at: :desc }
```
adrianna-chang-shopify, line 59: @etiennebarrie you'd mentioned no longer needing this index from the changes in this PR, but I'm wondering why. Won't `TaskData#completed_runs` still use this index to show historical run data?

@promulo did you add the order back manually here, or were you able to rerun the migrations to bring this back?

db/migrate/20220713131925_add_index_on_task_name_and_status_to_runs.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddIndexOnTaskNameAndStatusToRuns < ActiveRecord::Migration[6.0]
+  def change
+    add_index(:maintenance_tasks_runs, [:task_name, :status],
```
adrianna-chang-shopify, line 5: I was under the impression using an order on index was only necessary for multi-column indices needing to sort in different directions (ie if we were doing `task_name asc, status desc`). I'm not sure adding the order here is doing anything? 

It's also worth noting MySQL doesn't actually support ordering indices until 8.0: https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-add_index
> Note: [MySQL](https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/MySQL.html) only supports index order from 8.0.1 onwards (earlier versions accepted the syntax but ignored it).

app/models/maintenance_tasks/task_data.rb
```diff
@@ -157,6 +181,11 @@ def new
       MaintenanceTasks::Task.named(name).new
     end
 
+    # @return [Boolean] whether the Task has any Run.
+    def has_run?
```
adrianna-chang-shopify, line 185: Let's make this private?

adrianna-chang-shopify (commented):
This looks pretty close to me, although there's still some confusion on my end about ordering the indexes.

I think @etiennebarrie had also mentioned getting rid of the concept of `last_run` on `TaskData`, now that we're storing active / completed runs?

app/models/maintenance_tasks/task_data.rb
promulo: This method is called by `.find` on line 31, so making it private would not work. Unless you're suggesting we do `send(:has_run?)` there?

test/dummy/db/schema.rb
promulo: @adrianna-chang-shopify The order was added by the [first migration](https://github.com/Shopify/maintenance_tasks/blob/main/db/migrate/20201211151756_create_maintenance_tasks_runs.rb#L20) (after `rails db:reset db:migrate`). It was removed manually by https://github.com/Shopify/maintenance_tasks/commit/8cbe81099630c47c114b9e6e0dc05b7fa33fb4dd but there's no migration changing this index.

app/models/maintenance_tasks/task_data.rb
```diff
@@ -93,17 +103,23 @@ def last_run
       @last_run = runs.first
```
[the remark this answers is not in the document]
promulo: I've removed `TaskData#last_run` as a method but kept an alias for the old references. The instance variable is now called `related_run`. Updating all references can be done in a follow-up PR.

app/models/maintenance_tasks/task_data.rb
```diff
@@ -24,48 +24,64 @@ class << self
       #   a Run.
       def find(name)
         task_data = new(name)
-        task_data.last_run || Task.named(name)
+
+        task_data.active_runs.load
+
+        task_data.has_run? || Task.named(name)
+
```
etiennebarrie, line 31: We don't need all these new lines.

app/models/maintenance_tasks/task_data.rb
```diff
@@ -24,48 +24,64 @@ class << self
...
         )
-
         task_names.map do |task_name|
           last_run = last_runs.find { |run| run.task_name == task_name }
-          TaskData.new(task_name, last_run)
-        end.sort_by!(&:name)
+          tasks << TaskData.new(task_name, last_run)
+        end
+
+        tasks.sort_by! { |task| [task.name, task.status] }
```
etiennebarrie, line 65: Maybe add a comment that this is intended to ensure stable results across database adapters.

app/models/maintenance_tasks/task_data.rb
```diff
@@ -83,27 +99,23 @@ def code
...
+    # This collection represents a historic of past Runs for information
+    # purposes, since the base for Task Data information comes
+    # primarily from currently active runs.
     #
     # @return [ActiveRecord::Relation<MaintenanceTasks::Run>] the relation of
-    #   record previous to the last Run.
-    def previous_runs
-      return Run.none unless last_run
-
-      runs.where.not(id: last_run.id)
```
etiennebarrie, line 106: We used to take the current active on one side, and all the others on the other, now we take active + completed. It means there's some kind of coupling between those two. It may be worth having some kind of test that ensures all valid statuses fall in either one of these buckets. Maybe just `(Run::ACTIVE_STATUSES + Run::COMPLETED_STATUSES).sort == Run::STATUSES.sort` is enough.

app/models/maintenance_tasks/task_data.rb
```diff
@@ -114,8 +126,9 @@ def deleted?
       true
     end
 
-    # The Task status. It returns the status of the last Run, if present. If the
-    # Task does not have any Runs, the Task status is `new`.
+    # The Task status. It returns the status of the latest active or
```
etiennebarrie, line 129:
```suggestion
    # Returns the status of the latest active or
```

app/models/maintenance_tasks/task_data.rb
```diff
@@ -157,6 +170,11 @@ def new
       MaintenanceTasks::Task.named(name).new
     end
 
+    # @return [Boolean] whether the Task has any Run.
```
etiennebarrie, line 173: Maybe we could call it `ran?` or `has_any_run?`. We can also make this `@api private` because it shouldn't really be used beyond `find` ?

adrianna-chang-shopify (approved):
LGTM, great work on this! I chatted with Étienne about some changes that will be made in a follow up PR, but I think this is good to go in the meantime :shipit:

etiennebarrie (approved):
🎉

PR #672, opened 2022-07-20
Handle job still existing for a cancelled run

### LINKED ISSUE #616: Validation failed: Status Cannot transition run from status cancelled to errored
Example error: https://app.bugsnag.com/shopify/banking/errors/6260400a95d9870009401416?event_id=6260400a0093682418dc0000&i=sk&m=nw

```
gems/maintenance_tasks-1.9.0models/maintenance_tasks/run.rb:133
```

Bug seems to come from
https://github.com/Shopify/maintenance_tasks/blob/main/app/validators/maintenance_tasks/run_status_validator.rb#L10

etiennebarrie:
If a Run is force-cancelled but the job still exists in the queue, when it gets worked off, it causes an exception because of the state changes we expect.

The Run moves from cancelled to running, which causes an exception which is rescued and the Run is moved to errored but that causes another exception moving from cancelled to errored.

We now consider a cancelled Run as if it is cancelling, and finish the job early without processing any items.

Fixes #616

### DIFF
--- a/test/models/maintenance_tasks/run_test.rb
+++ b/test/models/maintenance_tasks/run_test.rb
@@ -222,18 +222,22 @@ module MaintenanceTasks
       assert_equal 2, query_count
     end
 
-    test "#stopping? returns true if status is pausing or cancelling" do
+    test "#stopping? returns true if status is pausing or cancelling or cancelled" do
       run = Run.new(task_name: "Maintenance::UpdatePostsTask")
 
-      (Run.statuses.keys - ["pausing", "cancelling"]).each do |status|
-        run.status = status
-        refute_predicate run, :stopping?
-      end
+      (Run.statuses.keys - ["pausing", "cancelling", "cancelled"])
+        .each do |status|
+          run.status = status
+          refute_predicate run, :stopping?
+        end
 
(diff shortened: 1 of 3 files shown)


test/models/maintenance_tasks/run_test.rb
```diff
@@ -222,19 +222,23 @@ class RunTest < ActiveSupport::TestCase
...
 
-    test "#stopping? returns true if status is pausing or cancelling" do
+    test "#stopping? returns true if status is pausing or cancelling or cancelled" do
       run = Run.new(task_name: "Maintenance::UpdatePostsTask")
 
-      (Run.statuses.keys - ["pausing", "cancelling"]).each do |status|
-        run.status = status
-        refute_predicate run, :stopping?
-      end
+      (Run.statuses.keys - ["pausing", "cancelling", "cancelled"])
```
promulo, line 228: nit: maybe this is more future-proof? 
```suggestion
      (Run.statuses.keys - Run::STOPPING_STATUSES)
```

promulo (approved):
this looks fine but I wonder if it makes sense to introduce a new `force_cancelled` status?

Issue #687, opened 2022-08-25
Don't rescue errors raised in callbacks


adrianna-chang-shopify:
The documentation currently dictates that errors raised in any callback other than `after_error` will bubble up to be handled by the error handled, and will halt task execution: https://github.com/Shopify/maintenance_tasks#using-task-callbacks

This is not how things work though -- when refactoring the callbacks code in https://github.com/Shopify/maintenance_tasks/commit/7aff74ee00eaa21956a9e517f66bcca9f1adea1f , we started rescuing any errors raised inside any of the task callbacks. There may have been a reason we did this at the time, but it's not the desired behaviour today, nor what is documented in the README.

We shouldn't be rescuing any errors outside of the `after_error` callback. Users can choose to rescue callback errors themselves, but in all other cases the error should bubble up to the handler and stop the task. This is especially important because users might be explicitly using callbacks to dictate task flow / halt execution.

PR #689, opened 2022-09-01
Don't rescue callback errors

adrianna-chang-shopify:
Closes: https://github.com/Shopify/maintenance_tasks/issues/687

Callback errors should be allowed to bubble up, halting task execution
and triggering the error handler. We should only rescue errors raised
in the `after_error` callback, since these are not handled any other way.

We _do_ need to rescue `Task::NotFound` errors though, in case a Task is deleted in the middle of running.

### DIFF
--- a/app/models/maintenance_tasks/run.rb
+++ b/app/models/maintenance_tasks/run.rb
@@ -140,7 +140,7 @@ module MaintenanceTasks
         backtrace: MaintenanceTasks.backtrace_cleaner.clean(error.backtrace),
         ended_at: Time.now,
       )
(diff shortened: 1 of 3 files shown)


nvasilevski (approved):

etiennebarrie:
Do you think we should push that to a v1 version? It is a breaking change of sorts, but it's also a bug, we meant not to rescue them. If we only ship it for a v2, then that answers the question, but we're not that close to shipping a v2. On the other hand, it's easy enough for a user to add the rescue we expected them to have in the first place.

Issue #675, opened 2022-07-28
Rename `TaskData#last_run`


etiennebarrie:
We added an alias in #667 but we should change that name since the concept is no longer "last run" and "previous run" but instead "active runs" and "completed runs" but on the index each active run is the "related run" of the task data. Maybe we want two classes for these two roles of "Task Data".

PR #696, opened 2022-09-13
Split `TaskData` into two separate classes

adrianna-chang-shopify:
Closes: https://github.com/Shopify/maintenance_tasks/issues/675

`TaskData` has become a combination of two separate interfaces: one for offering high-level representations of a task on the index page, and one for offering more details info on a specific task on the show page. This PR separates the two roles out into different classes, clarifying the responsibility of each object.

I've also dropped the `last_run` alias from `TaskDataIndex`, since these tasks have "related runs" instead of "last runs" with the changes introduced in https://github.com/Shopify/maintenance_tasks/pull/667.

Open to feedback on naming, I couldn't think of a good way to distinguish the two classes other than the context in which they're used 😅


nvasilevski (approved):
👍

adrianna-chang-shopify:
Hey @paarthmadan ! Thanks for bringing this up ❤️ To me, the current code is violating the _Interface Segregation Principle_:
> In the field of [software engineering](https://en.wikipedia.org/wiki/Software_engineering), the interface segregation principle (ISP) states that no code should be forced to depend on [methods](https://en.wikipedia.org/wiki/Method_(computer_programming)) it does not use.[[1]](https://en.wikipedia.org/wiki/Interface_segregation_principle#cite_note-ASD-1) ISP splits [interfaces](https://en.wikipedia.org/wiki/Interface_(computing)) that are very large into smaller and more specific ones so that clients will only have to know about the methods that are of interest to them.

IMHO it's less about `TaskData` being overloaded with too many responsibilities, and more about the fact that one set of responsibilities is used in one context (displaying tasks for the index page), and the other set is used in an entirely separate context (displaying tasks on the show page). Part of the reason the original issue was opened was because `TaskData` objects on the show page have _no related run_ -- so it seems odd for those `TaskData` instances to have a `related_run` / `last_run` attribute that is really has no intent to use.

Your point on separating querying vs presentation responsibilities is an interesting one! We could split out the querying / aggregating behaviour from the actual object that represents a `Task` in the UI, but to me it would still be fundamental to have `TaskPresenter#index` and `TaskPresenter#show` return different objects. And I'm not sure it buys us much at that point to have these two methods pulled out into a separate presenter object, rather than having each of the query methods as a class method in the respective `TaskData` class.

Does this make sense?

paarthmadan (approved):
Thanks for engaging in the discussion! ❤️ 

> ...but to me it would still be fundamental to have TaskPresenter#index and TaskPresenter#show return different objects.

👍 Your rationale makes sense given that it's important to return different objects.

Issue #644, opened 2022-06-02
Implicit `count` implementation


etiennebarrie:
We've moved `main` to prepare a future v2.0.0, mainly by removing deprecated code. One thing we hadn't tackled in order to preserve compatibility is implicit `count` method for tasks.

If you don't define a `count` method in your task, we show indeterminate progress in the general case, with exceptions:
- CSV and CSV batches
- Active Record batches

We could extend that, and prevent a lot of implementations of `count` which are just `collection.size`. Also that would avoid calling `collection` twice in a lot of cases (once in `build_enumerator` and once in `count`).

PR #698, opened 2022-09-16
Implicitly calculate collection size for Active Record Relations and Arrays

adrianna-chang-shopify:
Closes: https://github.com/Shopify/maintenance_tasks/issues/644

For v2.0, we've decided to start implicitly calculating a task's `tick_total` using the enumerator's size, rather than requiring users to define a `#count` method. `tick_total` is automatically calculated for several types of tasks -- CSV tasks, batch tasks, and batch CSV tasks, notably. CSV tasks and batch CSV tasks have their `#count` method overridden via a collection builder pattern, and batch tasks relies on the size of the enumerator constructed in `TaskJob#build_enumerator` to set the tick total.

For Arrays and Active Record Relations, we can also use the enumerator size to set the `#tick_total`. I modified the code in `TaskJobConcern` to always fall back to the size of the `@collection_enum` when setting the `tick_total`, if calling `@task.count` returns `:no_count` (our null collection default, which indicates that count hasn't been overridden).

I've updated the docs to drop references to `#count` where it's not necessary, and to indicate that this is only really necessary to no-op the operation. Note that this will be a **breaking change for some apps**, since leaving out the `#count` method previously meant not attempting to calculate the tick total. Users will need to explicitly add `def count; nil; end` to tasks that should not attempt to calculate the collection size.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,8 +69,12 @@ The generated task is a subclass of `MaintenanceTasks::Task` that implements:
 * `collection`: return an Active Record Relation or an Array to be iterated
   over.
 * `process`: do the work of your maintenance task on a single record
-* `count`: return the number of rows that will be iterated over (optional, to be
-  able to show progress)
+
+Optionally, tasks can also implement a custom `#count` method, defining the number
+of elements that will be iterated over. Your task's `tick_total` will be calculated
+automatically based on the collection size, but this value may be overriden if desired
+using the `#count` method (this might be done, for example, to avoid the query that would
+be produced to determine the size of your collection).
 
 Example:
 
@@ -81,10 +85,6 @@ module Maintenance
   class UpdatePostsTask < MaintenanceTasks::Task
     def collection
       Post.all
-    end
-
-    def count
-      collection.count
     end
 
     def process(post)
@@ -257,10 +257,6 @@ module Maintenance
       Post.all
     end
 
-    def count
-      collection.count
-    end
-
     def process(post)
       post.update!(content: "New content added on #{Time.now.utc}")
     end
@@ -307,10 +303,6 @@ module Maintenance
 
     def collection
       Post.all
-    end
-
-    def count
-      collection.count
     end
 
     def process(post)
--- a/test/dummy/app/tasks/maintenance/callback_test_task.rb
+++ b/test/dummy/app/tasks/maintenance/callback_test_task.rb
@@ -13,10 +13,6 @@ module Maintenance
       [1, 2]
     end
 
-    def count
-      collection.count
-    end
-
     def process(number)
       Rails.logger.debug("number: #{number}")
     end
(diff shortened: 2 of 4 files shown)


test/jobs/maintenance_tasks/task_job_test.rb
```diff
@@ -596,5 +595,27 @@ class << self
...
+    end
+
+    test "array-based tasks have their count calculated implicitly" do
+      Maintenance::TestTask.any_instance.expects(:process).once.with do
+        @run.cancelling!
+      end
+
+      TaskJob.perform_now(@run)
+
+      assert_equal 2, @run.reload.tick_total
```
nvasilevski, line 618: I'm not sure yet what my preference is but I wanted to start a quick discussion to see whether we should consider using `Post.count` as an expectation instead of a hard-coded number.
Generally I prefer explicit and hard-coded expectations as it makes easier to read and make assumptions about the test. But in this case it would be unfortunate for this test to fail just because someone added a new fixture record for an unrelated reason. It's not a big problem to bump this number every time we add a new fixture as our codebase is small and need for a new fixture record is rare but generally for larger codebase such unrelated failures cause unnecessary friction and slow down feature development.

Perhaps the best option for us would be to keep the explicit expectation of `2` ticks, but also make sure that the task always iterates over collection of `2` by either stubbing the `collection` method or maybe adding an argument to the task that will contain a comma-separated list of ids so we can pass something like `post_ids: [post(:one), post(:two)].map(&:id).join(",")` from the test to make sure the context in this test never changes

nvasilevski (approved):
I have a tiny non-blocking nitpick about the test expectation but the code change itself looks good 👍

test/jobs/maintenance_tasks/task_job_test.rb
adrianna-chang-shopify: Hey Nikita! This is an interesting point of discussion. I also usually prefer hardcoded expectations, although in this case I'm not sure changing `2` => `Post.count` really violates that 🤔 In general, I feel like when there are implementation-specific values in tests, it becomes a bit of a "smell" because the test implementation is too coupled to the code implementation (so if the code changes, the test will be forced to as well). But in this case, the fact that `Post.count` is the expected `tick_total` value is not tied to implementation, but rather an explicit _public API_ -- unless something changes around how we calculate ticks for Active Record Relations, this won't change. 

Consequently, I'd be fine to change the assertions to check `Post.count` as the expected value. I'd prefer not to have to use stubs here when we have the setup already with the fixtures -- IMHO stubs just make the test less clear. I'd also be fine just leaving as-is -- as you mentioned, we have very few tests that assert on this count, so I think hardcoding the count value is a fair tradeoff. For a larger codebase we'd maybe want to reconsider, but I think the tradeoffs should be weighed within the scope of the codebase we're working in 😄 

All that to say, I'll leave as-is for this PR, but if you feel strongly and would like to change in a follow-up, feel free to do so!

PR #701, opened 2022-09-23
Release 2.0

adrianna-chang-shopify:
With our all of our 2.0 milestones tackled, I think we're ready to release v2.0 for the gem!

I've drafted up a release [here](https://github.com/Shopify/maintenance_tasks/releases/edit/untagged-ace2100246d37ac480bf) that could use feedback.

Let me know if there's anything else we should address before performing a major version bump.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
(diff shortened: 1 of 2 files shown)


Schwad (approved):
🚀

PR #709, opened 2022-10-03
Prevent double loading of net-protocol in Ruby 2.7

etiennebarrie:
Ruby 2.7 stdlib's net-http, loaded by capybara, loads net/protocol from the stdlib ([with require_relative](https://github.com/ruby/ruby/blob/v2_7_6/lib/net/http.rb#L23)), and then we get [warnings for re-defining constants](https://github.com/Shopify/maintenance_tasks/actions/runs/3174329071/jobs/5170982966#step:7:8) when loading the gem version from net-smtp.

This fixes it by also using net-http as a gem, which since 3.0 loads it correctly with `require "net/protocol"`

Resolves #707 

It was also necessary to update rubygems for Ruby 2.7, otherwise there was an issue of stdlib's uri being loaded before the gem.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,7 @@ jobs:
       with:
         ruby-version: ${{ matrix.ruby }}
         bundler-cache: true
+        rubygems: ${{ matrix.ruby == '2.7' && 'latest' || 'default' }}
     - name: Rails version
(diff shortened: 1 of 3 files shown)


etiennebarrie:
The problem with this is that now in addition to Rails edge, all ruby 2.7 jobs may fail randomly when a gem gets updated…

PR #711, opened 2022-10-11
Add documentation option for tasks

packrat386:
I wanted to request the ability to add documentation for a task that would be displayed in the UI and I figured a Pull Request was the best way to get the conversation started on that. 

### The Problem

Imagine I had the following maintenance task:

```ruby
module Maintenance
  class DeletePostsTask < MaintenanceTasks::Task
    attribute :title_match, :string

    def collection
      Post.all
    end

    def process(post)
      return unless /#{title_match}/.match? post.title
      post.destroy
    end
  end
end
```

This can be conveniently used from the UI to delete posts. However I would like to notify the user of some things before they hit `Run`, notably that `title_match` is a regex and that this deletion is permanent. This example is somewhat contrived, but as it stands there's not a great way for me to warn the user about those things. I can put a comment in the task itself which they'll see if they read the displayed code, but I don't really expect people to do that.

### Proposal

I'd like to add something like a `documentation` field:

```ruby
module Maintenance
  class DeletePostsTask < MaintenanceTasks::Task
    documentation <<~TEXT
      Deletes posts whose titles match title_match as a regular expression.
      WARNING: deletion is PERMANENT.
    TEXT

    attribute :title_match, :string

    def collection
      Post.all
    end

    def process(post)
      return unless /#{title_match}/.match?(post.title)

      post.destroy
    end
  end
end
```

This will be displayed on the task's page like so:

image

### Caveats

1: The current implementation doesn't support styling at all. You basically just get one paragraph of plain text. Could possibly support injecting markdown or something but that seems like overkill.

2: I'm not a designer so this might be ugly.

3: It'd be nice to add the ability to document individual attributes as well, but in order to do something like 

```ruby
module Maintenance
  class DeletePostsTask < MaintenanceTasks::Task
    attribute :title_match, :string, documentation: "a regular expression matching post titles to delete"
    # ...
  end
end
```

One would need to override the definition of `.attribute` provided by `ActiveModel::Attributes` which seems hairy even if it's just `super` plus parsing out some new options.

An alternative would be to have a separate method call like:

```ruby
module Maintenance
  class DeletePostsTask < MaintenanceTasks::Task
    attribute :title_match, :string
    attribute_documentation :title_match, "a regular expression matching post titles to delete"
    # ...
  end
end
```

but that feels a bit awkward (should it error if the attribute isn't present?)

---

I tried my best to follow what seem to be the conventions of the repo, but this is my first time touching it. I'm happy to work with someone on making this adhere to the standards of the project, and I don't mind changing the details of the implementation if needed.

### DIFF
--- a/app/models/maintenance_tasks/task.rb
+++ b/app/models/maintenance_tasks/task.rb
@@ -21,6 +21,9 @@ module MaintenanceTasks
     # @api private
     class_attribute :collection_builder_strategy, default: NullCollectionBuilder.new
 
+    # @api private
+    class_attribute :task_documentation
+
     define_callbacks :start, :complete, :error, :cancel, :pause, :interrupt
 
     class << self
@@ -167,6 +170,13 @@ module MaintenanceTasks
       #   (see https://api.rubyonrails.org/classes/ActiveSupport/Callbacks/ClassMethods.html#method-i-set_callback)
       def after_error(*filter_list, &block)
         set_callback(:error, :after, *filter_list, &block)
+      end
+
+      # Set the documentation string for the task
+      #
+      # @param doc [String] the documentation for the task
+      def documentation(doc)
+        self.task_documentation = doc
       end
 
       private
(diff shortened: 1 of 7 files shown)


packrat386:
Also I am happy to sign the CLA but I don't know how 😖

PR #755, opened 2023-01-24
Remove `pg` gem as a development requirement

nvasilevski:
PosgreSQL is not an actual development requirement for this gem and asking everyone to install it only discourages contributions and make it less approachable for cloud-based environments like codespaces.
For many parts of the maintenance tasks even an in-memory `sqlite` database will suffice. 

So this PR removes the `pg` gem from the `Gemfile` and moves it to a separate gemfile which will be used on CI only.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
         ruby: ["2.7", "3.0", "3.1"]
         database: [sqlite]
         include:
-          - gemfile: Gemfile
+          - gemfile: "gemfiles/postgresql.gemfile"
             ruby: 3.1
             database: postgres
     env:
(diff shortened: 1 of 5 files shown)


gmcgibbon (approved):
I think the expected check can be removed from the required in the default branch settings.

nvasilevski:
>I think the expected check can be removed from the required checks in the default branch settings.

Right ❤️ Thank you! I'm going to bring the `postgres` steps as required after merging this PR as those don't seem to show up until merged

.github/workflows/ci.yml
```diff
@@ -13,7 +13,7 @@ jobs:
         ruby: ["2.7", "3.0", "3.1"]
         database: [sqlite]
         include:
-          - gemfile: Gemfile
+          - gemfile: "gemfiles/postgresql.gemfile"
```
Schwad, line 16: Ah, that's a cool way to isolate pg for just CI. I like it.

adrianna-chang-shopify:
Nice 👍

PR #756, opened 2023-01-24
Add Rails 7 and Ruby 3.2 to ci matrix

nvasilevski:
Additions to the CI:

`Ruby 3.2` - Just a new ruby version

`Rails 7` - Every day Rails 7 differs from rails edge more and more and now it's actually closer to Rails 7.1 so I would like to test specifically against Rails 7.

❓ 
- Ruby 2.7 has end of life in a few month. I don't think it means we should drop the support right now but we might want to do it soon

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -40,6 +40,11 @@ require "csv"
 module Warning
   class << self
     def warn(message)
+      # gems/selenium-webdriver-4.7.1/lib/selenium/webdriver/remote/bridge.rb:633:
+      # warning: Expected selenium/webdriver/remote/commands
+      # to define Selenium::WebDriver::Remote::COMMANDS but it didn't
+      return if message.match?(/Selenium::WebDriver::Remote::COMMANDS/)
+
       raise message.to_s
     end
   end
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,8 +9,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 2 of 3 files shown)


test/test_helper.rb
```diff
@@ -40,6 +40,11 @@
 module Warning
   class << self
     def warn(message)
+      # gems/selenium-webdriver-4.7.1/lib/selenium/webdriver/remote/bridge.rb:633:
```
nvasilevski, line 43: the warning is not under our control and it's not causing test failures so I decided to ignore it until we or someone else addresses it upstream

gmcgibbon: It might be nice to raise an error if it doesn't happen anymore so we know when we can remove this code. I assume it will be fixed eventually.

gmcgibbon (approved):
I think we can remove 2.7 support when we have a valid reason to. For now, it can stay.

adrianna-chang-shopify (approved):

test/test_helper.rb
nvasilevski: I spend some time trying to come up with a quick solution for this because I do agree that it's always great to know when things become unnecessary in an ignore list but I couldn't come up with a solution I would like as it was a little messy to try to keep track of warning still happening while only performing the check for ruby >= 3.2
If you don't mind, I'll merge this PR and try to address the warning in the gem itself to avoid spending too much time on this. If something goes wrong with the upstream fix I'll open a separate PR with a way to track warning status

etiennebarrie:
I found that we can `require "selenium/webdriver/remote/commands"` in `test/application_system_test_case.rb` to remove the warning. But there are more issues with Ruby 3.2 under macOS. PR incoming.

PR #759, opened 2023-01-26
Add basic spin bootstrap

nvasilevski:
Currently spin automatically infers `rails` from the gemfile and installs `mysql` along with setting the `DATABASE_URL` which breaks the env as we are using `sqlite` for development.

Setting `.spin/svc.yml` to an empty file overrides the automatically inferred setup and basically means "we do not need anything". This is enough to boot an instance, run `bundle` and be able to run `bin/rails test` unit tests. This is not enough to run system integration tests but we will tackle it separately once we have such need.


nvasilevski:
>Shouldn't we focus on things like https://github.com/Shopify/maintenance_tasks/pull/544

Actually I agree

Closed in favour of https://github.com/Shopify/maintenance_tasks/pull/760

PR #760, opened 2023-01-27
Stop spin from taking snapshots of the repo

nvasilevski:
Substitutes - https://github.com/Shopify/maintenance_tasks/pull/759

Since the effort to fully onboard the repo onto Spin right now is not justified let's opt-out the repo from snapshots to reduce internal noise.

The repo will still be bootable on Spin for basic contributions which may already be covering most of the contribution needs. 
As @etiennebarrie pointed out, it would be more beneficial to invest into publicly available and publicly usable environment setup, for example https://github.com/Shopify/maintenance_tasks/pull/544


Schwad:
Okay cool. So essentially I'll keep this as a tool in my back pocket if snapshots are falling over etc. :)

PR #764, opened 2023-01-31
Fix `Style/InvertibleUnlessCondition`

etiennebarrie:
This is enabled in rubocop-shopify v2.12.0.

https://github.com/Shopify/ruby-style-guide/releases/tag/v2.12.0

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -204,8 +204,8 @@ GEM
       unicode-display_width (>= 2.4.0, < 3.0)
     rubocop-ast (1.24.1)
       parser (>= 3.1.1.0)
-    rubocop-shopify (2.11.1)
-      rubocop (~> 1.42)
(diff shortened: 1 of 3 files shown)


Gemfile.lock
```diff
@@ -272,4 +272,4 @@ DEPENDENCIES
   yard
 
 BUNDLED WITH
-   2.3.4
+   2.3.24
```
adrianna-chang-shopify, line 275: Should this have changed?

adrianna-chang-shopify (approved):

Gemfile.lock
etiennebarrie: Hmm I left it because I don't think it's an issue. Maybe I should have gone with latest though, it's a bit random. I can remove it as well.

PR #765, opened 2023-01-31
Remove deprecation warning from Selenium

etiennebarrie:
It's been removed in https://github.com/rails/rails/pull/47117

Follow up to #762 https://github.com/Shopify/maintenance_tasks/actions/runs/4043742587/jobs/6953035208


adrianna-chang-shopify (approved):
Nice find 👍

PR #773, opened 2023-02-14
Fix input for decimal\float attributes via web form

gdossantos:
#### Why

Decimal values cannot be input via web form for task Decimal\Float attributes as 'step' attribute is not specified on the 'number' input element.

If a decimal value is entered, the default browser form validation will prevent form submission.
Screen Shot 2023-02-13 at 10 56 31 PM

#### What
Set the 'step' attribute to 'any' for Decimal\Float 'number' input elements on web form to allow unrestricted decimal values to be provided as input.


gdossantos:
@nvasilevski cla should be updated.

nvasilevski (approved):
Thank you!

app/helpers/maintenance_tasks/tasks_helper.rb
```diff
@@ -104,9 +104,10 @@ def csv_file_download_path(run)
     # Return the appropriate field tag for the parameter
     def parameter_field(form_builder, parameter_name)
       case form_builder.object.class.attribute_types[parameter_name]
-      when ActiveModel::Type::Integer, ActiveModel::Type::Decimal,
-           ActiveModel::Type::Float
+      when ActiveModel::Type::Integer
         form_builder.number_field(parameter_name)
+      when ActiveModel::Type::Decimal, ActiveModel::Type::Float
+        form_builder.number_field(parameter_name, { step: "any" })
```
etiennebarrie, line 110: I'm surprised Rubocop didn't complain about the Hash style.

PR #782, opened 2023-03-02
Make error handler proc be called with job as self

szymonlipka:
In error handler I would like to have access to job's methods, so I think it would be perfect to use instance_exec to make `self` inside error_handler proc, instance of job in which error was raised.

### DIFF
--- a/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
+++ b/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
@@ -177,7 +177,7 @@ module MaintenanceTasks
       end
       errored_element = @errored_element if defined?(@errored_element)
     ensure
-      MaintenanceTasks.error_handler.call(error, task_context, errored_element)
+      instance_exec(*[error, task_context, errored_element], &MaintenanceTasks.error_handler)
     end
   end
 end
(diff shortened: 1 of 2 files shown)


szymonlipka:
Gendle bump on that ^. I need that in my company, for specific use of the handler

szymonlipka:
@etiennebarrie we have this script that runs maintenance tasks after every deploy, to test environments, that weren't run yet or errored. But there is problem that it's hard to ensure that while the script has run, sidekiq is reloaded with new code. So script runs task that sidekiq might not yet know and it picks it up and raises error.

I wanted to write simple error handler that would retry the task in this and only this situation, when there is no progress on the task, there is correct status and error is correct.

shalvah:
Hi @etiennebarrie you're probably busy, but it would be super cool if we could get this merged. 🙏 There seems to be a lot of folks who'll find it quite valuable. From a conceptual sense, it makes sense too: the error handler is typically used to report errors from the task; it can't do that effectively without proper knowledge of the task (the current task_context and errored_element are helpful, but not enough).

gmcgibbon:
Closing in favor of https://github.com/Shopify/maintenance_tasks/pull/942

PR #790, opened 2023-03-16
Handle Rails 7.1 deprecation of TestFixtures#fixture_path

etiennebarrie:
We still have to support 7.0 and below.

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -24,12 +24,15 @@ end
 Minitest.backtrace_filter = Minitest::BacktraceFilter.new
 
 # Load fixtures from the engine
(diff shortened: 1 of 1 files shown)


adrianna-chang-shopify (approved):
Can we leave a TODO indicating that `TestFixtures#fixture_path` is deprecated as of 7.1, so we remember why this branching exists and can remove it later?

etiennebarrie:
I feel like Git log and this PR may be enough to figure out why we did this 😛

adrianna-chang-shopify:
Fair, but I do like being able to search for explicit TODOs to make sure we're on top of removing support for different versions once they reach EOL. I don't think _I'll_ remember otherwise, but if you will that's fine 😉

PR #809, opened 2023-04-19
Allow tasks to be loaded from nested task directories

styrmis:
In [shop-server](https://github.com/Shopify/shop-server/tree/main/app/tasks/maintenance) we have tasks which are in sub-folders (e.g. `app/tasks/maintenance/deep_understanding`). These are available in production, but in development they are not automatically loaded, making them harder to test (TODO: verify this).

Before this PR, only tasks directly under `app/tasks` will be loaded in development. After this commit, one level of nesting is supported.

### DIFF
--- a/README.md
+++ b/README.md
@@ -696,6 +696,26 @@ MaintenanceTasks.tasks_module = "TaskModule"
 
 If no value is specified, it will default to `Maintenance`.
 
+#### Organizing tasks using namespaces
+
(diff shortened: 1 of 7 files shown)


styrmis:
Thanks @etiennebarrie, I wasn't aware of that one. I've cherry-picked the commit from that PR and have extended the tests to cover two levels of nesting. I've also added a section to the README.

PR #813, opened 2023-04-26
Fix parameter name when creating a run fails due to an argument that's too long

drcapulet:
@adrianna-chang-shopify this is what I asked about earlier today - technically SQLite does have a limit (default of 1 billion bytes), but trying that breaks Selenium.


drcapulet:
I have signed the CLA!

adrianna-chang-shopify (approved):
Thanks!

PR #814, opened 2023-04-27
Removes unused code

Schwad:
If my understanding is correct, this constant is no longer used. There is no reference to it elsewhere in the codebase.

Unless there is something special about this constant that a dependency relies upon?


etiennebarrie (approved):
Yes it looks like its only usage was removed in 1a44d6953e8228e47e89e2a4dc115f4dc9d5e8d2

PR #815, opened 2023-04-29
feat(tags): Adding the ability to tag tasks for filtering on the index page

lawrencewong:
- In our team's usage of maintenance tasks we found that with the additions of new tasks, and previous runs displaying raw data it made it difficult to find specific tasks.
- We wanted a way to be able to group certain tasks either by domain or usage.
- Adding the ability to filter by using links defined by tags on the tasks.
- If there are tags assigned to any task, it will be listed at the top of the index page. The tags are links to the tasks controller with a parameter of the selected tag. This will only surface tasks that have the given tag.

When no tasks have assigned tags:
Screenshot 2023-04-29 at 12 16 26 AM

When tasks have assigned tags:
Screenshot 2023-04-29 at 12 15 40 AM

Demo:
https://user-images.githubusercontent.com/3063931/235284910-65f7fd35-a18f-4214-b7e9-8b7f3efe3035.mov

### DIFF
--- a/app/controllers/maintenance_tasks/tasks_controller.rb
+++ b/app/controllers/maintenance_tasks/tasks_controller.rb
@@ -12,7 +12,9 @@ module MaintenanceTasks
     # Renders the maintenance_tasks/tasks page, displaying
     # available tasks to users, grouped by category.
     def index
-      @available_tasks = TaskDataIndex.available_tasks.group_by(&:category)
+      tag = params[:tag] || nil
+
+      @available_tasks = TaskDataIndex.available_tasks(tag).group_by(&:category)
     end
 
     # Renders the page responsible for providing Task actions to users.
--- a/app/controllers/maintenance_tasks/tasks_controller.rb
+++ b/app/controllers/maintenance_tasks/tasks_controller.rb
@@ -12,7 +12,7 @@ module MaintenanceTasks
     # Renders the maintenance_tasks/tasks page, displaying
     # available tasks to users, grouped by category.
     def index
-      tag = params[:tag] || nil
+      tag = params[:tag]
 
       @available_tasks = TaskDataIndex.available_tasks(tag).group_by(&:category)
     end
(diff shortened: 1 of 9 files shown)


app/controllers/maintenance_tasks/tasks_controller.rb
```diff
@@ -12,7 +12,9 @@ class TasksController < ApplicationController
     # Renders the maintenance_tasks/tasks page, displaying
     # available tasks to users, grouped by category.
     def index
-      @available_tasks = TaskDataIndex.available_tasks.group_by(&:category)
+      tag = params[:tag] || nil
```
desheikh, line 15: nil should be the default here

desheikh, line 15:
```suggestion
      tag = params[:tag]
```

params[:tag] will be nil if nothing is passed in

app/helpers/maintenance_tasks/tasks_helper.rb
```diff
@@ -63,6 +63,15 @@ def status_tag(status)
       tag.span(status.capitalize, class: ["tag"] + STATUS_COLOURS.fetch(status))
     end
 
+    # Returns the list of all available tags.
+    #
+    # This list is used to filter Tasks by tags in the index page.
+    #
+    # @return [Array<String>] the list of all available tags.
+    def tag_names
+      MaintenanceTasks::Task.available_tasks.map(&:tags).flatten.uniq
```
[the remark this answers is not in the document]
desheikh: We should memoize the result here

gmcgibbon (changes requested):
As discussed in the issue. We seem to be more open to grouping by task module / nesting. Can we please take that direction instead?

gmcgibbon:
Thanks for your contribution, but I'm closing this because there's no recent activity. If you're still interested in working on this feature, please open another PR.

PR #817, opened 2023-05-05
Inherit from `::ApplicationController`

shawnfrench:
I think we should default to inheriting from the application's `ApplicationController` as documented here: https://guides.rubyonrails.org/engines.html#using-a-controller-provided-by-the-application

This ensures the web UI handles requests/security etc as the including application has already specified.

### DIFF
--- a/app/controllers/maintenance_tasks/application_controller.rb
+++ b/app/controllers/maintenance_tasks/application_controller.rb
@@ -4,7 +4,7 @@ module MaintenanceTasks
   # Base class for all controllers used by this engine.
   #
   # Can be extended to add different authentication and authorization code.
-  class ApplicationController < ActionController::Base
+  class ApplicationController < ::ApplicationController
     BULMA_CDN = "https://cdn.jsdelivr.net"
 
     content_security_policy do |policy|


nvasilevski:
>2) requires the host app to have an ApplicationController

That would be my biggest concern as well, but I agree with

>and I'd expect most apps using maintenance_tasks to be Rails applications with ActionController

So I agree we should make this change and the only question is whether we want to treat it as a breaking one. I guess there is no harm in treating it as a breaking change even though realistically we don't expect it breaking any apps

adrianna-chang-shopify:
Oh good idea! I know I've seen Devise [follow that pattern](https://github.com/heartcombo/devise/blob/main/app/controllers/devise_controller.rb) too. Even if virtually all our consuming apps define their own `ApplicationController`s, it would be nice to not force apps to specify that class. Let me see if I can figure out how to appease YARD 👀

adrianna-chang-shopify (changes requested):
Chatted this through with Rafael, one issue with the current approach is that the app could have conflicting CSP headers that would prevent the maintenance task's UI from being able to render properly.

I don't think it's great for apps to have to override the engine's controller to hook into the main app's security policies though, so I think a flexible approach like what you've suggested in https://github.com/Shopify/maintenance_tasks/compare/main...srf/parent_controller would be ideal, just so that we don't break existing apps.

YARD unfortunately doesn't seem to do well with meta-programmed / dynamic class definitions, but I'm trying to see if we can do something with `@!parse`. At the very least, we can just ignore that warning in our docs test....

adrianna-chang-shopify (changes requested):
Chatted this through with Rafael, one issue with the current approach is that the app could have conflicting CSP headers that would prevent the maintenance task's UI from being able to render properly.

I don't think it's great for apps to have to override the engine's controller to hook into the main app's security policies though, so I think a flexible approach like what you've suggested in https://github.com/Shopify/maintenance_tasks/compare/main...srf/parent_controller would be ideal, just so that we don't break existing apps.

YARD unfortunately doesn't seem to do well with meta-programmed / dynamic class definitions, but I'm trying to see if we can do something with `@!parse`. At the very least, we can just ignore that warning in our docs test....

adrianna-chang-shopify:
> We could also make this feature optional but still default to the engine's ApplicationController.

I think this is what's being suggested in https://github.com/Shopify/maintenance_tasks/compare/main...srf/parent_controller @etiennebarrie , right?

etiennebarrie:
> > We could also make this feature optional but still default to the engine's ApplicationController.
> 
> I think this is what's being suggested in [main...srf/parent_controller](https://github.com/Shopify/maintenance_tasks/compare/main...srf/parent_controller) @etiennebarrie , right?

The default is `ActionController::Base`:
https://github.com/Shopify/maintenance_tasks/blob/216a63d274c9e1eac761e9e94fe7ad5281004284/lib/maintenance_tasks.rb#L84

PR #819, opened 2023-05-09
Add `parent_controller` configuration option

shawnfrench:
Add `MaintenanceTasks.parent_controller` configuration applications can assign the parent controller for the `maintenance_tasks` web UI. It defaults to `"ActionController::Base"` providing a flexible but non-breaking change.

This change allows applications with common logic in their `ApplicationController` (or any other controller) to optionally configure this web UI to inherit that logic with a simple assignment in the initializer.
eg.
`MaintenanceTasks.parent_controller = 'ApplicationController'`

## Why?
Many applications rely on every controller inheriting from their `ApplicationController` to provide consistency in pre/post processing, security, logging, etc.

Currently, because all controllers for this web UI inherit from `MaintenanceTasks::ApplicationController` (which inherits from `ActionController::Base`) the only way (afaik) to provide consistency across all controllers would be to extend the gem's `ApplicationController` with something like:
```
  MaintenanceTasks::ApplicationController.class_eval do
    include CommonThings
  end
```

That works if your application relies on a concern or two, but does not work for applications relying on `ApplicationController` inheritance, and is a bit unwieldy when you get into a large number of concerns you want to keep consistent across all controllers.

## Notes
Replaces https://github.com/Shopify/maintenance_tasks/pull/817 where I had suggested using `::ApplicationController` as the parent instead of `ActionController::Base` but that would be a breaking change, and also assumes an `ApplicationController` exists in the host app. This approach removes both obstacles.

The other part of this change is to ignore the YARD warning when trying to document our `ApplicationController` which now has a "Undocumentable superclass".

### DIFF
--- a/README.md
+++ b/README.md
@@ -804,6 +804,33 @@ MaintenanceTasks.backtrace_cleaner = cleaner
 If none is specified, the default `Rails.backtrace_cleaner` will be used to
 clean backtraces.
 
+#### Customizing the parent controller for the web UI
+
+`MaintenanceTasks.parent_controller` can be configured to specify a controller class 
+for all of the web UI engine's controllers to inherit from.
+
+This allows applications with common logic in their `ApplicationController` (or any other controller) 
+to optionally configure the web UI to inherit that logic with a simple assignment in the initializer.
+
+```ruby
+# config/initializers/maintenance_tasks.rb
+
+MaintenanceTasks.parent_controller = "CustomController"
+
+# app/controllers/services/custom_controller.rb
+
+class Services::CustomController < ActionController::Base
+  include CustomSecurityThings
+  include CustomLoggingThings
+  ...
+end
+```
+
+The parent controller value **must** be a string corresponding to an existing 
+controller class which **must inherit** from `ActionController::Base`. 
+
+If no value is specified, it will default to `"ActionController::Base"`.
+
 ## Upgrading
 
 Use bundler to check for and upgrade to newer versions. After installing a new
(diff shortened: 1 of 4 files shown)


adrianna-chang-shopify (approved):
This LGTM -- CI failures are unrelated (they're [stemming from a Capybara bug](https://github.com/teamcapybara/capybara/pull/2667)).

etiennebarrie (approved):
@adrianna-chang-shopify Do you think it's a feature we want to add to the README too?

shawnfrench:
Added note: https://github.com/Shopify/maintenance_tasks/blob/f9573e53094bf033eb700f044d52ad4d37897826/README.md#customizing-the-parent-controller-for-the-web-ui

README.md
```diff
@@ -804,6 +804,33 @@ MaintenanceTasks.backtrace_cleaner = cleaner
...
+`MaintenanceTasks.parent_controller` can be configured to specify a controller class 
+for all of the web UI engine's controllers to inherit from.
+
+This allows applications with common logic in their `ApplicationController` (or any other controller) 
+to optionally configure the web UI to inherit that logic with a simple assignment in the initializer.
+
+```ruby
+# config/initializers/maintenance_tasks.rb
+
+MaintenanceTasks.parent_controller = "CustomController"
```
etiennebarrie, line 818:
```suggestion
MaintenanceTasks.parent_controller = "Services::CustomController"
```

etiennebarrie (approved):
I'll fix the README and ship it once I get main green again.

PR #820, opened 2023-05-10
Lock selenium-webdriver to 4.9.0

etiennebarrie:
selenium-webdriver 4.9.1 is incompatible with the latest capybara

#818


adrianna-chang-shopify (approved):
Thanks!

PR #821, opened 2023-05-11
Fix parent_controller example

shawnfrench:
Fixing typo I introduced in #819


etiennebarrie:
Ugh sorry I thought I had fixed it. Thanks!

PR #831, opened 2023-05-31
feat(description): Add description option for tasks

SheldonNunes:
### What does this PR do and why is it necessary?

This PR adds the feature support to add descriptions to tasks. This is useful as it brings any comment/description from the source code into the UI itself and can be helpful for those who aren't developers.

### How does this PR work?

* Creates a new method `description` in the `task.rb` file. This is based off of a [similar method in graphql-ruby](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/member/base_dsl_methods.rb#L43-L51) that provides similar accessor functionality of setting the description.

### What could go wrong?

* It is possible that consumers may already have a property called `description` that could cause unexpected effects. Potential mitigation would be to rename to `task_description` which would have lower chance of collision.

### Questions

* Would reviewers prefer if this change did not impact the template?
* If we keep the template change should I update the `README` examples or keep them as is (description doesn't provide too much value for readers)?

### 🎩 🎩 🎩 🎩 🎩

image

image

image


velimovich:
Hey folks!
This is an amazing change, would love to have it in our project as soon as possible! Is there any chance that it will be included in one of next releases?

PR #833, opened 2023-06-01
Add helper text to datetime-local form fields

adrianna-chang-shopify:
Currently, datetime form fields are ambiguous because it's unclear whether the application timezone or system timezone will be used. This PR adds helper text to datetime form fields specifying that the system timezone will be used for datetime-local parameter inputs. We can even show the system timezone via `Time.now.zone` so that it's clearer for debugging purposes (e.g. local dev envs will use the user's local timezone vs prod generally using UTC).

🎩 

Screenshot 2023-06-01 at 9 17 37 AM

### DIFF
--- a/test/dummy/app/tasks/maintenance/params_task.rb
+++ b/test/dummy/app/tasks/maintenance/params_task.rb
@@ -29,7 +29,7 @@ module Maintenance
     def process(post)
       sleep(1) unless self.class.fast_task
 
-      post.update!(content: "New content added on #{Time.now.utc}")
+      post.update!(content: "New content added on #{Time.now.utc}:\ndatetime_attr: #{datetime_attr.inspect}")
     end
 
     private


app/helpers/maintenance_tasks/tasks_helper.rb
```diff
@@ -120,5 +120,13 @@ def parameter_field(form_builder, parameter_name)
         form_builder.text_area(parameter_name, class: "textarea")
       end
     end
+
+    # Return helper text for the datetime-local form field.
+    def datetime_field_help_text
+      tag.div(
+        tag.p("Datetime inputs will use the system timezone, which is #{Time.now.zone}."),
```
etiennebarrie, line 127: This may become incorrect if `ActiveModel::Type::DateTime` ever handles local times in accordance to `Rails.configuration.time_zone`/`Time.zone`. We should add a test for that to make sure we update this if it changes.

e.g. currently, `ActiveModel::Type::DateTime.new.cast("2023-06-01T12:00:00")` will not respect `Time.zone` but only the system time zone.

We could write a test that changes `Time.zone` to something that's not `Time.now.zone`, then tries to parse a local time as sent by a browser and checks that the timezone is correct.

etiennebarrie: Actually it's already incorrect if the application is not setting `config.time_zone`. In this case, it will always use UTC. If `config.time_zone` is set, it won't use it but use the system time zone instead… It will always ignore `Time.zone` though.

etiennebarrie (approved):

PR #858, opened 2023-07-26
Add metadata to keep track on who initiated a run

elfassy:
# Why

It is important for auditing purposes to know who initiated a maintenance task run. This PR adds an option to pass along run metadata that will be persisted in DB. 

part of https://hackdays.shopify.io/projects/16380

# What

- Adding a new optional `MaintenanceTasks.metadata` so users can set which method to call on their controller to get extra controller context (optional)
- Persist this data on the`maintenance_tasks_runs` table

```ruby
MaintenanceTasks.parent_controller = "YourAdminController"

# NEW
MaintenanceTasks.metadata = -> do 
  { 
    user_email: current_user.email, 
    request_ip: request.ip,
  }
end 
```

```ruby
class YourAdminController < ActionController::Base
  def current_user
    ...
  end
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -832,6 +832,19 @@ controller class which **must inherit** from `ActionController::Base`.
 
 If no value is specified, it will default to `"ActionController::Base"`.
 
+### Metadata
+
+`MaintenanceTasks.metadata` can be configured to specify a proc from which to get extra information about the run.
+Since this proc will be ran in the context of the `MaintenanceTasks.parent_controller`, it can be used to keep the id
+or email of the user who performed the maintenance task.
+
+```ruby
+# config/initializers/maintenance_tasks.rb
+MaintenanceTasks.metadata = -> do
+ { user_email: current_user.email }
+end
+```
+
 ## Upgrading
 
 Use bundler to check for and upgrade to newer versions. After installing a new
--- a/app/controllers/maintenance_tasks/runs_controller.rb
+++ b/app/controllers/maintenance_tasks/runs_controller.rb
@@ -14,6 +14,7 @@ module MaintenanceTasks
         name: params.fetch(:task_id),
         csv_file: params[:csv_file],
         arguments: params.fetch(:task_arguments, {}).permit!.to_h,
+        metadata: MaintenanceTasks.metadata&.call,
         &block
       )
       redirect_to(task_path(task))
--- a/lib/maintenance_tasks.rb
+++ b/lib/maintenance_tasks.rb
@@ -82,4 +82,11 @@ module MaintenanceTasks
   #
   #   @return [String] the name of the parent controller for web UI.
   mattr_accessor :parent_controller, default: "ActionController::Base"
+
+  # @!attribute metadata
+  #  @scope class
+  #   The Proc to call from the controller to generate metadata that will be persisted on the Run.
+  #
+  #   @return [Proc] generates a hash containing the metadata to be stored on the Run
+  mattr_accessor :metadata, default: nil
 end
--- a/app/models/maintenance_tasks/run.rb
+++ b/app/models/maintenance_tasks/run.rb
@@ -48,9 +48,11 @@ module MaintenanceTasks
     if Rails.gem_version >= Gem::Version.new("7.1.alpha")
       serialize :backtrace, coder: YAML
       serialize :arguments, coder: JSON
+      serialize :metadata, coder: JSON
     else
       serialize :backtrace
       serialize :arguments, JSON
+      serialize :metadata, JSON
     end
 
     scope :active, -> { where(status: ACTIVE_STATUSES) }
--- a/app/models/maintenance_tasks/runner.rb
+++ b/app/models/maintenance_tasks/runner.rb
@@ -39,8 +39,8 @@ module MaintenanceTasks
     #   creating the Run.
     # @raise [ActiveRecord::ValueTooLong] if the creation of the Run fails due
     #   to a value being too long for the column type.
-    def run(name:, csv_file: nil, arguments: {}, run_model: Run)
-      run = run_model.new(task_name: name, arguments: arguments)
+    def run(name:, csv_file: nil, arguments: {}, run_model: Run, metadata: nil)
+      run = run_model.new(task_name: name, arguments: arguments, metadata: metadata)
       if csv_file
         run.csv_file.attach(csv_file)
         run.csv_file.filename = filename(name)
--- a/test/system/maintenance_tasks/runs_test.rb
+++ b/test/system/maintenance_tasks/runs_test.rb
@@ -7,12 +7,38 @@ module MaintenanceTasks
     test "run a Task" do
       visit maintenance_tasks_path
 
-      click_on("Maintenance::UpdatePostsTask")
-      click_on "Run"
+      assert_difference("Run.count") do
+        click_on("Maintenance::UpdatePostsTask")
+        click_on "Run"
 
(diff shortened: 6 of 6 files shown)


db/migrate/20230622035228_add_user_email_to_runs.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddUserEmailToRuns < ActiveRecord::Migration[6.0]
+  def change
+    add_column(:maintenance_tasks_runs, :user_email, :string)
+  end
+end
```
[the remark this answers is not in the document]
elfassy: it's magic... in that they need to run `bin/rails maintenance_tasks:install:migrations` or manually as was done here https://github.com/Shopify/banking/pull/7880 😅 . This is a rough edge of engines gems

lib/maintenance_tasks.rb
```diff
@@ -82,4 +82,12 @@ module MaintenanceTasks
...
   #   @return [String] the name of the parent controller for web UI.
   mattr_accessor :parent_controller, default: "ActionController::Base"
+
+  # @!attribute user_email_method
+  #  @scope class
+  #   The method to call to get the email address of the current user to use when creating a Run.
+  #   This method must be defined on the parent_controller.
+  #
+  #   @return [Symbol] the name of the method to call on the current user.
+  mattr_accessor :user_email_method, default: nil
```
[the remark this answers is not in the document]
etiennebarrie: If we keep the option, I would definitely bike shed that. Rather than `user`, I'd rather something specific about the role they're playing, so maybe `initiator` or `creator`? And instead of `email`, we should be less specific because nothing forces it to be an email, since it's defined by the application. Implementors may use the field to store a `user id` instead.

test/application_system_test_case.rb
```diff
@@ -5,6 +5,7 @@
 require "action_dispatch/system_testing/server"
 
 ActionDispatch::SystemTesting::Server.silence_puma = true
+Webdrivers::Chromedriver.required_version = "114.0.5735.90"
```
etiennebarrie, line 8: We shouldn't need that either, we'll take care of the issue on main.

nvasilevski:
>in a typical deployment, there is visibility/auditability on who writes a migration (and who reviews it) and who/when the migration was deployed to production. Here we're missing the information about who ran the task (apart from logs), so that might be a good argument in favor.

I think this is a good argument this is why I would be an favor of such feature but I find the `user_email` concept to be very limited or way too specific. I think we might be able to provide much more flexibility by not calling this concept as anything specific but instead build a way for an application to attach some "metadata" to the run

So the `user_email` column in the database could be a serializable field instead and `user_email_method` method could instead be a `proc` that knows how to populate the metadata field. On top of that we could add an empty partial that is being rendered on the UI so applications can override it to render certain parts of metadata on the UI

lib/maintenance_tasks.rb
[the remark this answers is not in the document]
nvasilevski: I'm also leaning towards being less specific but was wondering if we should go even further and just allow some data to be stored in a serialized column. This way applications can decide how to call their actors and even decide to store more context about the actor in the field. A serialized column won't be searchable but I don't think this is something applications will need

elfassy:
@nvasilevski I agree! I'll leave the UI part as a follow up PR, but happy to update this one with your recommendations

test/system/maintenance_tasks/tasks_test.rb
```diff
@@ -64,8 +64,8 @@ class TasksTest < ApplicationSystemTestCase
 
       assert_equal ["Active Runs", "Previous Runs"], page.all("h4").map(&:text)
       runs = page.all("h5").map(&:text)
-      assert_includes runs, "July 18, 2022 11:05\nPaused"
-      assert_includes runs, "January 01, 2020 01:00\nSucceeded"
+      assert_includes runs, "July 18, 2022 21:05\nPaused"
+      assert_includes runs, "January 01, 2020 12:00\nSucceeded"
```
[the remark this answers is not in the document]
elfassy: will move this commit to a new PR, this test was invalid due to the `to_prepare` block being overwritten

test/application_system_test_case.rb
elfassy: opened a PR here https://github.com/Shopify/maintenance_tasks/pull/860

test/test_helper.rb
```diff
@@ -9,6 +9,7 @@
 ActiveRecord::Migrator.migrations_paths <<
   File.expand_path("../db/migrate", __dir__)
 require "rails/test_help"
+require "minitest/unit"
```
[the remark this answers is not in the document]
elfassy: ex: https://github.com/Shopify/maintenance_tasks/actions/runs/5672921352/job/15373220595

robinbrandt (approved):

README.md
```diff
@@ -832,6 +832,19 @@ controller class which **must inherit** from `ActionController::Base`.
 
 If no value is specified, it will default to `"ActionController::Base"`.
 
+### Metadata
+
+`MaintenanceTasks.metadata` can be configured to specify a proc from which to get extra information about the run.
+Since this proc will be ran in the context of the MaintenanceTasks.parent_controller, it can be used to keep the id
+or email of the user who performed the maintenaince task.
```
adrianna-chang-shopify, line 839:
```suggestion
or email of the user who performed the maintenance task.
```

lib/maintenance_tasks.rb
```diff
@@ -82,4 +82,11 @@ module MaintenanceTasks
   #
   #   @return [String] the name of the parent controller for web UI.
   mattr_accessor :parent_controller, default: "ActionController::Base"
+
+  # @!attribute metadata
+  #  @scope class
+  #   The Proc to call from the controller to generate metadata that will be persisted on the Run.
+  #
+  #   @return [Proc] generates a hash containting the metadata to be stored on the Run
```
adrianna-chang-shopify, line 90:
```suggestion
  #   @return [Proc] generates a hash containing the metadata to be stored on the Run
```

README.md
```diff
@@ -832,6 +832,19 @@ controller class which **must inherit** from `ActionController::Base`.
 
 If no value is specified, it will default to `"ActionController::Base"`.
 
+### Metadata
+
+`MaintenanceTasks.metadata` can be configured to specify a proc from which to get extra information about the run.
+Since this proc will be ran in the context of the MaintenanceTasks.parent_controller, it can be used to keep the id
```
adrianna-chang-shopify, line 838:
```suggestion
Since this proc will be ran in the context of the `MaintenanceTasks.parent_controller`, it can be used to keep the id
```

adrianna-chang-shopify (approved):
This solution seems appropriately generic to me 👍 Let's wait on feedback from @nvasilevski and @etiennebarrie before merging though.

app/controllers/maintenance_tasks/runs_controller.rb
```diff
@@ -14,6 +14,7 @@ def create(&block)
         name: params.fetch(:task_id),
         csv_file: params[:csv_file],
         arguments: params.fetch(:task_arguments, {}).permit!.to_h,
+        metadata: MaintenanceTasks.metadata&.call,
```
segiddins, line 17: The README states that the `metadata` proc will be exec'd in the scope of the controller, but this isn't doing that -- did you intend to `instance_exec(MaintenanceTasks.metadata) if MaintenanceTasks.metadata` ?

PR #859, opened 2023-07-27
Expose ActiveJob ability to customize the columns used to order the cursor

bbraschi:
`JobIteration::ActiveRecordCursor#condition` adds an `order by` clause to the relation returned by the `collection` method. It default to order on the `id` column, see `JobIteration::ActiveRecordCursor#initialize`.

Ordering on `id` only can degrade performance significantly depending on which indexes are defined.

`job-iteration` supports configuring which columns are used to order the cursor. This PR explores a way to expose this ability in the `maintenance-tasks` gem.

### DIFF
--- a/app/models/maintenance_tasks/task.rb
+++ b/app/models/maintenance_tasks/task.rb
@@ -230,6 +230,18 @@ module MaintenanceTasks
       self.class.collection_builder_strategy.collection(self)
     end
 
+    # The columns used to build the `ORDER BY` clause of the query for iteration.
+    #
+    # If cursor_columns returns nil, the query is ordered by the primary key.
+    # If cursor columns values change during an iteration, records may be skipped or yielded multiple times.
+    # More details in the documentation of JobIteration::EnumeratorBuilder.build_active_record_enumerator_on_records:
+    # https://www.rubydoc.info/gems/job-iteration/JobIteration/EnumeratorBuilder#build_active_record_enumerator_on_records-instance_method
+    #
+    # @return the cursor_columns.
+    def cursor_columns
+      nil
+    end
+
     # Placeholder method to raise in case a subclass fails to implement the
     # expected instance method.
     #
@@ -264,7 +276,7 @@ module MaintenanceTasks
       when :no_collection
         job_iteration_builder.build_once_enumerator(cursor: nil)
       when ActiveRecord::Relation
-        job_iteration_builder.active_record_on_records(collection, cursor: cursor)
+        job_iteration_builder.active_record_on_records(collection, cursor: cursor, columns: cursor_columns)
       when ActiveRecord::Batches::BatchEnumerator
         if collection.start || collection.finish
           raise ArgumentError, <<~MSG.squish
@@ -279,6 +291,7 @@ module MaintenanceTasks
           collection.relation,
           cursor: cursor,
           batch_size: collection.batch_size,
+          columns: cursor_columns,
         )
       when Array
         job_iteration_builder.build_array_enumerator(collection, cursor: cursor&.to_i)
--- a/README.md
+++ b/README.md
@@ -413,6 +413,37 @@ to run. Since arguments are specified in the user interface via text area
 inputs, it’s important to check that they conform to the format your Task
 expects, and to sanitize any inputs if necessary.
 
+### Custom cursor columns to improve performance
+
+The [job-iteration gem](https://www.rubydoc.info/gems/job-iteration), on which this gem depends, adds an `order by`
+clause to the relation returned by the `collection` method, in order to
+iterate through records. It defaults to order on the `id` column.
+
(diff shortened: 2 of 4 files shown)


app/models/maintenance_tasks/task.rb
```diff
@@ -21,6 +21,9 @@ class NotFoundError < NameError; end
     # @api private
     class_attribute :collection_builder_strategy, default: NullCollectionBuilder.new
 
+    # As defined by JobIteration::EnumeratorBuilder#build_active_record_enumerator_on_records as `columns`
+    class_attribute :cursor_columns, default: nil
```
bbraschi, line 25: I searched for a way to let subclasses of `MaintenanceTasks::Task` customize how the `enumerator_builder` instantiates the underlying relation, without success. 

Using a `class_attribute` as an alternative. Is it appropriate?

app/jobs/concerns/maintenance_tasks/task_job_concern.rb
```diff
@@ -39,7 +39,11 @@ def build_enumerator(_run, cursor:)
       when :no_collection
         enumerator_builder.build_once_enumerator(cursor: nil)
       when ActiveRecord::Relation
-        enumerator_builder.active_record_on_records(collection, cursor: cursor)
+        enumerator_builder.active_record_on_records(
+          collection,
+          cursor: cursor,
+          **{ columns: @task.class.cursor_columns }.compact,
```
rafaelfranca, line 45: Why the `compact`? Isn't better to check if `cursor_columns` are set?

app/models/maintenance_tasks/task.rb
```diff
@@ -21,6 +21,9 @@ class NotFoundError < NameError; end
     # @api private
     class_attribute :collection_builder_strategy, default: NullCollectionBuilder.new
 
+    # As defined by JobIteration::EnumeratorBuilder#build_active_record_enumerator_on_records as `columns`
```
rafaelfranca, line 24: Can you improve this documentation telling what this does instead of linking to method in another library? For the sake of this library, this method is only to configure which column will be used in the Active Record enumerator. Let's explain this and then, we can link to the other library, but if we are going to do that, we should really link to the library documentation, not just mention the reference of the method.

test/application_system_test_case.rb
```diff
@@ -5,6 +5,7 @@
 require "action_dispatch/system_testing/server"
 
 ActionDispatch::SystemTesting::Server.silence_puma = true
+Webdrivers::Chromedriver.required_version = "114.0.5735.90"
```
rafaelfranca, line 8: We can remove this now I believe

gmcgibbon:
@bbraschi are you still interested in working on this?

bbraschi:
@gmcgibbon If you have you have the time to bring this to a conclusion, through this PR or another, please do. Else, I can work on it at the end of Jan 2024.

app/models/maintenance_tasks/task.rb
```diff
@@ -21,6 +21,16 @@ class NotFoundError < NameError; end
     # @api private
     class_attribute :collection_builder_strategy, default: NullCollectionBuilder.new
 
+    # The cursor_columns attribute is used to build the `ORDER BY` clause of the query for iteration.
```
gmcgibbon, line 24:
```suggestion
    # Used to build the `ORDER BY` clause of the query for iteration.
```

app/models/maintenance_tasks/task.rb
```diff
@@ -21,6 +21,16 @@ class NotFoundError < NameError; end
...
+    # If the cursor_columns attribute is not set or is set to nil, the query is ordered by the primary key.
```
gmcgibbon, line 25:
```suggestion
    # If the attribute is not set, the query is ordered by the primary key.
```

app/models/maintenance_tasks/task.rb
```diff
@@ -21,6 +21,16 @@ class NotFoundError < NameError; end
...
+    # If the values in the cursor columns change for the records in scope during an
```
gmcgibbon, line 26:
```suggestion
    # If cursor columns values change during an
```

app/models/maintenance_tasks/task.rb
```diff
@@ -21,6 +21,16 @@ class NotFoundError < NameError; end
...
+    # iteration, records may be skipped or yielded multiple times depending on the nature of the update and the
+    # cursor's values. If the values gets updated to greater values than the cursor's values, records will get yielded
```
gmcgibbon, line 28:
```suggestion
```
I think the rest is covered by the statement above. Let's link to job-iteration and provide an example of values here instead of going into more detail about the potentially unwanted behaviour.

app/models/maintenance_tasks/task.rb
gmcgibbon: This would be overridden by specific tasks, so I think the best thing to do is provide a simple method like `collection` that can be overridden.

gmcgibbon (approved):

PR #860, opened 2023-07-28
fix tests using a non-production version of chrome in CI

elfassy:
Fix errors in the test suite ex: https://github.com/Shopify/maintenance_tasks/actions/runs/5696018238/job/15440308078?pr=858

> Unable to find latest point release version for 115.0.5790. You appear to be using a non-production version of Chrome. Please set `Webdrivers::Chromedriver.required_version = <desired driver version>` to a known chromedriver version: https://chromedriver.storage.googleapis.com/index.html

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,8 +20,7 @@ else
 end
 gem "rubocop"
 gem "rubocop-shopify"
-gem "selenium-webdriver", "< 4.10.1"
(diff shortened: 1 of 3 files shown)


elfassy:
@adrianna-chang-shopify good call, made the change and tests seems to like it 👍

adrianna-chang-shopify (approved):

PR #866, opened 2023-08-18
Avoid loading all tasks to validate a `Run#task_name`

davidstosik:
Currently, creating a `Run` record validates that its `task_name` belongs to the list returned by `Task.available_tasks`.

Generating this list requires the gem to browse through the task namespace(s) to find and load all tasks in the application (see `Task#load_constants`).

Although in most situations this should not be a problem, the Shopify monolith defines a relatively large amount of maintenance tasks, which take a considerable amount of time to load.

Loading all tasks is only absolutely necessary when the application wants to display a list of all tasks. In any other scenario, for example testing a single task, it should not be necessary to load all tasks.

This commit makes that possible by making use of `Task.named` to confirm that the task name is valid.
Because `available_tasks` relies on `.descendants`, and `Task.named` uses ` < Task` to test inheritance, I don't think this will have any unintended effect on the gem's behaviour.

### DIFF
--- a/app/models/maintenance_tasks/run.rb
+++ b/app/models/maintenance_tasks/run.rb
@@ -36,9 +36,7 @@ module MaintenanceTasks
 
     enum status: STATUSES.to_h { |status| [status, status.to_s] }
 
-    validates :task_name, on: :create, inclusion: {
-      in: ->(_) { Task.available_tasks.map(&:to_s) },
-    }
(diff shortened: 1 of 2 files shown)


etiennebarrie (approved):
Yeah I think that's what we really meant by checking for inclusion. We just wanted the task name to correspond to an existing task.

PR #867, opened 2023-08-18
Make the task preloading process configurable

davidstosik:
In ordered to be displayed in the UI, tasks have to be loaded by the time `MaintenanceTasks::Task.descendants` is called.
To achieve that, the `Task.load_constants` method was previously implemented, and called from `Task.available_tasks`.

A few inconvenients of that method are:

- It could easily be abused and used in lower levels of the gem, where ideally it should not be necessary to load all tasks to achieve the gem's goal. See for example my previous PR [Avoid loading all tasks to validate a `Run#task_name`](https://github.com/Shopify/maintenance_tasks/pull/866).

- It can introduce unnecessary overhead. For example, displaying the CLI's basic help message with `maintenance_tasks help` will necessitate all tasks to be loaded (even though the help message does not display anything related to the available tasks), and this loading time will delay the appearance of that help message.
  Similarly, a single test case that wants to test a single task's behaviour, while never displaying the whole list of available tasks, would be negatively impacted by a misuse of `Task.available_tasks`.

- It does not allow an application to define an alternative way to load all constants. As a matter of fact, the Shopify Core monolith monkey-patched that method to allow multiple "task modules" instead of a single one. With a configurable `task_loader`, monkey-patching would not be necessary. (Which is good. 🙈)

This PR introduces `MaintenanceTasks.task_loader`, a configuration that allows applications to change the logic applied to ensure all task are loaded before being displayed.
As a bonus:

- it changes how the CLI generates the long description containing all task names, so that tasks get loaded at the last minute, only when necessary
- it gets rid of `Task.available_tasks` (its only use was for UI purposes, so I changed it to a direct call to `Task.descendants`)
- it introduces some tests of the loading behaviour

Please note that this PR is built on top of https://github.com/Shopify/maintenance_tasks/pull/866 as it is a requirement.

### DIFF
--- a/test/models/maintenance_tasks/default_task_loader_test.rb
+++ b/test/models/maintenance_tasks/default_task_loader_test.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+module MaintenanceTasks
+  class DefaultTaskLoaderTest < ActiveSupport::TestCase
+    include ActiveSupport::Testing::Isolation
+
+    test ".load_all loads all autoloadable Task subclasses within the Maintenance namespace" do
+      [
+        Maintenance::BatchImportPostsTask,
+        Maintenance::CallbackTestTask,
+        Maintenance::CancelledEnqueueTask,
+        Maintenance::EnqueueErrorTask,
+        Maintenance::ErrorTask,
+        Maintenance::ImportPostsTask,
+        Maintenance::Nested,
+        Maintenance::Nested::NestedMore,
+        Maintenance::Nested::NestedMore::NestedMoreTask,
+        Maintenance::Nested::NestedTask,
+        Maintenance::NoCollectionTask,
+        Maintenance::ParamsTask,
+        Maintenance::TestTask,
+        Maintenance::UpdatePostsInBatchesTask,
+        Maintenance::UpdatePostsModulePrependedTask,
+        Maintenance::UpdatePostsTask,
+        Maintenance::UpdatePostsThrottledTask,
+      ].each do |constant|
+        constant
+          .module_parent
+          .expects(:const_get)
+          .with(constant.name.demodulize.to_sym)
+          .returns(constant)
+      end
+
+      DefaultTaskLoader.load_all
+    end
+  end
+end
--- a/README.md
+++ b/README.md
@@ -41,6 +41,10 @@ Please ensure your application is using
 information, please consult the [Rails guides on autoloading and reloading
 constants](https://guides.rubyonrails.org/autoloading_and_reloading_constants.html).
 
+You might be able to define a different
+[_task loader_](#customizing-the-way-the-gem-loads-all-tasks) that would allow
+you to circumvent this limitation.
+
 ## Usage
 
 The typical Maintenance Tasks workflow is as follows:
@@ -714,6 +718,35 @@ module Maintenance
     end
   end
 end
+```
+
+#### Customizing the way the gem loads all tasks
+
+For the UI (Web and CLI) to be able to display all available tasks, their
+respective classes must have been previously loaded. This can be an issue
+for applications running without
+[eager loading](https://guides.rubyonrails.org/configuring.html#config-eager-load)
+(for example a default Rails application uses lazy loading in development
+mode).
+
+To work around that, this gem uses a task loader, which can be customized
+by changing the `MaintenanceTasks.task_loader` in the configuration:
+
+```rb
+# config/initializers/maintenance_tasks.rb
+
+MaintenanceTasks.task_loader = -> { DefaultTaskLoader.load_all }
+```
+
+By using a different proc, you can define a different way to preload all tasks.
+You could for example browse for tasks in a given directory:
+
+```rb
+# config/initializers/maintenance_tasks.rb
+
+MaintenanceTasks.task_loader = -> {
+  Dir[Rails.root.join("app", "tasks", "**", "*_task.rb")].each { |file| require file }
+}
 ```
 
 #### Customizing the underlying job class
--- a/app/models/maintenance_tasks/default_task_loader.rb
+++ b/app/models/maintenance_tasks/default_task_loader.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+module MaintenanceTasks
+  # This class defines how the gem will find and load all tasks defined by the application.
+  # It can be overridden by the application to change the way tasks are loaded.
+  class DefaultTaskLoader
+    class << self
+      # Recursively browse the MaintenanceTasks.tasks_module namespace to load all defined tasks.
+      #
+      # @return [void]
+      def load_all
+        load_constants(MaintenanceTasks.tasks_module.safe_constantize)
+      end
+
+      private
+
+      def load_constants(namespace)
+        namespace.constants.each do |name|
+          object = namespace.const_get(name)
+          load_constants(object) if object.instance_of?(Module)
+        end
+      end
+    end
+  end
+end
--- a/app/models/maintenance_tasks/task.rb
+++ b/app/models/maintenance_tasks/task.rb
@@ -41,15 +41,6 @@ module MaintenanceTasks
         task
       end
 
-      # Returns a list of concrete classes that inherit from the Task
-      # superclass.
-      #
-      # @return [Array<Class>] the list of classes.
-      def available_tasks
-        load_constants
-        descendants
-      end
-
       # Make this Task a task that handles CSV.
       #
       # @param in_batches [Integer] optionally, supply a batch size if the CSV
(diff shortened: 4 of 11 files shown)


test/models/maintenance_tasks/default_task_loader_test.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+module MaintenanceTasks
+  class DefaultTaskLoaderTest < ActiveSupport::TestCase
+    include ActiveSupport::Testing::Isolation
+
+    test ".load_all loads all autoloadable Task subclasses within the Maintenance namespace" do
```
davidstosik, line 9: This new test fails in Rails 6.
Has it possibly identified an issue with the gem in older Rails versions? I haven't altered the recursive loading algorithm, just moved it from `Task.load_constants` to `DefaultTaskLoader.load_constants`. 🤔

davidstosik:
```
Failure:
MaintenanceTasks::DefaultTaskLoaderTest#test_.load_all_loads_all_autoloadable_Task_subclasses_within_the_Maintenance_namespace [/home/runner/work/maintenance_tasks/maintenance_tasks/vendor/bundle/ruby/3.0.0/gems/activesupport-6.0.6.1/lib/active_support/inflector/methods.rb:284]:
unexpected invocation: Maintenance.const_get("Nested")
unsatisfied expectations:
- expected exactly once, invoked never: Maintenance.const_get(:Nested)
- expected exactly once, invoked never: Maintenance.const_get(:ImportPostsTask)
- expected exactly once, invoked never: Maintenance.const_get(:ErrorTask)
- expected exactly once, invoked never: Maintenance.const_get(:EnqueueErrorTask)
- expected exactly once, invoked never: Maintenance.const_get(:CancelledEnqueueTask)
- expected exactly once, invoked never: Maintenance.const_get(:CallbackTestTask)
- expected exactly once, invoked never: Maintenance.const_get(:BatchImportPostsTask)
```

Wonder if it's just because of a different order... 🤔

davidstosik: No, the order returned by `.constants` seems to differ depending on the version of Rails.
This would work in Ruby 6.1 for example:

```diff
diff --git i/app/models/maintenance_tasks/default_task_loader.rb w/app/models/maintenance_tasks/default_task_loader.rb
index 14f1624..fed0ea5 100644
--- i/app/models/maintenance_tasks/default_task_loader.rb
+++ w/app/models/maintenance_tasks/default_task_loader.rb
@@ -15,7 +15,8 @@ module MaintenanceTasks
       private
 
       def load_constants(namespace)
-        namespace.constants.each do |name|
+        namespace.constants.sort.each do |name|
+          puts "#{namespace}.const_get(#{name})"
           object = namespace.const_get(name)
           load_constants(object) if object.instance_of?(Module)
         end
diff --git i/test/models/maintenance_tasks/default_task_loader_test.rb w/test/models/maintenance_tasks/default_task_loader_test.rb
index 8fd8ef9..b029518 100644
--- i/test/models/maintenance_tasks/default_task_loader_test.rb
+++ w/test/models/maintenance_tasks/default_task_loader_test.rb
@@ -8,6 +8,9 @@ module MaintenanceTasks
 
     test ".load_all loads all autoloadable Task subclasses within the Maintenance namespace" do
       [
+        Maintenance::Nested::NestedMore::NestedMoreTask,
+        Maintenance::Nested::NestedTask,
+        Maintenance::Nested::NestedMore,
         Maintenance::BatchImportPostsTask,
         Maintenance::CallbackTestTask,
         Maintenance::CancelledEnqueueTask,
@@ -15,9 +18,6 @@ module MaintenanceTasks
         Maintenance::ErrorTask,
         Maintenance::ImportPostsTask,
         Maintenance::Nested,
-        Maintenance::Nested::NestedMore,
-        Maintenance::Nested::NestedMore::NestedMoreTask,
-        Maintenance::Nested::NestedTask,
         Maintenance::NoCollectionTask,
         Maintenance::ParamsTask,
         Maintenance::TestTask,
```

adrianna-chang-shopify: Actually @davidstosik if I change the order to what you proposed in https://github.com/Shopify/maintenance_tasks/pull/867#discussion_r1304083757, it passes against all the Rails versions for me? Were you seeing failures locally?

davidstosik: @adrianna Huh, wow, this is weird, you seem to be right. Although both orders work with Rails 7, it seems only a specific order works with Rails 6.1?

For fun (and because I hadn't imagined the right order would work for all supported Rails versions), I've also implemented a spy. Here it is, just to show what it could look like:

```rb
# frozen_string_literal: true

require "test_helper"

module MaintenanceTasks
  class DefaultTaskLoaderTest < ActiveSupport::TestCase
    class ConstGetSpy < SimpleDelegator
      cattr_reader :calls, default: Hash.new { |h, k| h[k] = [] }

      def const_get(name)
        calls[spied_module] << name

        return_value = super

        if return_value.is_a?(Module)
          self.class.new(return_value)
        else
          return_value
        end
      end

      def instance_of?(klass)
        spied_module.instance_of?(klass)
      end

      private

      def spied_module
        __getobj__
      end
    end

    SPIED_MAINTENANCE_TASKS_MODULE = ConstGetSpy.new(Maintenance)

    test ".load_all loads all autoloadable Task subclasses within the Maintenance namespace" do
      original_tasks_module = MaintenanceTasks.tasks_module
      MaintenanceTasks.tasks_module = "MaintenanceTasks::DefaultTaskLoaderTest::SPIED_MAINTENANCE_TASKS_MODULE"

      expected_calls_to_const_get = {
        Maintenance => [
          :BatchImportPostsTask,
          :CallbackTestTask,
          :CancelledEnqueueTask,
          :EnqueueErrorTask,
          :ErrorTask,
          :ImportPostsTask,
          :Nested,
          :NoCollectionTask,
          :ParamsTask,
          :TestTask,
          :UpdatePostsInBatchesTask,
          :UpdatePostsModulePrependedTask,
          :UpdatePostsTask,
          :UpdatePostsThrottledTask,
        ],
        Maintenance::Nested => [
          :NestedMore,
          :NestedTask,
        ],
        Maintenance::Nested::NestedMore => [
          :NestedMoreTask,
        ]
      }

      DefaultTaskLoader.load_all

      assert_equal(expected_calls_to_const_get, ConstGetSpy.calls.transform_values(&:sort))

      MaintenanceTasks.tasks_module = original_tasks_module
    end
  end
end
```

(I saw your more general comment about the PR, will reply next. 🙇🏻‍♂️ )

etiennebarrie:
> this would also allow apps using the classic autoloader to use the gem (though most apps should be on Zeitwerk by now).

We dropped support for the classic autoloader in https://github.com/Shopify/maintenance_tasks/pull/620 released in 2.0.

Also let's keep in mind that `Task.available_tasks` is a public API so I'm not super motivated to bump the major to remove it. 

@davidstosik Ultimately didn't #866 fix the main issue you had with `available_tasks`?

Similar point to Adrianna, but maybe we could keep the stuff around the Thor command to avoid loading all the tasks unless necessary to display the help for perform.

adrianna-chang-shopify:
@davidstosik thank you for expanding on your thoughts ❤️ 

> Not when I do bundle exec maintenance_tasks or bundle exec maintenance_tasks perform my_task_name.

Yep, sorry, I meant that regardless of whether we extract a configurable loader, we still need to load all the tasks to show the long version of the `perform` command. Either way, I think redefining that command to delay loading the task names is a great improvement 👍 

> I've been advised in the past against writing monkey patches for gems we own.

I definitely understand your point here (and usually feel the same way). I think the challenge here is striking a balance between designing the gem in a way that avoids us having to write monkey patches in Core, while also ensuring that the gem remains simple and easy-to-use for the majority of our community, without a ton of knobs that can be configured solely because _our_ monolith needs them. Core has a ton of custom stuff around maintenance tasks (the whole CPT system, custom controllers and views, etc.) so overriding `#available_tasks` seems minor in comparison. 

I think it's also worth mentioning that in Core, we need that patch because we're allowing tasks to be nested in components (reusable tasks and all that), but that the goal is to go back to having all tasks in the top-level `maintenance` namespace (which should remove the need for the patch).

I'm not sure we want to remove `.available_tasks`, but I'd be okay with renaming it to something more meaningful, as you've suggested. Also in favour of that tweak to the CLI (which we can tackle separately) if you're up for it 😄 

Let me know if this makes sense, or if you have other questions or concerns.

davidstosik:
@adrianna-chang-shopify Thank you for your message!

I have a few ideas:

- I'll open a separate PR to improve the CLI
- we could rename `Task.available_tasks` to `Task.load_all` and keep the old method with a deprecation message
- we could also have the method warn about what it does, and cancel that warning by calling `Task.available_tasks(warn: false)`

What do you think?

davidstosik:
Closing this in favor of #877 and #878.

PR #878, opened 2023-08-29
Rename `Task.available_tasks` to `.load_all` and deprecate

davidstosik:
`Task.available_tasks` not only returns all descendants of the `Task` class, it also `require`s all files in the application that may be defining such class.

`available_tasks` does not carry explicit meaning of how heavy the operation could be, so this renames it to `load_all`, which should be better at communicating that, and deprecates the old name.

I'll have to adjust this PR once https://github.com/Shopify/maintenance_tasks/pull/877 is merged.

### DIFF
--- a/app/models/maintenance_tasks/task.rb
+++ b/app/models/maintenance_tasks/task.rb
@@ -41,13 +41,25 @@ module MaintenanceTasks
         task
       end
 
-      # Returns a list of concrete classes that inherit from the Task
-      # superclass.
+      # Loads and returns a list of concrete classes that inherit
+      # from the Task superclass.
+      #
+      # @return [Array<Class>] the list of classes.
+      def load_all
+        load_constants
+        descendants
+      end
+
+      # Loads and returns a list of concrete classes that inherit
+      # from the Task superclass.
       #
       # @return [Array<Class>] the list of classes.
       def available_tasks
-        load_constants
-        descendants
+        warn(<<~MSG.squish, category: :deprecated)
+          MaintenanceTasks::Task.available_tasks is deprecated and will be
+          removed from maintenance-tasks 3.0.0. Use .load_all instead.
+        MSG
+        load_all
       end
 
       # Make this Task a task that handles CSV.
--- a/app/models/maintenance_tasks/task_data_index.rb
+++ b/app/models/maintenance_tasks/task_data_index.rb
@@ -25,7 +25,7 @@ module MaintenanceTasks
       def available_tasks
         tasks = []
 
-        task_names = Task.available_tasks.map(&:name)
+        task_names = Task.load_all.map(&:name)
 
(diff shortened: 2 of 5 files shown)


etiennebarrie (commented):
I'm not in favor of this. I prefer the name to be a noun than a verb. We can add documentation instead.
I also don't think it's correct to say that it requires or loads classes, Zeitwerk does that.

davidstosik:
@etiennebarrie Thank you for your comment.

I disagree with your statement. The whole purpose of defining this method instead of calling `descendants` is to make sure that all Ruby files defining a task have been **required**. (Using `const_get` is just a trick to force Zeitwerk to require them if they haven't been already.) Also note that it calls a method named `load_constants`.

A noun is exactly what lacks meaning of a possibly heavy task, in my opinion. Developers calling `Task.available_tasks` might expect the list of tasks to be just there, ready to be looked at, whereas in reality calling that method will sometimes trigger the requirement of multiple files (more than 600 in Shopify's case, for example), and every time it'll trigger a recursive process of looking into all constant names under the `Maintenance` namespace.

I'm afraid relying on documentation alone would likely not reach most developers.

I also assumed that I had obtained agreement to do this before opening this PR, [here](https://github.com/Shopify/maintenance_tasks/pull/867#issuecomment-1697908879).
Happy to discuss and study alternatives though.

adrianna-chang-shopify (approved):
This change makes sense to me, although let's discuss further if you still have reservations @etiennebarrie .

While I understand your perspective on the noun vs verb stance, noun (`#available_tasks`) to me implies a simple getter, and the case @davidstosik is making is that this method is often _not_ a simple getter and can be computationally expensive. The fact that Zeitwerk does this under the hood is unimportant -- saying that the method "loads" classes is appropriate IMHO because it forces Zeitwerk to load constants that aren't yet loaded.

While I agree that we should reach for changes in the documentation ahead of changing the API, I support David's point that the current method name of `#available_tasks` fails to convey the computational side of the method.

davidstosik:
Thank you @adrianna-chang-shopify!
I'll merge my PR now, but please let me know if there's anything you'd like me to improve on this. 🙏🏻

adrianna-chang-shopify:
> And in that case, I don't expect a verb but really a name.

Right, but the reality is that the method _does_ load constants in dev / test envs (for developer exp reasons, but that was a decision we made). To me, this change is simply about signalling to users that this method loads constants and can thus be slow, which should hopefully prompt users to think before reaching for it in a test or in their own code.

> BTW we've previously told users some features we didn't want to build in could be handled by overriding available_tasks

I didn't like telling people to patch it 😂 And anyways, we didn't suggest it in any official capacity, so I'm not too worried about breaking that contract. I did grok our internal repos, and the only repo overriding it currently is Core.

PR #892, opened 2023-09-15
Update readme with example for backfill

csalvato:
We found a nice pattern for running queries without pulling and uploading large CSVs.

For our own records, we wanted to add it to our own docs, but felt that the rest of the community could benefit, so added it to the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,6 +201,37 @@ Note that `#count` is calculated automatically based on the number of batches in
 your collection, and your Task’s progress will be displayed in terms of batches
 (not the number of records in the relation).
 
+On databases with large tables, sometime using an `ActiveRecord::Relation` to define a collection will
+cause your queries to timeout. One pattern to to have performant tasks is to look up the range of IDs that
+need to be operated on, then create a collection of ranges to query your models. For example:
+
+```ruby
+module Maintenance
+  class BackfillALotOfPosts < MaintenanceTasks::Task
+    # Generates an array like
+    # [(1..1_000_000),(1_000_001..2_000_000),(2_000_001..3_000_000), ...]
+    def collection
+      ranges = []
+      100.times do |i|
+        range_start = (1_000_000 * i) + 1
+        range_end = 1_000_000 * (i + 1)
+        ranges << (range_start..range_end)
+      end
+      ranges
+    end
+
+    def process(range)
+      Post
+        .where(id: range, some_column: "Some Value")
+        .in_batches { |batch| batch.update_all(some_column: "Some New Value") }
+    end
+  end
+end
+```
+
+This pattern will keep your individual queries fast, while performing updates in small batches.
+With this approach, updating or deleteing 10s of millions of records can take under an hour.
+
 **Important!** Batches should only be used if `#process` is performing a batch
 operation such as `#update_all` or `#delete_all`. If you need to iterate over
 individual records, you should define a collection that [returns an


README.md
```diff
@@ -201,6 +201,38 @@ Note that `#count` is calculated automatically based on the number of batches in
...
+        range_end = 1_000_000 * (i + 1)
+        ranges << (range_start..range_end)
+      end
+      ranges
+    end
+
+    def process(range)
+      Post
+        .where(id: range, some_column: "Some Value")
+        .in_batches { |batch| batch.update_all(some_column: "Some New Value") }
```
nvasilevski, line 226: Would it be more reasonable to recommend building smaller ranges instead of using `in_batches` inside the `process` method? It will make the `collection` method slower but will allow the task to be interrupted more frequently and reduce the risk of the `process` method taking more than the max configured time for a single iteration

README.md
```diff
@@ -201,6 +201,38 @@ Note that `#count` is calculated automatically based on the number of batches in
...
+    # Generates an array like
+    # [(1..1_000_000),(1_000_001..2_000_000),(2_000_001..3_000_000), ...]
+    def collection
+      ranges = []
+      100.times do |i|
+        range_start = (1_000_000 * i) + 1
+        range_end = 1_000_000 * (i + 1)
+        ranges << (range_start..range_end)
+      end
+      ranges
```
nvasilevski, line 220: Wonder if we should consider recommending building "static" collections in a class-level memoized instance variable since we don't expect the collection to change between executions. It could also be a constant but there is no reason to slow down the boot time 

```suggestion
      100.times.map do |i|
        range_start = (1_000_000 * i) + 1
        range_end = 1_000_000 * (i + 1)
        (range_start..range_end)
      end
```

README.md
```diff
@@ -201,6 +201,38 @@ Note that `#count` is calculated automatically based on the number of batches in
...
+        .where(id: range, some_column: "Some Value")
+        .in_batches { |batch| batch.update_all(some_column: "Some New Value") }
+    end
+  end
+end
+```
+
+This pattern will keep your individual queries fast, while performing updates in small batches.
+With this approach, updating or deleteing 10s of millions of records can take under an hour, even if
+the columns being accessed/changed are not indexed.
```
nvasilevski, line 234: >are not indexed.

I'm a little concerned about this claim since a relation like `where(id: 1_000_000...2_000_000, some_column: "Some Value")` should generate SQL like `where id=>1000000 and id<=2000000 and some_column="Some Value"`  and if there are plenty of records in the `id` range then database will have to apply the `some_column` in-memory

Which is not drastically different from using `Post.where(some_column: "Some Value").in_batches` as a `collection`

README.md
nvasilevski: Which also made me think that the collection can be as simple as:

`(0..100).to_a` and the `process` method could build the "current range" on-demand:

```ruby
def process(cursor_position)
        range_start = (1_000_000 * cursor_position) + 1
        range_end = 1_000_000 * (cursor_position + 1)
        range = range_start..range_end
end
```

Unlikely it will have any significant impact on performance but it should definitely help with making the collection as small as possible

README.md
csalvato: >Which is not drastically different from using `Post.where(some_column: "Some Value").in_batches` as a collection

I may be misunderstanding your point, but the difference is that when you use `Post.where(some_column: "Some Value").in_batches` in the collection, it will do a full table scan for `some_column` to build the association before batching it out.

I also thought that it would be the same as `in_batches`, but in practice, there were queries that resulted in full table scans, whereas this approach only does a scan on the ID that is the size of the range...and you can make the range as small or as large as necessary for the operation you're performing.

In our case, 1M records was fast and easy to grok, though 2-3M records could have worked in our case. In other cases, a much smaller range may be better.

README.md
csalvato: >Which also made me think that the collection can be as simple as:

I don't understand the internals well enough to understand why we want a collection that's as small as possible...especially since, practically speaking, an array of hundreds/thousands of ranges is likely not going to be a big deal compared to a collection of billions of records, but I defer to you on that as a neophyte 🙈

csalvato: Also, how embarrassing that I didn't use `map` 😂 

I've updated that bit of code to be less crappy 😂

README.md
csalvato: Wouldn't that result in a lot more queries, thus be slower?

Not a DB expert by any means, but my understanding is that `SELECT` is generally going to be a lot faster than `UPDATE`/`DELETE`. 

Using an example of a job that needs to iterate over 100M rows, the approach here would do 100 `SELECT` queries, and somewhere between 0-1000 `UPDATE`/`DELETE` queries per each SELECT. In practice on my last run, this was normally <200 `DELETE` queries per `SELECT`. Back of the envelope math:

>100 `SELECT` * 200 `DELETE`/`UPDATE` = 20,000 total `DELETE`/`UPDATE` queries

If I did that in batches over, say, 100M rows, that would be 100,000 `DELETE` queries, which are inherently slower.

Maybe I'm wrong on that. If I am, I'd be happy to update the example. Though, either way, I'd hope people will use this as inspiration and write whatever is best for their app.

etiennebarrie:
Have you tried removing the WHERE clause causing the query to underperform, while still keeping ActiveRecord relations?

```ruby
   def collection
      Post.in_batches(of: 1_000_000)
    end

    def process(batch)
      batch
        .where(some_column: "Some value")
        .update_all(some_column: "Some New Value")
    end
```

That may not be as performant as your proposal, since this will load all the ids of each batch in memory (due to the implementation of job-iteration), but it's a better first step as a solution when queries are timing out because of the collection's relation. The high-level idea is shared (move the condition from `collection` to `process` to avoid timeouts when iterating), but the implementation is also more straightforward (just move the condition from `collection` to `process`).

Also it feels like we could improve this in job-iteration, if there's no condition on the batch, we could keep only the boundaries of the range. (The reason why job-iteration plucks all the ids is precisely to avoid keeping the potentially expensive WHERE clause in all the queries related to a single batch, but if there's no condition, it could be optimized away).

csalvato:
> That may not be as performant as your proposal,

I may be mistaken here, because I was really hacking at this in the weeds and don't remember all the details of what I tried, but when I put the `in_batches` in the collection, the call to `collection` to establish the collect will result in an unbounded query, because the SQL executes immediately.

For example:

```rb
    class DeleteInvalidCycleDataTask < MaintenanceTasks::Task
      def collection
        Cycle.in_batches(of: 1_000_000)
      end

      def process(batch)
        batch.where(is_deleted: true, subject_type: "Billing::Account").delete_all
      end
    end
```

Then in console:

```rb
irb(main):005:0> DeleteInvalidCycleDataTask.new.collection

D, [2023-09-18T16:07:25.399593 #19214] DEBUG -- :   Cycle Load (1.8ms)  /*application=Billing*/ SELECT `billing_cycles`.* FROM `billing_cycles` WHERE `billing_cycles`.`is_deleted` = FALSE
```

The collection is not lazy-enumerated (though, I'd expect it to be), and queries immediately with an unbounded query.

-----------

In the end, I defer to the example you all as maintainers want to include here: all I can say is that I tried a bunch of queries and the one I posted as a suggestion in my original PR was performant and straightforward, and I think other teams can benefit from having it listed there.

I know, for example, that I plan to execute the same exact query within the next few weeks for other backfills, and it would be nice to have it in the README here, so I don't need to dig through our old PRs to find the example.

csalvato:
>What you're seeing in the console is the implementation of inspect which loads records.

This was my understanding too, but in practice, it wasn't working that way. My memory is that calling `in_batches` in `collection` was loading the whole relation.

It's possibly my memory is off, because I was hacking at various optimizations at the same time and could have come to a faulty conclusion somewhere in there.

Either way, I'm happy to update this with your example in addition to, or in place of, mine. I am hesitant to use your example because, in practice, I believe I tried it but it didn't work as expected, whereas my query is what actually ran in production and worked.

What do you prefer to give this an approval?

csalvato:
> While there's no issue with using range-based collections in your application tasks

I am rolling out some new tasks in our production environment within the next few days that test @etiennebarrie's design. I will post here with an update.

I do think that an example of how to iterate over large, unindexed sets would help a lot of people. 

So long as it's documented in a way that makes sense for the gem, I am OK with it. 

I just think it would be beneficial if that common case is covered in the docs.  I'll report back with findings here soon.

csalvato:
So here's an example:

https://billing.shopify.io/internal/maintenance_tasks/tasks/Maintenance::OneTime::BackfillContractVersionItemTypeTask

```
# typed: false
# frozen_string_literal: true

module Maintenance
  module OneTime
    class BackfillContractVersionItemTypeTask < MaintenanceTasks::Task
      def collection
        ContractVersion.in_batches(of: 1_000_000)
      end

      def process(batch)
        batch.where(item_type: "Billing::Contract").update_all(item_type: "Contract")
      end

      def count
        (ContractVersion.maximum(:id) / 1_000_000.to_f).ceil
      end
    end
  end
end
```

In this example, we are hitting query timeouts because operating on a batch of 1M is too large.

Updating the code to this:

```rb
# typed: false
# frozen_string_literal: true

module Maintenance
  module OneTime
    class BackfillContractVersionItemTypeTask < MaintenanceTasks::Task
      def collection
        ContractVersion.in_batches(of: 1_000_000)
      end

      def process(batch)
        batch.where(item_type: "Billing::Contract").in_batches.update_all(item_type: "Contract")
      end

      def count
        (ContractVersion.maximum(:id) / 1_000_000.to_f).ceil
      end
    end
  end
end
```

Also yields a similar result.

I am not sure a practical AR approach exists for this situation.

adrianna-chang-shopify:
Ah, thanks @nvasilevski ! 🤦‍♀️ 
> We could consider this to be a bug and implement an optimization similar to https://github.com/rails/rails/pull/45414 when collection is a all-records collection

This makes sense to me -- not necessarily a bug, but it does make sense to construct a relation that avoids using `WHERE IN ()` for large batches, since that was provably faster for Rails. (I think this is essentially the same optimization @etiennebarrie was suggesting, where we'd change job-iteration to use a range-based clause in the query instead of plucking all of the ids).

I'm wondering if smaller batch sizes in `#collection` would be a good alternative though. It _does_ mean more SELECT queries, but the actual "batches" being updated would be the same as the example from this PR (assuming the default batch size is being used there), so presumably the same number of UPDATE queries. Might make the task slower overall, but I think we'd avoid timeout issues that way.

gmcgibbon:
@csalvato are you still interested in working on this?

PR #909, opened 2023-10-18
Fix test in Task with params example

tlatsas:
Fix example test in readme.


nvasilevski:
This is not wrong but I think it's a bit unusual to define an attr reader for the test and most likely we intended to always use `@task` across the whole test

cc: @gmcgibbon I think you added this example, just wanted to make sure that it was a simple typo and the intention was to use the `@task`  instance variable directly

tlatsas:
No worries! Updated

PR #911, opened 2023-10-29
Add a no-collection option for rspec spec template

Martin-Alexander:
When you generate a no-collection task for an application using RSpec the resultant spec file assumes that the task is for collections.

Running `rails g maintenance_tasks:task sing_songs --no-collection`

Before 

```ruby
# frozen_string_literal: true

require "rails_helper"

module Maintenance
  RSpec.describe SingSongsTask do
    describe "#process" do
      subject(:process) { described_class.process(element) }
      let(:element) {
        # Object to be processed in a single iteration of this task
      }
      pending "add some examples to (or delete) #{__FILE__}"
    end
  end
end
```

After 

```ruby
# frozen_string_literal: true

require "rails_helper"

module Maintenance
  RSpec.describe SingSongsTask do
    describe "#process" do
      subject(:process) { described_class.process }
      pending "add some examples to (or delete) #{__FILE__}"
    end
  end
end
```


Martin-Alexander:
I have signed the CLA!

nvasilevski (approved):
Thank you!

nvasilevski:
CI failures are unrelated so let's merge it since the change is extremely safe.

PR #914, opened 2023-10-31
Introduce `list` command in CLI

davidstosik:
This is an alternative to https://github.com/Shopify/maintenance_tasks/pull/913, which I actually prefer.

A couple months ago, in https://github.com/Shopify/maintenance_tasks/pull/877, I did some weird stuff in order to defer task loading when it's actually necessary. The issue was that we needed the list of all tasks to set the long description of the `perform` command.

I realized that it might actually allow us to keep code more standard and simple, if instead of that, we provided a `list` command that loads and lists all available tasks.

## Pros

- The CLI still doesn't need to load all tasks at "boot". (the original goal of #877)
- We don't need to use meta-programming to `define_singleton_method :long_description`, which relies on not-so-public APIs, and as shown by the release of Thor 1.3.0, is brittle.
- The output of `help perform` does not flood the terminal anymore. If I ran `maintenance_tasks help perform`, chances are I wanted to see what it does, parameters it takes and its syntax. These things were pushed off screen when there are a lot of tasks available.
- The `list` command is purely made to output a list of tasks, no indentation, no empty lines in between. It's more favorable to exporting the list to a text file or grepping for something.
- It works the same for the new Thor 1.3.0 and older versions. No weird version switching.

## Cons

- I can only think of one: CLI behaviour changes: `maintenance_tasks help perform` will not return the task list anymore. Anyone relying on that will have to update what they do.

I like this approach better than the one I had first which was trying to work around the issue in complicated ways.

Merging this will close #913 as it'll be unnecessary.

### DIFF
--- a/test/lib/maintenance_tasks/cli_test.rb
+++ b/test/lib/maintenance_tasks/cli_test.rb
@@ -86,17 +86,14 @@ module MaintenanceTasks
     end
 
     test "`list` loads all tasks and displays them" do
-      Task
-        .expects(:load_all)
-        .at_least_once
-        .returns([stub(name: "Task1"), stub(name: "Task2")])
+      Task.expects(:load_all).returns([stub(name: "Task1"), stub(name: "Task2")])
 
       expected_output = <<~OUTPUT
         Task1
         Task2
       OUTPUT
 
-      assert_output(Regexp.union(expected_output)) do
+      assert_output(expected_output) do
         CLI.start(["list"])
       end
     end
--- a/test/lib/maintenance_tasks/cli_test.rb
+++ b/test/lib/maintenance_tasks/cli_test.rb
@@ -102,7 +102,6 @@ module MaintenanceTasks
       OUTPUT
 
       assert_output(Regexp.union(expected_output)) do
-        Thor::Base.shell.any_instance.stubs(:terminal_width).returns(200)
         CLI.start(["help", "perform"])
       end
     end
--- a/test/lib/maintenance_tasks/cli_test.rb
+++ b/test/lib/maintenance_tasks/cli_test.rb
@@ -85,18 +85,6 @@ module MaintenanceTasks
       end
     end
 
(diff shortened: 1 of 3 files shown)


test/lib/maintenance_tasks/cli_test.rb
```diff
@@ -85,25 +85,31 @@ class CLITest < ActiveSupport::TestCase
...
+        `maintenance_tasks perform` will run the Maintenance Task specified by the [TASK NAME] argument.
+
+        Use `maintenance_tasks list` to get a list of all available tasks.
+      OUTPUT
+
+      assert_output(Regexp.union(expected_output)) do
+        CLI.start(["help", "perform"])
+      end
+    end
+
```
davidstosik, line 99: This test is probably unnecessary. (It tests Thor's standard behaviour, not anything related to maintenance_tasks' code.)

Remove?

test/test_helper.rb
```diff
@@ -3,6 +3,9 @@
 # Configure Rails Environment
 ENV["RAILS_ENV"] = "test"
 
+# Don't wrap output in tests
+ENV["THOR_COLUMNS"] = "200"
```
davidstosik, line 7: This is more resilient than the old `Thor::Base.shell.any_instance.stubs(:terminal_width).returns(200)`, since `THOR_COLUMNS` has been around for more than 10 years, and it still works with Thor 1.3.0.

nvasilevski (approved):
I totally support the idea of not listing all available tasks in the help message for `perform` 
Thank you!

test/lib/maintenance_tasks/cli_test.rb
adrianna-chang-shopify: Is this comment about the `help perform` test or the `list` one? I'd prefer to keep the `list` one, but fine with dropping the `help perform` one since we're matching on a simple string.

adrianna-chang-shopify (approved):
LGTM! 🚀

PR #918, opened 2023-11-09
Ensure that maintenance tasks have a default max runtime

simbasdad:
Prior to Shopify/job-iteration#240, the maximum runtime of an iteration was determined by reading `JobIteration.max_job_runtime` while the job was running.  After this change, the value of that variable is cached with the job when the class is loaded.

The maintenance_tasks engine sets `JobIteration.max_job_runtime` to 5 minutes in an `after_initialize` block, unless it is already configured.  Prior to the aforementioned PR, this was fine because the value was read at runtime.  However, in staging and production environments where eager loading is used, `MaintenanceTasks::TaskJob` is loaded before the `after_initialize` block is run.  If an application did not explicitly configure `JobIteration.max_job_runtime`, `MaintenanceTasks::TaskJob` will use the default value of `nil` provided by the job-iteration gem.  This means maintenance tasks will never be pre-empted.

This change explicitly sets the `job_iteration_max_job_runtime` class attribute on `MaintenanceTasks::TaskJob`.  If `JobIteration.max_job_runtime` has been configured, it uses that, otherwise, it defaults to 5 minutes.

I did not remove the `after_initialize` block, as applications that are not using eager loading may depend on the existing behaviour.

### DIFF
--- a/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
+++ b/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
@@ -9,6 +9,8 @@ module MaintenanceTasks
     include JobIteration::Iteration
 
     included do
+      self.job_iteration_max_job_runtime = JobIteration.max_job_runtime || 5.minutes
+
       before_perform(:before_perform)
 
       on_start(:on_start)
--- a/test/jobs/maintenance_tasks/task_job_test.rb
+++ b/test/jobs/maintenance_tasks/task_job_test.rb
@@ -617,5 +617,9 @@ module MaintenanceTasks
 
       assert_equal 2, @run.reload.tick_total
     end
+
+    test "job_iteration_max_job_runtime defaults to 5 minutes" do
+      assert_equal 5.minutes, MaintenanceTasks::TaskJob.job_iteration_max_job_runtime
+    end
   end
 end


adrianna-chang-shopify (commented):
Hey @simbasdad ! Thanks for catching this -- I'm sure it must've been a bit tricky to nail down why tasks weren't pausing to that change in job-iteration 😩 

What do you think about setting the value inside the `after_initialize` block instead? I believe something like this would work:

```ruby
# lib/maintenance_tasks/engine.rb
config.after_initialize do
  JobIteration.max_job_runtime ||= 5.minutes
  MaintenanceTasks::TaskJob.job_iteration_max_job_runtime ||= JobIteration.max_job_runtime
end
```
This way, we keep our "default" logic together, and we we can rely directly on the value of `JobIteration.max_job_runtime` instead of having that magic "5 minutes" default encoded in multiple places.

adrianna-chang-shopify:
@simbasdad I can tweak and merge this PR later today, or early next week at the latest! Thanks for bringing this up, though ❤️

app/jobs/concerns/maintenance_tasks/task_job_concern.rb
```diff
@@ -9,6 +9,8 @@ module TaskJobConcern
     include JobIteration::Iteration
 
     included do
+      self.job_iteration_max_job_runtime = JobIteration.max_job_runtime || 5.minutes
```
sambostock, line 12: This will still result in us reading the config at the time the concern is included, which may be before the host application sets the value. I think it would make sense to make it dynamic by using a method instead.

```diff
-    included do
+    included do |base|
```

```suggestion
      class << base
        def job_iteration_max_job_runtime = JobIteration.max_job_runtime || 5.minutes
      end
```

although maybe that could be defined in `class_methods` instead... 🤔 

---

From there, I think it would make sense to deprecate, and eventually remove

```ruby
# lib/maintenance_tasks/engine.rb
config.after_initialize do
  JobIteration.max_job_runtime ||= 5.minutes
```

because _really_ this gem shouldn't be setting a global config affecting other jobs (acknowledging the per-job option didn't exist at the time that was added).

We could also make that `5.minutes` configurable as `MaintenanceTasks.max_job_runtime`, if we wanted, and just have `5.minutes` as the default, but that's not a big deal. 🤷

PR #921, opened 2023-11-16
Allow force pause, when stuck pausing

Schwad:
## Background

If your pause action dies, it can get locked on "Pausing". The user, short of hacking the console or cancelling the whole task, doesn't have recourse.

Whilst in "Pausing" state, this creates a "Force pause" action that called `paused!` on the run.

image
image

## For Reviewers

Should I pair this with some further tests, is there anything I'm not considering?

### DIFF
--- a/app/controllers/maintenance_tasks/runs_controller.rb
+++ b/app/controllers/maintenance_tasks/runs_controller.rb
@@ -35,6 +35,14 @@ module MaintenanceTasks
       redirect_to(task_path(@run.task_name), alert: error.message)
     end
 
+    # Forces paused state to be set, used if stuck
+    def force_pause
+      @run.paused!
+      redirect_to(task_path(@run.task_name))
+    rescue ActiveRecord::RecordInvalid => error
+      redirect_to(task_path(@run.task_name), alert: error.message)
+    end
+
     # Updates a Run status to cancelling.
     def cancel
       @run.cancel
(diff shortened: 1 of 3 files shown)


app/controllers/maintenance_tasks/runs_controller.rb
```diff
@@ -35,6 +35,14 @@ def pause
       redirect_to(task_path(@run.task_name), alert: error.message)
     end
 
+    # Forces paused state to be set, used if stuck
+    def force_pause
```
adrianna-chang-shopify, line 39: What do you think about keeping the same endpoint, and adding a `pause` method on `MaintenanceTasks::Run` (similar to `#cancel`) that handles either `pausing!` the run or marking it as `paused` in the case where it's stuck?

adrianna-chang-shopify (commented):
With cancelling, we make users wait 5 minutes before being able to force cancelling, so that they give the job a reasonable amount of time to shut down. Should we do something similar for pausing?

Tests would be great -- if we move the logic to the Run class it might be easier to unit test. Can we also add a note to the README? ❤️

test/system/maintenance_tasks/runs_test.rb
```diff
@@ -162,7 +162,6 @@ class RunsTest < ApplicationSystemTestCase
 
       assert_text "Cancelling…"
       refute_button "Cancel"
-
```
adrianna-chang-shopify, line 165: Nit: I think this was probably removed accidentally, but let's keep this line between the assertions and the `travel` line.

adrianna-chang-shopify (approved):
LGTM, thanks Nick!

etiennebarrie:
Sorry one other thing: https://github.com/Shopify/maintenance_tasks/blob/0f8e684a0332cc9ed471a4660e8e85a25eaff204/.github/CONTRIBUTING.md?plain=1#L28

I've pushed forced on main with a squash.

PR #924, opened 2023-12-03
Pass run instance to task

shalvah:
This PR adds a `.run` property to the task object, allowing the task to access information about the  current run.

Some example use cases for this:
- Richer logs: automatically tagging logs emitted from the task with the run ID
- Richer error context: adding the run ID as a tag for better grouping/filtering
- Enriching the run `metadata` from within the task
- Persisting task information (for instance, marking an item as "failed" without crashing and interrupting the whole task)

It shouldn't be a breaking change, but I've wrapped it in a guard that first checks if the method exists.

Also, arguments are assigned afterwards, so if the task defines an attribute named `run`, it isn't overriden.


olivier-thatch:
Any chance this could be reopened and considered for merging? We could certainly use this feature :)

PR #925, opened 2023-12-03
Render a dropdown list when an inclusion validator is used

shalvah:
This allows you to write:

```rb
attribute :limit, :integer
attribute :type, :string

validates :limit, inclusion: [4, 6, 8]
validates_inclusion_of :type, in: ['new', 'old']
```

And get:

image

instead of:

image

Reduces the chance of making mistakes, saves work, serves as documentation for unfamiliar users (tells them what options are available).


app/helpers/maintenance_tasks/tasks_helper.rb
```diff
@@ -103,6 +103,16 @@ def csv_file_download_path(run)
 
     # Return the appropriate field tag for the parameter
     def parameter_field(form_builder, parameter_name)
+      # If the parameter has a `validates_inclusion_in` parameter,
+      # generate a dropdown list of options
```
adrianna-chang-shopify, line 107:
```suggestion
      # If the parameter has a `validates_inclusion_of` validator,
      # generate a dropdown list of options
```
Nit, can we also move this above the method definition so it becomes part of the docs?

shalvah:
FYI, tests are currently failing because of a ChromeDriver change which broke Capybara (https://github.com/teamcapybara/capybara/issues/2725). They've fixed it, but not released yet, so I guess we have to wait a bit. This affects other PRs in this repo too.

etiennebarrie:
The Selenium issues were resolved in #936.

shalvah:
Test added ✅ .

Also, could someone help me take a look at https://github.com/Shopify/maintenance_tasks/pull/924 ? 🙏 I've signed the CLA, but the CI continually fails with:

> Error: Although you appear to have the correct authorization credentials, the `Shopify` organization has an IP allow list enabled, and your IP address is not permitted to access this resource.

PR #934, opened 2023-12-13
Add Rails 7.1 to CI matrix

nvasilevski:
Rails 7.1 and edge are not the same anymore and we should be explicitly testing against it


adrianna-chang-shopify (approved):

etiennebarrie:
We had the same where we were testing for 7.0 in two gemfiles https://github.com/Shopify/maintenance_tasks/pull/900#pullrequestreview-1672437119

PR #937, opened 2023-12-19
Run CI on a schedule

etiennebarrie:
This will allow us to catch issues with gem upgrades early, hopefully before someone writes a PR and all the jobs fail except the default Gemfile one (with the locked versions).

Due to the way GitHub Actions workflow notifications work, I'll be the one to receive notifications (until someone else edits the `cron` line):

> Notifications for scheduled workflows are sent to the user who initially created the workflow. If a different user updates the cron syntax in the workflow file, subsequent notifications will be sent to that user instead. If a scheduled workflow is disabled and then re-enabled, notifications will be sent to the user who re-enabled the workflow rather than the user who last modified the cron syntax.
> https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/notifications-for-workflow-runs

 If anyone else cares about this, we could automatically create an issue when the scheduled workflow run fails.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,14 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  push:
+  pull_request:
+  schedule:
+  - cron: "42 23 * * 0" # Run CI on Sundays at 23:42 UTC
+
+permissions:
+  contents: read # actions/checkout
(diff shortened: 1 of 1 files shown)


gmcgibbon (approved):
@etiennebarrie This is good, but can you look into creating an issue instead, so that more folks can see potential failures?

PR #941, opened 2023-12-21
Show Run#id in the web UI

shouichi:
It makes correlating runs and errors from an error reporting system easier.

### DIFF
--- a/app/views/maintenance_tasks/runs/_run.html.erb
+++ b/app/views/maintenance_tasks/runs/_run.html.erb
@@ -2,6 +2,7 @@
   <h5 class="title is-5">
     <%= time_tag run.created_at, title: run.created_at.utc.iso8601 %>
     <%= status_tag run.status %>
+    <span class="is-pulled-right" title="Run ID">#<%= run.id %></span>
   </h5>
 
   <%= progress run %>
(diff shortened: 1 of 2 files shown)


nvasilevski:
Hey, I believe a few assertions need to be tweaked

@Shopify/rails May I ask for opinions on this? I believe we are a little hesitant about exposing some internal concept of maintenance tasks on UI but I also want to recognize that internally we might have better debugging tooling (splunk links) that in most cases should remove the need to know the run id. But I can imagine that having more details about this particular run may help with debugging if all you have is access to the db and perhaps raw job worker logs

etiennebarrie:
I've wanted to add the id as a `dom_id` for a while. It does require to open up the developer tools to inspect the page, but since the gem is meant for developers, I think displaying the id directly wouldn't be a bad idea. It needs to be a bit clearer that it's an id, and it could be a bit nicer than just outputting it next to the date.

image


Maybe as a display: float on the right-hand side of the card (`.box` in Bulma)? and with a `title` attribute explaining it's an ID?

shouichi:
Thank you for the feedback. Do you mean displaying IDs on the right side?

image

etiennebarrie (approved):
Thanks! Let's see if anyone else has feedback.

adrianna-chang-shopify (approved):

PR #942, opened 2023-12-21
Make error handler proc be called with job as self

shalvah:
This is the same as https://github.com/Shopify/maintenance_tasks/pull/782, but updated on main, so CI passes.

Having access to the job which errorred enables richer context for error reporting. The current `task_context` and `errored_element` are helpful, but not enough. For instance, in our case, we need access to the job so we can record and provide correlation links in our error reporter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -719,14 +719,14 @@ you can define an error handler:
 ```ruby
 # config/initializers/maintenance_tasks.rb
 
-MaintenanceTasks.error_handler = ->(error, task_context, _errored_element) do
+MaintenanceTasks.error_handler = ->(error, task_context, errored_element, task_job) do
   Bugsnag.notify(error) do |notification|
     notification.add_metadata(:task, task_context)
   end
 end
 ```
 
-The error handler should be a lambda that accepts three arguments:
+The error handler should be a lambda that accepts four arguments:
 
 * `error`: The exception that was raised.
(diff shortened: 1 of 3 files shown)


gmcgibbon:
- Looks like this needs to be rebased
- Please sign the CLA
- It might make more sense to take the task as an argument instead of binding the proc, does that work for your use-case?

shalvah:
- Rebased ✅
- Already signed
- Yep, that would work, too. I can switch to that if you prefer?

shalvah:
I think I kinda prefer the`instance_exec` approach, because it lines up nicely with how things are done for `metadata`, and it's a fully transparent change. But I've added a commit which switches to the task as argument. I can revert it if we decide on `instance_exec`.

adrianna-chang-shopify (commented):
Hm, I'm still on the fence about these changes, although I could be convinced. The job instance exposes a lot of maintenance task internals -- the `MaintenanceTasks::Run`, object, the `MaintenanceTasks::Ticker`, Job Iteration internals...

I agree that the error handler in its current state doesn't provide a lot of info to go off of though. Is there other information we could expose as part of the `task_context` that would be sufficient for your use case with error reporting? The Job ID, other metadata from the `Run`, etc.? What data do you need to provide correlation links?

PR #944, opened 2023-12-26
Custom enumerators

rianmcguire:
This is based on the work in https://github.com/Shopify/maintenance_tasks/pull/326 and the design suggested by @adrianna-chang-shopify in [this comment](https://github.com/Shopify/maintenance_tasks/pull/326#pullrequestreview-613240141).

It allows tasks to define their own custom enumerators, in order to support use cases like iteration over external API resources, or alternative database querying strategies.

It does this by exposing an overridable `enumerator_builder` method on `Task`, whose default implementation is the existing `collection` type inspection logic.

Fixes https://github.com/Shopify/maintenance_tasks/issues/207

### DIFF
--- a/README.md
+++ b/README.md
@@ -299,6 +299,35 @@ module Maintenance
 end
 ```
 
+### Tasks with Custom Enumerators
+
+If you have a special use case requiring iteration over an unsupported
+collection type, such as external resources fetched from some API, you can
+implement the `enumerator_builder(cursor:)` method in your task.
+
+This method should return an `Enumerator`, yielding pairs of
+`[item, cursor]`. Maintenance Tasks takes care of persisting the current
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -299,6 +299,49 @@ module Maintenance
...
+      end
+    end
+
+    def process(post)
+      Post.create!(post)
+    end
+  end
+end
+
+```
```
bdewater, line 343: The example repeats what's written at   https://github.com/Shopify/job-iteration/blob/main/guides/custom-enumerator.md - should we link to it instead?

rianmcguire:
> Please squash your commits and let's give Sam co-author please. Thank you!

Done! Thanks for the feedback and reviews everyone.

PR #949, opened 2024-01-12
Add `MaintenanceTasks.stuck_task_duration ` configuration

nvasilevski:
Having `STUCK_TASK_TIMEOUT` as a constant is not helpful since the actual value will be highly dependent on the job infrastructure of the applications. Simple setups may consider the task being stuck after no updates for ~ 1 minute while more complex and distributed systems with job retries and backoffs may legitimately prefer higher threshold to account for natural delays in a distributed system

It's a breaking change for applications that relied on  `STUCK_TASK_TIMEOUT` though the fix is going to be pretty straightforward. 

### What reviewers should focus on

Is `MaintenanceTasks` module a right place for the config to live? One common alternative would be to define it on the `Task` class which will allow redefining it on per-task basis. But I don't think this would make sense and the behavior is job-infra specific rather than task-specific.

### DIFF
--- a/app/models/maintenance_tasks/run.rb
+++ b/app/models/maintenance_tasks/run.rb
@@ -32,7 +32,6 @@ module MaintenanceTasks
       :cancelled,
     ]
     COMPLETED_STATUSES = [:succeeded, :errored, :cancelled]
-    STUCK_TASK_TIMEOUT = 5.minutes
 
     enum status: STATUSES.to_h { |status| [status, status.to_s] }
 
@@ -342,7 +341,7 @@ module MaintenanceTasks
     #
     # @return [Boolean] whether the Run is stuck.
     def stuck?
-      (cancelling? || pausing?) && updated_at <= STUCK_TASK_TIMEOUT.ago
+      (cancelling? || pausing?) && updated_at <= MaintenanceTasks.stuck_task_duration.ago
     end
 
     # Performs validation on the task_name attribute.
--- a/lib/maintenance_tasks.rb
+++ b/lib/maintenance_tasks.rb
@@ -89,4 +89,11 @@ module MaintenanceTasks
   #
   #   @return [Proc] generates a hash containing the metadata to be stored on the Run
   mattr_accessor :metadata, default: nil
+
+  # @!attribute stuck_task_duration
+  #  @scope class
+  #  The duration after which a task is considered stuck and can be force cancelled.
+  #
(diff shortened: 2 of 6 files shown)


app/models/maintenance_tasks/run.rb
```diff
@@ -32,7 +32,6 @@ class Run < ApplicationRecord
       :cancelled,
     ]
     COMPLETED_STATUSES = [:succeeded, :errored, :cancelled]
-    STUCK_TASK_TIMEOUT = 5.minutes
```
adrianna-chang-shopify, line 35: I was wondering about using [`#deprecate_constant`](https://www.rubydoc.info/stdlib/core/Module:deprecate_constant), but `MaintenanceTasks::Run` is `@api private`, so I don't think we need to deprecate 👍 (Apps shouldn't be referencing this).

lib/maintenance_tasks.rb
```diff
@@ -89,4 +89,11 @@ module MaintenanceTasks
   #
   #   @return [Proc] generates a hash containing the metadata to be stored on the Run
   mattr_accessor :metadata, default: nil
+
+  # @!attribute task_stuck_threshold
+  #  @scope class
+  #  The threshold in seconds after which a task is considered stuck.
+  #
+  #  @return [Integer] the threshold in seconds after which a task is considered stuck.
+  mattr_accessor :task_stuck_threshold, default: 5.minutes
```
adrianna-chang-shopify, line 98: I don't think it can be an integer since we call `#ago` on it in `#stuck?` -- I think it needs to be an `AS::Duration`.

```suggestion
  #  The duration after which a task is considered stuck and can be force cancelled.
  #
  #  @return [ActiveSupport::Duration, Numeric] duration after which a task is considered stuck.
  mattr_accessor :task_stuck_threshold, default: 5.minutes
```

adrianna-chang-shopify (commented):
I'm okay with this as a feature. I'm not a huge fan of the config name, but can't think of anything more appropriate right now. Maybe someone else on the team has ideas? `threshold` seems a bit verbose 😅 Maybe even `stuck_task_duration` or `stuck_task_period` to imply it's a duration?

nvasilevski:
@adrianna-chang-shopify 

I like `stuck_task_duration`, let me make a change

etiennebarrie (approved):

adrianna-chang-shopify (approved):
LGTM! Can we add a section to the README under https://github.com/Shopify/maintenance_tasks?tab=readme-ov-file#configuring-the-gem to document this option before we merge?

PR #951, opened 2024-01-19
Add stale action

gmcgibbon:
Cleans up old issues that have become stale / inactive. We have a few open issues that are quite old. Instead of going through them, I'd like to use a robot to mark them as stale, and if the issuer is still waiting, they will remove the label, and we can get back to them.


.github/workflows/stale.yml
```diff
@@ -0,0 +1,24 @@
...
+    - cron: "30 1 * * *""
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
+          days-before-issue-stale: 60
+          exempt-issue-labels: "Blocked"
+          days-before-issue-close: 7
```
nvasilevski, line 16: Just a personal opinion but perhaps this should be closer to `14` days just to account for long holiday breaks and vacations

gmcgibbon: 👍

Issue #747, opened 2023-01-18
Support for ActiveSupport Notifications


elfassy:
Instead of using callbacks, should we add support for notifications instead? It would make it more consistent with the[ rest of rails ](https://guides.rubyonrails.org/active_support_instrumentation.html)to subscribe to these events

https://api.rubyonrails.org/classes/ActiveSupport/Notifications.html

nvasilevski:
>That doesn't seem to be what's hinted at in the readme

This is just a way to mass-define a callback. It's fair that in this case the callback won't go away with the deletion of the task but execution wise it still stays specific to a given task. So a callback is something that is attached to a task when it comes to an execution and a notification subscriber is something that exists in an application and just listens for events. Ultimately in an application there could be two maintenance tasks frameworks that issue the same notifications on similar events while callbacks will have to be duplicated in both of them even if it's going to be a single meta-programming solution that evals code in parent classes. This is why I don't consider notifications to be a complete substitution to callbacks. We should have both

PR #961, opened 2024-02-02
send active support notifications on run events

elfassy:
closes https://github.com/Shopify/maintenance_tasks/issues/747

# Why 

Callbacks on the task are not meant to be used for instrumentation across all tasks. `Task` callbacks also don't have access to the `Run` object, preventing this logic from accessing the metadata or other information stored on the run. This makes it difficult to set notifications for status changes, for instance to send slack messages when a job is completed

# What

Send [active support notifications](https://api.rubyonrails.org/classes/ActiveSupport/Notifications.html) on the Run status changes

```
      maintenance_tasks.enqueued,    # The task has been enqueued by the user.
      maintenance_tasks.succeeded,   # The task finished without error.
      maintenance_tasks.cancelled,   # The user explicitly halted the task's execution.
      maintenance_tasks.paused,      # The task was paused in the middle of the run by the user.
      maintenance_tasks.errored,     # The task code produced an unhandled exception.
 ```
 
 # Example
 
 ```ruby
 ActiveSupport::Notifications.subscribe("maintenance_tasks.succeeded") do |*args|
        payload = args.last
        run = payload[:run]
        run.task_name
        run.arguments
        run.metadata
end

 ActiveSupport::Notifications.subscribe("maintenance_tasks.errored") do |*args|
        payload = args.last
        run = payload[:run]
        run.task_name
        run.arguments
        run.error_message
        run.error_class
end

# or

class MaintenanceTasksInstrumenter < ActiveSupport::Subscriber
  attach_to :maintenance_tasks
  
  def enqueued(event)
    run = event.payload[:run]
    SlackNotifier.broadcast(SLACK_CHANNEL, "Job #{run.task_name} was started by #{run.metadata[:user_email]}} with arguments #{run.arguments.to_s.truncate(255)}")
  end
  
  def errored(event)
    run = event.payload[:run]
    SlackNotifier.broadcast(SLACK_CHANNEL, "Job #{run.task_name} halted with an error")
  end
  
  def succeeded(event)
    run = event.payload[:run]
    # do something anytime a maintenance task succeeds
    SlackNotifier.broadcast(SLACK_CHANNEL, "Job #{run.task_name} completed successfully")
  end
  
  ...
end
 ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -479,6 +479,8 @@ Usage example:
   time_running = payload[:time_running]
   started_at = payload[:started_at]
   ended_at = payload[:ended_at]
+rescue => e
+  Rails.logger.error(e)
 end
 
 ActiveSupport::Notifications.subscribe("errored.maintenance_tasks") do |*, payload|
@@ -487,6 +489,8 @@ ActiveSupport::Notifications.subscribe("errored.maintenance_tasks") do |*, paylo
   error_message = error[:message]
   error_class = error[:class]
   error_backtrace = error[:backtrace]
+rescue => e
+  Rails.logger.error(e)
 end
 
 # or
@@ -501,6 +505,8 @@ class MaintenanceTasksInstrumenter < ActiveSupport::Subscriber
 
     SlackNotifier.broadcast(SLACK_CHANNEL,
       "Job #{task_name} was started by #{metadata[:user_email]}} with arguments #{arguments.to_s.truncate(255)}")
+  rescue => e
+    Rails.logger.error(e)
   end
 end
 ```
--- a/README.md
+++ b/README.md
@@ -449,6 +449,40 @@ module Maintenance
     def process(post)
       post.update!(content: "updated content")
     end
+
+### Subscribing to instrumentation events
+
+If you are interested in actioning a specific task event, please refer to the `Task Callbacks` section below. However, if you want to subscribe to all events, irrespective of the task, you can use the following Active Support notifications:
+
+```ruby
+maintenance_tasks.enqueued    # This event is published when a task has been enqueued by the user.
+maintenance_tasks.succeeded   # This event is published when a task has finished without any errors.
+maintenance_tasks.cancelled   # This event is published when the user explicitly halts the execution of a task.
+maintenance_tasks.paused      # This event is published when a task is paused by the user in the middle of its run.
+maintenance_tasks.errored     # This event is published when the task's code produces an unhandled exception.
+```
+
+These notifications offer a way to monitor the lifecycle of maintenance tasks in your application.
+
+Usage example:
+
+ ```ruby
+ ActiveSupport::Notifications.subscribe("maintenance_tasks.enqueued") do |*args|
+  payload = args.last
+  run = payload[:run]
+  run.task_name
+  run.arguments
+  run.metadata
+end
+
+# or
+
+class MaintenanceTasksInstrumenter < ActiveSupport::Subscriber
+  attach_to :maintenance_tasks
+
+  def enqueued(event)
+    run = event.payload[:run]
+    SlackNotifier.broadcast(SLACK_CHANNEL, "Job #{run.task_name} was started by #{run.metadata[:user_email]}} with arguments #{run.arguments.to_s.truncate(255)}")
   end
 end
 ```
--- a/README.md
+++ b/README.md
@@ -458,11 +458,11 @@ end
 If you are interested in actioning a specific task event, please refer to the [Using Task Callbacks](#using-task-callbacks) section below. However, if you want to subscribe to all events, irrespective of the task, you can use the following Active Support notifications:
 
 ```ruby
-maintenance_tasks.enqueued    # This event is published when a task has been enqueued by the user.
-maintenance_tasks.succeeded   # This event is published when a task has finished without any errors.
-maintenance_tasks.cancelled   # This event is published when the user explicitly halts the execution of a task.
-maintenance_tasks.paused      # This event is published when a task is paused by the user in the middle of its run.
-maintenance_tasks.errored     # This event is published when the task's code produces an unhandled exception.
+enqueued.maintenance_tasks    # This event is published when a task has been enqueued by the user.
+succeeded.maintenance_tasks   # This event is published when a task has finished without any errors.
+cancelled.maintenance_tasks   # This event is published when the user explicitly halts the execution of a task.
+paused.maintenance_tasks      # This event is published when a task is paused by the user in the middle of its run.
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: test/models/maintenance_tasks/run_test.rb [hunk 1]
```diff
@@ -690,6 +704,16 @@ class RunTest < ActiveSupport::TestCase
 
     private
 
+    def assert_notification_for(notification, expected_payload)
+      data = nil
+      subscription = ActiveSupport::Notifications.subscribe(notification) do |_name, _start, _finish, _id, payload|
```
fridaynosaur, line 709:
```suggestion
      notification = ActiveSupport::Notifications.subscribe(notification) do |_name, _start, _finish, _id, payload|
```

app/models/maintenance_tasks/run.rb
```diff
@@ -448,6 +450,15 @@ def task
 
     private
 
+    def instrument_status_change
+      return unless status && status != previous_changes["status"].first
+      return if status[/ing\z/] # pausing vs paused, cancelling vs cancelled
```
fridaynosaur, line 455: Not opinionated much about this - would also skip `running` - just wondering if it's worth to exclude them here rather than filtering these out if needed when subscribing to them

app/models/maintenance_tasks/run.rb
```diff
@@ -448,6 +450,15 @@ def task
 
     private
 
+    def instrument_status_change
+      return unless status && status != previous_changes["status"].first
```
fridaynosaur, line 454: Just a nit, probably just me spending too much time reading multiple negative conditions:
`return if status.nil? || status == previous_changes["status"].first`

fridaynosaur (approved):

app/models/maintenance_tasks/run.rb
elfassy: I think it could get confusing as to the difference between `running` and `enqueued` (running can happen multiple times after each interruption if I'm not mistaken). I'd rather remove them for now and let someone with a valid use case add them back. IMO the effort to add them back here is less than the sum of code changes required each time someone gets confused by these states. That being said, I can be explicit about which are ignored rather than using a regex, will fix 🙌

app/models/maintenance_tasks/run.rb
elfassy: haha my brain has `return unless` == `if` hardwired. Will update 👍

fridaynosaur (approved):
Nice changes, re-read everything and it looks awesome.

test/models/maintenance_tasks/run_test.rb
```diff
@@ -690,6 +702,17 @@ class RunTest < ActiveSupport::TestCase
 
     private
 
+    def assert_notification_for(name, expected_payload)
```
robinbrandt, line 705: Small improvement: You should be able to use `ActiveSupport::Notifications.subscribed` which takes a block and unsubscribes after.

https://api.rubyonrails.org/classes/ActiveSupport/Notifications.html

README.md
```diff
@@ -413,6 +413,44 @@ to run. Since arguments are specified in the user interface via text area
...
+end
+
+# or
+
+class MaintenanceTasksInstrumenter < ActiveSupport::Subscriber
+  attach_to :maintenance_tasks
+
+  def enqueued(event)
+    run = event.payload[:run]
+    SlackNotifier.broadcast(SLACK_CHANNEL, "Job #{run.task_name} was started by #{run.metadata[:user_email]}} with arguments #{run.arguments.to_s.truncate(255)}")
```
ryanseys, line 448: why truncate in the example?

ryanseys (approved):
LGTM, couple questions

app/models/maintenance_tasks/run.rb
```diff
@@ -452,6 +454,15 @@ def task
 
     private
 
+    def instrument_status_change
+      return if status.nil? || !status_previously_changed?
+      return if status.in?(["running", "pausing", "cancelling", "interrupted"])
+
+      ActiveSupport::Notifications.instrument("maintenance_tasks.#{status}", run: self)
+    rescue
```
nvasilevski, line 462: I understand the idea to ensure that issuing a notification will not result in the whole flow being halted but I think this `rescue` overly defensive. Realistically we don't expect anything to go wrong in this method but `rescue` will swallow all the problems including even `NoMethodError`

app/models/maintenance_tasks/run.rb
```diff
@@ -452,6 +454,15 @@ def task
 
     private
 
+    def instrument_status_change
+      return if status.nil? || !status_previously_changed?
+      return if status.in?(["running", "pausing", "cancelling", "interrupted"])
+
+      ActiveSupport::Notifications.instrument("maintenance_tasks.#{status}", run: self)
```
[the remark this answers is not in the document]
elfassy: We can, was just keeping things simple. Anything I'm missing?
```
       run_id
       task_name
       arguments
       metadata
       error_message
       error_class
```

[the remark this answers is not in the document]
adrianna-chang-shopify: Do we care about some of the timestamp fields, e.g. `created_at` / `started_at` / `ended_at`? Some of the terminal statuses (succeeded / errored / cancelled) might want the `ended_at` timestamp.

The `job_id` might also be valuable -- I can imagine notifications around enqueues / interruptions might benefit from having the job id available for cross-referencing with the job infra. 

The `ticks` and `cursor` fields probably don't provide a lot of value, I'm okay with leaving those out. `status` is obviously redundant given the notification events. Backtrace seems unnecessary as well, given that's provided to the error handler.

Just making note that we should probably update the error handler to match whatever fields we decide to pass into the notification payload (not in scope for this PR, of course).

adrianna-chang-shopify (approved):
Thanks for working on this! ❤️ One small request for additional docs, and let's wait to see if anyone else on the team has feedback, but otherwise 👍

app/models/maintenance_tasks/run.rb
elfassy: 🤔 lol, I need another coffee.... this TaskJobTest passes if I don't add this `after_save`, but if I add this method it starts failing (and I'm fairly certain it's not this method making it fail, it's a validation error)

etiennebarrie: Yeah removing the rescue clause is probably what makes the most sense. If we were to keep it, we should find a way to re-raise the exception from the job after the data is correctly persisted. Maybe we could add the rescue to the readme example to show the best practice that prevents the instrumentation block from raising.

nvasilevski:
Let's squash the commits before merging please

PR #962, opened 2024-02-03
Fix dropdown options from not rendering views with proc, lamda, symbol inclusion options

WCGPR0:
Super excited about this [dropdown feature](https://github.com/Shopify/maintenance_tasks/pull/925)! -- but I noticed for example `validates_inclusion_of`'s`:in` also supports proc, lambda, or symbol, so would like to ~~extend on this feature~~ fix these types as the views aren't rendering when they are passed.

This allows you to write:
```ruby
attribute :type, :string
validates_inclusion_of :type, in: ->(_record) { ['new', 'old'] }
```

### DIFF
--- a/app/helpers/maintenance_tasks/tasks_helper.rb
+++ b/app/helpers/maintenance_tasks/tasks_helper.rb
@@ -101,16 +101,31 @@ module MaintenanceTasks
       )
     end
 
+    # Resolves values covered by the inclusion validator for a task attribute.
+    # Only Arrays are supported, option types such as:
+    # Procs, lambdas, symbols, and Range are not supported and return nil.
+    #
+    # Returned values are used to populate a dropdown list of options.
+    #
+    # @param task_class [Class<Task>] The task class for which the value needs to be resolved.
+    # @param parameter_name [String] The parameter name.
+    #
+    # @return [Array] value of the resolved inclusion option.
+    def resolve_inclusion_value(task_class, parameter_name)
+      inclusion_validator = task_class.validators_on(parameter_name).find do |validator|
+        validator.kind == :inclusion
+      end
+      return unless inclusion_validator
+
+      in_option = inclusion_validator.options[:in] || inclusion_validator.options[:within]
+      in_option if in_option.is_a?(Array)
+    end
+
     # Return the appropriate field tag for the parameter, based on its type.
     # If the parameter has a `validates_inclusion_of` validator, return a dropdown list of options instead.
     def parameter_field(form_builder, parameter_name)
-      inclusion_validator = form_builder.object.class.validators_on(parameter_name).find do |validator|
-        validator.kind == :inclusion
-      end
-
-      return form_builder.select(
-        parameter_name, inclusion_validator.options[:in], prompt: "Select a value"
-      ) if inclusion_validator
+      inclusion_values = resolve_inclusion_value(form_builder.object.class, parameter_name)
+      return form_builder.select(parameter_name, inclusion_values, prompt: "Select a value") if inclusion_values
 
       case form_builder.object.class.attribute_types[parameter_name]
       when ActiveModel::Type::Integer
--- a/test/dummy/app/tasks/maintenance/params_task.rb
+++ b/test/dummy/app/tasks/maintenance/params_task.rb
@@ -17,9 +17,22 @@ module Maintenance
     attribute :date_attr, :date
     attribute :time_attr, :time
     attribute :boolean_attr, :boolean
-    attribute :enum_attr, :integer
 
-    validates_inclusion_of :enum_attr, in: [100, 200, 300], allow_nil: true
+    # Dropdown options with supported scenarios
+    attribute :integer_dropdown_attr, :integer
+    attribute :boolean_dropdown_attr, :boolean
+
+    validates_inclusion_of :integer_dropdown_attr, in: [100, 200, 300], allow_nil: true
+    validates_inclusion_of :boolean_dropdown_attr, within: [true, false], allow_nil: true
+
+    # Dropdown options with unsupported scenarios
+    attribute :text_integer_attr, :integer
+    attribute :text_integer_attr2, :integer
+    attribute :text_integer_attr3, :integer
+
+    validates_inclusion_of :text_integer_attr, in: proc { [100, 200, 300] }, allow_nil: true
+    validates_inclusion_of :text_integer_attr2, in: :undefined_symbol, allow_nil: true
+    validates_inclusion_of :text_integer_attr3, in: (100..), allow_nil: true
 
     class << self
       attr_accessor :fast_task
--- a/test/helpers/maintenance_tasks/tasks_helper_test.rb
+++ b/test/helpers/maintenance_tasks/tasks_helper_test.rb
@@ -91,14 +91,19 @@ module MaintenanceTasks
         csv_file_download_path(run)
     end
 
+    test "#resolve_inclusion_value resolves inclusion validator for task attributes" do
+      assert_match "Select a value", markup("integer_dropdown_attr").squish
+
+      assert_match "Select a value", markup("boolean_dropdown_attr").squish
+
+      ["text_integer_attr", "text_integer_attr2", "text_integer_attr3"].each do |text_integer_attr|
+        refute_match "Select a value", markup(text_integer_attr).squish
+      end
+    end
+
     test "#parameter_field adds information about datetime fields when Time.zone_default is not set" do
       with_zone_default(nil) do
-        markup = render(inline: <<~TEMPLATE)
-          <%= fields_for(Maintenance::ParamsTask.new) do |form| %>
-            <%= parameter_field(form, 'datetime_attr') %>
-          <% end %>
-        TEMPLATE
-        assert_match "UTC", markup
+        assert_match "UTC", markup("datetime_attr")
       end
     end
 
@@ -111,12 +116,7 @@ module MaintenanceTasks
 
     test "#parameter_field adds information about datetime fields when Time.zone_default is set" do
       with_zone_default(Time.find_zone!("EST")) do # ignored
-        markup = render(inline: <<~TEMPLATE)
-          <%= fields_for(Maintenance::ParamsTask.new) do |form| %>
-            <%= parameter_field(form, 'datetime_attr') %>
-          <% end %>
-        TEMPLATE
-        assert_match Time.now.zone.to_s, markup
+        assert_match Time.now.zone.to_s, markup("datetime_attr")
       end
     end
 
@@ -143,6 +143,14 @@ module MaintenanceTasks
       yield
     ensure
       Time.zone_default = zone
+    end
+
+    def markup(attribute)
+      render(inline: <<~TEMPLATE)
+        <%= fields_for(Maintenance::ParamsTask.new) do |form| %>
+          <%= parameter_field(form, '#{attribute}') %>
+        <% end %>
+      TEMPLATE
     end
   end
 end
--- a/test/models/maintenance_tasks/task_data_show_test.rb
+++ b/test/models/maintenance_tasks/task_data_show_test.rb
@@ -87,7 +87,11 @@ module MaintenanceTasks
           "date_attr",
           "time_attr",
           "boolean_attr",
-          "enum_attr",
+          "integer_dropdown_attr",
+          "boolean_dropdown_attr",
+          "text_integer_attr",
+          "text_integer_attr2",
+          "text_integer_attr3",
         ],
         TaskDataShow.new("Maintenance::ParamsTask").parameter_names,
       )
--- a/test/system/maintenance_tasks/tasks_test.rb
+++ b/test/system/maintenance_tasks/tasks_test.rb
@@ -114,10 +114,25 @@ module MaintenanceTasks
       boolean_field = page.find_field("task[boolean_attr]")
       assert_equal("input", boolean_field.tag_name)
       assert_equal("checkbox", boolean_field[:type])
-      enum_field = page.find_field("task[enum_attr]")
-      assert_equal("select", enum_field.tag_name)
-      enum_field_options = enum_field.find_all("option").map { |option| option[:value] }
-      assert_equal(["", "100", "200", "300"], enum_field_options)
+
+      integer_dropdown_field = page.find_field("task[integer_dropdown_attr]")
+      assert_equal("select", integer_dropdown_field.tag_name)
+      assert_equal("select-one", integer_dropdown_field[:type])
+      integer_dropdown_field_options = integer_dropdown_field.find_all("option").map { |option| option[:value] }
+      assert_equal(["", "100", "200", "300"], integer_dropdown_field_options)
+
+      boolean_dropdown_field = page.find_field("task[boolean_dropdown_attr]")
+      assert_equal("select", boolean_dropdown_field.tag_name)
+      assert_equal("select-one", boolean_dropdown_field[:type])
+      boolean_dropdown_field_options = boolean_dropdown_field.find_all("option").map { |option| option[:value] }
+      assert_equal(["", "true", "false"], boolean_dropdown_field_options)
+
+      ["text_integer_attr", "text_integer_attr2", "text_integer_attr3"].each do |text_integer_attr|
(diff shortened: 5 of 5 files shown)


adrianna-chang-shopify (commented):
Thanks for contributing to the gem! 

Can we add unit tests for the new helper method please (covering support for both Rails 6 and Rails 7)?

app/helpers/maintenance_tasks/tasks_helper.rb
```diff
@@ -101,13 +101,39 @@ def csv_file_download_path(run)
...
     end
 
+    # Returns the appropriate value for a Proc, lambda, symbol, or other types.
+    #
+    # @param validator [ActiveModel::Validations::*] The validator object used to resolve the value.
+    # @param record [Task] The record for which the value needs to be resolved.
+    # @param delimiter [Object, Symbol, Proc] The delimiter used to determine the value.
+    #
+    # @return [Object] value of the delimiter.
+    def resolve_value(validator, record, delimiter)
```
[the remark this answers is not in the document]
WCGPR0: Good idea -- applied in https://github.com/Shopify/maintenance_tasks/pull/962/commits/4eec6124987b914fd8a8c3c87e6703475944b001

app/helpers/maintenance_tasks/tasks_helper.rb
```diff
@@ -101,13 +101,39 @@ def csv_file_download_path(run)
       )
     end
 
+    # Returns the appropriate value for a Proc, lambda, symbol, or other types.
+    #
+    # @param validator [ActiveModel::Validations::*] The validator object used to resolve the value.
```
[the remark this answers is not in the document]
WCGPR0: Yep - it would always be the inclusion validator, I refactored the helper method and it's arguments in https://github.com/Shopify/maintenance_tasks/pull/962/commits/4eec6124987b914fd8a8c3c87e6703475944b001

app/helpers/maintenance_tasks/tasks_helper.rb
```diff
@@ -101,13 +101,39 @@ def csv_file_download_path(run)
...
+    # @param record [Task] The record for which the value needs to be resolved.
```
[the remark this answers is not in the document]
WCGPR0: That makes sense -- applied in https://github.com/Shopify/maintenance_tasks/pull/962/commits/4eec6124987b914fd8a8c3c87e6703475944b001
(I used record because I had been looking [at this](https://github.com/rails/rails/blob/6b2a9f1213d7aa4f9a25db7640ad68ec0d8c31ce/activemodel/lib/active_model/validations/resolve_value.rb#L6))

app/helpers/maintenance_tasks/tasks_helper.rb
```diff
@@ -101,16 +101,38 @@ def csv_file_download_path(run)
...
+      inclusion_validator = object.class.validators_on(parameter_name).find do |validator|
         validator.kind == :inclusion
       end
+      return unless inclusion_validator
+
+      in_option = inclusion_validator.options[:in]
 
+      # Rails 7
+      if inclusion_validator.respond_to?(:resolve_value)
+        return inclusion_validator.resolve_value(object, in_option)
```
etiennebarrie, line 120: That's not a public API.

etiennebarrie (commented):
I think it's a bit too dangerous to call any random Proc like this.

Validators can take a bit of time e.g. a few queries, etc, it happens during `POST maintenance_tasks/runs#create`, we're fine living with this. But this is used in a view during a GET request, which is less acceptable in my opinion.

Add to the fact that we'll need to either use a private API or reimplement `validates_inclusion_of`, I'm not in favor of this change.

Currently we have a bug with this, where if you use a Proc, the view will fail, we should focus on fixing this IMO. I think we should support Enumerable except Ranges (for some ranges it uses `cover?`, that's also too much reimplementation of Rails for my taste).

app/helpers/maintenance_tasks/tasks_helper.rb
```diff
@@ -101,16 +101,33 @@ def csv_file_download_path(run)
...
+    # @return [Object] value of the resolved inclusion option.
+    def resolve_inclusion_value(object, parameter_name)
+      inclusion_validator = object.class.validators_on(parameter_name).find do |validator|
         validator.kind == :inclusion
       end
+      return unless inclusion_validator
+
+      in_option = inclusion_validator.options[:in]
 
+      return in_option.call(object) if in_option.respond_to?(:call)
```
WCGPR0, line 118: > Currently we have a bug with this, where if you use a Proc, the view will fail, we should focus on fixing this IMO

I believe this PR actually does addresses this bug, it was actually what got me looking into this first, but then I thought let me try to support all the possible options `validates_inclusion_of` offers

> I think it's a bit too dangerous to call any random Proc like this.

To address this concern, the following change will fix the bug by not rendering the dropdown altogether
```suggestion
      return if in_option.respond_to?(:call)
```

app/helpers/maintenance_tasks/tasks_helper.rb
```diff
@@ -101,16 +101,33 @@ def csv_file_download_path(run)
...
+      return object.send(in_option) if in_option.respond_to?(:to_sym)
```
WCGPR0, line 119: > I think we should support Enumerable except Ranges

How do you feel about symbols? I would like to keep the symbol support here _if possible_, simply because:
1. I really like the dropdown UI over free text
2. But I would like to be able to make an API call to get these values, ie.
```ruby
validates_inclusion_of :enum_attr, in: :enum_array, allow_nil: true

def enum_array
   # make an API call here
end
```

etiennebarrie: The problem is actually doing network calls in the context of an HTTP request.

As it is, it's not super great that we may be doing this during the Run creation, in the POST maintenance_tasks/runs#create request. Removing that would mean a worse user experience (we couldn't show the validation errors if the validation ran in the job, and we would save a run just to show the errors), so we've decided that we wouldn't complicate validations and only run them on creation (we could have had two contexts of validations, one of them would be running before the run starts, in the job, allowing more expensive validations).

But using the same validations during the GET request to get to the Task page is not something I want. In particular with the refresh it means that your API call would happen every three seconds while a Run is running.

WCGPR0: That makes sense to me, I forgot that there's a refresh. I'm not sure if memoizing the results would prevent the repeated network calls. But does that mean we want to explicitly throw an error saying we don't support this? Because the approach I'm doing to fix the proc bug was to simply not render the dropdown, and let it flow to the default textarea, if we do the same for the symbol:

```ruby
return if in_option.respond_to?(:call) || in_option.respond_to?(:to_sym)
```

There's nothing stopping the user from still doing the code as above, the validation is still there, just the results are prettified into a dropdown form.

test/dummy/app/tasks/maintenance/params_task.rb
```diff
@@ -20,6 +20,10 @@ class ParamsTask < MaintenanceTasks::Task
     attribute :enum_attr, :integer
 
     validates_inclusion_of :enum_attr, in: [100, 200, 300], allow_nil: true
+    validates_inclusion_of :enum_attr, within: [100, 200, 300], allow_nil: true
+    validates_inclusion_of :enum_attr, in: proc { [100, 200, 300] }, allow_nil: true
+    validates_inclusion_of :enum_attr, in: ->(_record) { [100, 200, 300] }, allow_nil: true
+    validates_inclusion_of :enum_attr, in: :enum_array, allow_nil: true
```
etiennebarrie, line 26: These are not super helpful as they don't really do anything. They're ignored in the view because we only look for the first `:inclusion` validator (which I think is OK overall).

test/helpers/maintenance_tasks/tasks_helper_test.rb
```diff
@@ -91,6 +91,18 @@ class TasksHelperTest < ActionView::TestCase
         csv_file_download_path(run)
     end
 
+    test "#resolve_inclusion_value resolves enum_attr" do
```
etiennebarrie, line 94: This test already passes on main. We should probably use a separate `attribute` that uses a proc or symbol for `:in` and see it fail.

app/helpers/maintenance_tasks/tasks_helper.rb
```diff
@@ -101,16 +101,38 @@ def csv_file_download_path(run)
...
+    # @return [Object] value of the resolved inclusion option.
+    def resolve_inclusion_value(object, parameter_name)
+      inclusion_validator = object.class.validators_on(parameter_name).find do |validator|
         validator.kind == :inclusion
       end
+      return unless inclusion_validator
 
+      in_option = inclusion_validator.options[:in] || inclusion_validator.options[:within]
+
+      return if in_option.respond_to?(:call) || in_option.respond_to?(:to_sym)
```
etiennebarrie, line 124: I was thinking that we should actually test for the case we want instead of the cases we don't want. But it's probably fine that way.

app/helpers/maintenance_tasks/tasks_helper.rb
```diff
@@ -101,16 +101,38 @@ def csv_file_download_path(run)
...
+
+    # Return the appropriate field tag for the parameter, based on its type.
+    # If the parameter has a `validates_inclusion_of` validator, return a dropdown list of options instead.
+    def parameter_field(form_builder, parameter_name)
+      inclusion_values = resolve_inclusion_value(form_builder.object, parameter_name)
       return form_builder.select(
-        parameter_name, inclusion_validator.options[:in], prompt: "Select a value"
-      ) if inclusion_validator
+        parameter_name, inclusion_values, prompt: "Select a value"
+      ) if inclusion_values
```
etiennebarrie, line 135: Style nitpick, the `return` all fits into one line now, it will be a bit neater.

test/helpers/maintenance_tasks/tasks_helper_test.rb
```diff
@@ -91,6 +91,18 @@ class TasksHelperTest < ActionView::TestCase
...
+      with_zone_default(nil) do
```
adrianna-chang-shopify, line 95:
```suggestion
```

I don't think we need this ✂️

test/helpers/maintenance_tasks/tasks_helper_test.rb
adrianna-chang-shopify: I'm fine with keeping this test, but agree that we should add a new test for the case we're fixing 👍`enum_attr` in the test name is a bit vague. Perhaps we could say something like `test "#resolve_inclusion_value resolves inclusion validator for task attribute"` ?

app/helpers/maintenance_tasks/tasks_helper.rb
```diff
@@ -101,16 +101,38 @@ def csv_file_download_path(run)
...
+    #
+    # If there is a value returned, the dropdown list of options will be used.
+    #
+    # Procs, lambdas, and symbols options are not supported by the dropdown form,
+    # and return nil.
+    #
+    # @param object [Task] The object for which the value needs to be resolved.
+    # @param parameter_name [String] The parameter name.
+    #
+    # @return [Object] value of the resolved inclusion option.
```
adrianna-chang-shopify, line 115: Suggestion to condense / clarify the language a bit:

```suggestion
    # Resolves values covered by the inclusion validator for a task attribute.
    # Procs, lambdas, and symbols options are not supported and return nil.
    #
    # Returned values are used to populate a dropdown list of options.
    #
    # @param object [Task] The object for which the value needs to be resolved.
    # @param parameter_name [String] The parameter name.
    #
    # @return [Enumerable] value of the resolved inclusion option.
```

adrianna-chang-shopify: Also, what about passing `object.class` to this method instead? Then we could call the parameter `task_class` or `task`, which is a bit easier to gather meaning from than `object`.

```
# @param task [Class<Task>] The task whose inclusion validator needs to be resolved.
```

test/helpers/maintenance_tasks/tasks_helper_test.rb
WCGPR0: Yes, that makes sense -- related to your [point here](https://github.com/Shopify/maintenance_tasks/pull/962#discussion_r1482809233), that we only render the view for the first `validates_inclusion_of`. This test would only fail on main if we have the proc or symbol as the first validation:
```ruby
    # put the proc validates_inclusion_of before the array one
    validates_inclusion_of :enum_attr, in: proc { [100, 200, 300] }, allow_nil: true
    validates_inclusion_of :enum_attr, in: [100, 200, 300], allow_nil: true
```
(Fails on main)

test/dummy/app/tasks/maintenance/params_task.rb
adrianna-chang-shopify: Yeah, maybe when adding a test case to ensure that procs / symbols don't blow up, we can add a validator on another attribute?

app/helpers/maintenance_tasks/tasks_helper.rb
WCGPR0: > Also, what about passing object.class to this method instead? 

That's a good idea -- applied in https://github.com/Shopify/maintenance_tasks/pull/962/commits/af4e6b0ec4affe1f529043ec9ed93e0bb5f1167c

app/helpers/maintenance_tasks/tasks_helper.rb
WCGPR0: Oh nice -- applied in https://github.com/Shopify/maintenance_tasks/pull/962/commits/af4e6b0ec4affe1f529043ec9ed93e0bb5f1167c

app/helpers/maintenance_tasks/tasks_helper.rb
WCGPR0: Ah sorry - Misunderstood, thanks! And good catch on the Range edge case, handling for what's supported is definitely easier to manage than filtering out what isn't supported.

test/helpers/maintenance_tasks/tasks_helper_test.rb
```diff
@@ -91,14 +91,19 @@ class TasksHelperTest < ActionView::TestCase
...
 
+    test "#resolve_inclusion_value resolves inclusion validator for task attributes" do
+      assert_match "Select a value", markup("integer_dropdown_attr").squish
+
+      assert_match "Select a value", markup("boolean_dropdown_attr").squish
+
+      ["text_integer_attr", "text_integer_attr2", "text_integer_attr3"].each do |text_integer_attr|
+        refute_match "Select a value", markup(text_integer_attr).squish
+      end
+    end
```
WCGPR0, line 102: LMK if this unit test feels redundant now -- since it should fail on the [system test](https://github.com/Shopify/maintenance_tasks/blob/2fda52bf088a999009902a81c9207c700e88f922/test/system/maintenance_tasks/tasks_test.rb), and I can also get rid of this [minor refactor ](https://github.com/Shopify/maintenance_tasks/pull/962/commits/2fda52bf088a999009902a81c9207c700e88f922) to keep it short

WCGPR0:
> This LGTM once the last bits of feedback from @etiennebarrie are addressed! Can you squash your commits as well please? Thank you!

Related commits has been squashed

app/helpers/maintenance_tasks/tasks_helper.rb
```diff
@@ -101,16 +101,31 @@ def csv_file_download_path(run)
...
+    # Resolves values covered by the inclusion validator for a task attribute.
+    # Only Arrays are supported, option types such as:
+    # Procs, lambdas, symbols, and Range are not supported and return nil.
+    #
+    # Returned values are used to populate a dropdown list of options.
+    #
+    # @param task_class [Class<Task>] The task class for which the value needs to be resolved.
+    # @param parameter_name [String] The parameter name.
+    #
+    # @return [Enumerable] value of the resolved inclusion option.
```
adrianna-chang-shopify, line 113:
```suggestion
    # @return [Array] value of the resolved inclusion option.
```

adrianna-chang-shopify (approved):
Can you squash to a single commit please? I'll merge after. Thanks so much for working on this!

WCGPR0:
> Can you squash to a single commit please? I'll merge after. Thanks so much for working on this!

Thanks for the all the reviews - especially for my first contribution to the gem 🙌

Issue #965, opened 2024-02-07
please add a CHANGELOG.md file


tilo:
there is no CHANGELOG.md file in this project

this makes it hard to understand what changed, and if there are breaking changes

nvasilevski:
While it doesn't mean we shouldn't have a changelog I wanted to mention that we are trying our best providing the most accurate changes description, including breaking changes in the release notes, for example: https://github.com/Shopify/maintenance_tasks/releases/tag/v2.0.0

Having said that, I don't think it will be a huge burden to maintain a CHANGELOG. @Shopify/rails should we start maintaining one? Perhaps starting from the next release

lastgabs:
@tilo thanks for the request. 

As per @etiennebarrie 's reply, we're not planning on adding a changelog because it adds more friction than solves problems, so I'll close the issue for now.

tilo:
@etiennebarrie yes, I understand that you don't want to spend extra time on this!

oh - you already have the basic file!! 😮 it is just not in a standard location where people (or RubyGems.org) would expect it :) 

This should make your existing changelog more discoverable: https://github.com/Shopify/maintenance_tasks/pull/966

PR #966, opened 2024-02-08
add a CHANGELOG.html file that redirects to the actual changelog

tilo:
Re [issue #965](https://github.com/Shopify/maintenance_tasks/issues/965) 

I didn't realize that you already have awesome change documentation - just not in the location that has become standard for Ruby gems. 

This small change will make your existing changelog more discoverable -- both from RubyGems.org, as well as when someone browses your project.

Hope this helps

cheers

### DIFF
--- a/maintenance_tasks.gemspec
+++ b/maintenance_tasks.gemspec
@@ -15,6 +15,7 @@ Gem::Specification.new do |spec|
     "source_code_uri" =>
       "https://github.com/Shopify/maintenance_tasks/tree/v#{spec.version}",
     "allowed_push_host" => "https://rubygems.org",
+    "changelog_uri" => "https://github.com/Shopify/maintenance_tasks/blob/main/CHANGELOG.html",
   }
 
   spec.files = Dir["{app,config,db,lib}/**/*", "LICENSE.md", "README.md"]
--- a/CHANGELOG.html
+++ b/CHANGELOG.html
@@ -1,0 +1,1 @@
+<meta http-equiv="refresh" content="0; URL='https://github.com/Shopify/maintenance_tasks/releases'" />


nvasilevski:
This feels a little custom to be honest. GitHub releases is a very commonly used feature, what if we instead teach rubygems to point  to github releases if gem has its source listed on github. This way many more gems can benefit while still allowing gems to have custom changelogs

Actually, we don't even need to change anything in rubygems. Rails for example despite of having CHANGELOG file still points to a specific tag which must be much convenient for the reader 
https://github.com/rails/rails/blob/bab4aa7cb25112846e04cea907361a1de0f126ef/rails.gemspec#L25

So perhaps making `changelog_uri` pointing to a specific tag or a release would be a better option?

etiennebarrie:
Oh yes Rubocop does that, it's pretty nice:
https://github.com/rubocop/rubocop/blob/dc50e8034183646f05e7009e917a452ec1f7ba7e/rubocop.gemspec#L27

Yeah we don't need the html file, it doesn't really redirect in the first place (at least on GitHub), but having every gem version point to the release sounds useful.

tilo:
@nvasilevski the majority of Ruby gems does this, because this is the recommended process:

https://guides.rubygems.org/specification-reference/#metadata

nvasilevski:
> because this is the recommended process:

The guidance hasn't changed [for ~7 years](https://github.com/rubygems/guides/commit/a9e2c02f3205ab8b0c8f7e2308aa4ef5d03c5976), not even sure if Releases feature existed on GitHub at that time, so maybe it's a good opportunity to reconsider it, at least for cases when what seems to be a better alternative exists? Though understandable that rubygems guides should not be strictly recommending github as it would seem one-sided.

>as a user of a gem, people want to see the complete version history, not just the current tag.. This is important, so they can see what their upgrade path will look like

I guess we are keeping gems up-to-date well enough to rarely jump over several version at once so I haven't experienced such need. But even if I needed it, Releases "index" page https://github.com/Shopify/maintenance_tasks/releases
 is one click away from the release "show" page https://github.com/Shopify/maintenance_tasks/releases/tag/v2.5.0

maintenance_tasks.gemspec
```diff
@@ -15,6 +15,7 @@ Gem::Specification.new do |spec|
     "source_code_uri" =>
       "https://github.com/Shopify/maintenance_tasks/tree/v#{spec.version}",
     "allowed_push_host" => "https://rubygems.org",
+    "changelog_uri" => "https://github.com/Shopify/maintenance_tasks/blob/main/CHANGELOG.html",
```
gmcgibbon, line 18: Let's copy what Rails and Rubocop are doing and just link to the right release on GitHub instead: https://github.com/rails/rails/blob/fc9ed119c718ea0654e4a5d2477d8391252cbe6b/rails.gemspec#L28, https://github.com/rubocop/rubocop/blob/d8d8474edf63e4b60ec4b72cd78941520f977d44/rubocop.gemspec#L27

gmcgibbon (changes requested):

PR #981, opened 2024-02-27
Bump to 2.7

Shinomix:
### What are you trying to accomplish with this PR?

Release a minor update including the ActiveSupport notifications for tasks status changes.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    maintenance_tasks (2.6.0)
+    maintenance_tasks (2.7.0)
       actionpack (>= 6.0)
       activejob (>= 6.0)
       activerecord (>= 6.0)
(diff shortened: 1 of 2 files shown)


Shinomix:
(Not entirely sure while CI is failing on rails edge, looks like Capybara reports some flakiness on UI tests)

adrianna-chang-shopify (approved):
:shipit: 

Are you comfortable deploying it with Shipit + generating a GH release as well?

Shinomix:
@adrianna-chang-shopify I'm okay with generating the release but I'm puzzled by the failed CI actions on rails edge. I've re-run them several times, it doesn't seem flaky. Have you experienced that in the past (e.g. when you shipped the 2.6 release)?

adrianna-chang-shopify:
@Shinomix can you try rebasing and see if that gets CI green? We run CI against the latest version of every dependency, so if there's an issue with the latest Rails edge and Maintenance Tasks, it'll show up in CI. I just ran the test suite locally using Rails edge with `ref: "af85f74"` (the ref that's failing CI in the build above), and it fails, but if I use the latest HEAD, the tests pass.

etiennebarrie (approved):

PR #983, opened 2024-02-28
Pin Rails edge version to a518b5a

adrianna-chang-shopify:
I bisected the system test failures (e.g. https://github.com/Shopify/maintenance_tasks/actions/runs/8048576785/job/21980235854?pr=980), and traced them back to https://github.com/rails/rails/commit/e88dba8f0b704ecb134d051a18c6b528fced2ef8.

Unsure if it's an issue on the Maintenance Tasks side or the Rails side, so for now let's pin the version to the previous commit to get CI on main back to green.

cc @casperisfine in case you have an idea what's up here. We have system tests that assert behaviour around running a maintenance task, e.g.
https://github.com/Shopify/maintenance_tasks/blob/7363b64becac8363cab4991930abf1fc9d82db65/test/system/maintenance_tasks/runs_test.rb#L7-L22

Running a task creates a `MaintenanceTasks::Run` record in the db. We query a task's "active runs" here:
https://github.com/Shopify/maintenance_tasks/blob/7363b64becac8363cab4991930abf1fc9d82db65/app/models/maintenance_tasks/task_data_show.rb#L92-L94

to show them on the task page. This should pick up the newly-created run record, but with the changes to the query cache above, the new record is no longer being returned.

cc @Shinomix


casperisfine:
Uh. It's interesting that it's limited to system tests, I wonder what's up.

casperisfine:
I fixed it upstream, so we can close.

PR #991, opened 2024-03-15
CSV accurate count

### LINKED ISSUE #977: CSV count wrong when new line in cell
When a CSV has new lines inside a cell, the default count is incorrect.

CSV sample:
```csv
title,tags
"foo title","elephant
zebra
giraffe"
"bar title","one"

```

Representation:
SCR-20240220-mzuc

There are 2 content rows but with the current implementation, it is shown as 4 rows.
The current implementation use `"\n"` as divider:
https://github.com/Shopify/maintenance_tasks/blob/a9ed8f13fbf9069e93a44882f0c9af2d8c7bec24/app/models/maintenance_tasks/batch_csv_collection_builder.rb#L29-L39

The precessing of the rows themselves seems to be correct (precessing 2 rows).

The task is shown as finished with `Processed 2 out of 4 items (50%).`
SCR-20240220-ncpv

D-system:
Fix #977

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,13 +198,13 @@ My Title,Hello World!
 ```
 
 The files uploaded to your Active Storage service provider will be renamed to
-include an ISO 8601 timestamp and the Task name in snake case format. The CSV is
-expected to have a trailing newline at the end of the file.
+include an ISO 8601 timestamp and the Task name in snake case format.
 
-The `#count` load and parse the entire file to know the accurate number of rows.
-With files with millions of rows, it take several seconds to process. You might
-consider to skip the count or use an approximative approach, eg: count the
-number of new lines:
+The implicit `#count` method loads and parses the entire file to determine the
+accurate number of rows. With files with millions of rows, it takes several
+seconds to process. Consider skipping the count (defining a `count` that returns
+`nil`) or use an approximation, eg: count the number of new lines:
+
 ```ruby
 def count(task)
   task.csv_content.count("\n") - 1
--- a/README.md
+++ b/README.md
@@ -201,6 +201,16 @@ The files uploaded to your Active Storage service provider will be renamed to
 include an ISO 8601 timestamp and the Task name in snake case format. The CSV is
 expected to have a trailing newline at the end of the file.
 
+The `#count` load and parse the entire file to know the accurate number of rows.
+With files with millions of rows, it take several seconds to process. You might
+consider to skip the count or use an approximative approach, eg: count the
+number of new lines:
+```ruby
+def count(task)
+  task.csv_content.count("\n") - 1
+end
+```
+
 #### Batch CSV Tasks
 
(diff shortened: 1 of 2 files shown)


etiennebarrie:
The reason is that a new version of better_html was released. On main, you can repro with this:

```console
$ export BUNDLE_GEMFILE=$PWD/gemfiles/rails_6_0.gemfile
$ rm -f $BUNDLE_GEMFILE.lock
$ bundle
$ bin/rails test
```

etiennebarrie:
Sorry I had already reported at https://github.com/Shopify/better-html/pull/122#issuecomment-1999973339.

D-system:
I removed my better_error commit  and rebased the PR from main

README.md
```diff
@@ -201,6 +201,16 @@ The files uploaded to your Active Storage service provider will be renamed to
 include an ISO 8601 timestamp and the Task name in snake case format. The CSV is
 expected to have a trailing newline at the end of the file.
 
+The `#count` load and parse the entire file to know the accurate number of rows.
+With files with millions of rows, it take several seconds to process. You might
+consider to skip the count or use an approximative approach, eg: count the
+number of new lines:
```
etiennebarrie, line 207: We need to remove the bit about the newline at the end of file since it won't be necessary anymore.

```suggestion
include an ISO 8601 timestamp and the Task name in snake case format.

The implicit `#count` method loads and parses the entire file to determine the
accurate number of rows. With files with millions of rows, it takes several
seconds to process. Consider skipping the count (defining a `count` that returns
`nil`) or use an approximation, eg: count the number of new lines:

```

Also reworded a bit.

app/models/maintenance_tasks/csv_collection_builder.rb
```diff
@@ -16,13 +16,13 @@ def collection(task)
     end
 
     # The number of rows to be processed. Excludes the header row from the
```
etiennebarrie, line 18: We can stop mentioning the part about the header, since it's part of using the CSV library for an accurate row count.

app/models/maintenance_tasks/csv_collection_builder.rb
```diff
@@ -16,13 +16,13 @@ def collection(task)
...
-    # count and assumes a trailing newline is at the end of the CSV file.
-    # Note that this number is an approximation based on the number of
-    # newlines.
+    # count and use the CSV library for an accurate row count.
+    # Note that the entire file is loaded. Files with millions of rows, it will
+    # take several seconds.
```
etiennebarrie, line 21:
```suggestion
    # count and uses the CSV library for an accurate row count.
    # Note that the entire file is loaded. It will take several seconds with files with millions of rows.
```

app/models/maintenance_tasks/csv_collection_builder.rb
D-system: Removed

PR #1014, opened 2024-05-01
Add csv 3.3.0 as an explicit gemspec dependency to support Ruby 3.4.0 default gems changes.

tvongaza:
I get the following warning in our apps tests suite which makes use of the `maintenance_tasks` gem when run against Ruby 3.3.1:

```
~/.rbenv/versions/3.3.1/lib/ruby/gems/3.3.0/gems/bootsnap-1.18.3/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:30: warning: ~/.rbenv/versions/3.3.1/lib/ruby/3.3.0/csv.rb was loaded from the standard library, but will no longer be part of the default gems since Ruby 3.4.0. Add csv to your Gemfile or gemspec.
```
Ruby 3.3.0 adds a warning about Gems which will no longer be part of the default gems in Ruby 3.4.0. Details are here: https://www.ruby-lang.org/en/news/2023/12/25/ruby-3-3-0-released/

> Standard library updates
>RubyGems and Bundler warn if users do require the following gems without adding them to Gemfile or gemspec. This is because they will become the bundled gems in the future version of Ruby.
> 
> Targeted libraries are:
> - csv

This PR adds `csv` to the gemspec to make it compatible with 3.4.0 when it is released.


maintenance_tasks.gemspec
```diff
@@ -24,6 +24,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency("actionpack", ">= 6.0")
   spec.add_dependency("activejob", ">= 6.0")
   spec.add_dependency("activerecord", ">= 6.0")
+  spec.add_dependency("csv", ">= 3.3")
```
etiennebarrie, line 27: Is there anything specific about this version that we need to require it?

tvongaza: It is just the latest version as of today - https://rubygems.org/gems/csv/versions/3.3.0?locale=en. I'm not sure if this specific version is what will no longer exist in 3.4.0, but assuming so.

etiennebarrie: Let's not add any version requirement then.
```suggestion
  spec.add_dependency("csv")
```

etiennebarrie (commented):
You'll need to approve the CLA too.

tvongaza:
I have signed the CLA!

etiennebarrie (approved):

Issue #990, opened 2024-03-11
Figure out slow system tests


etiennebarrie:
It used to be that system tests would take something under 30 seconds, now they often take way more. I think a bunch of them must be waiting for a refresh or something like this. Also it's much more flaky than it used to be locally. In CI it seems to pass much more reliably.
We should investigate and bring that to a reasonable amount of time.

PR #1015, opened 2024-05-08
Adding disable-gpu flag

JoseWhite:
Before disabling this flag on MacOs: Runtime was: 90+ seconds 
After this change MacOs: Runtime down to 17.80s: 

The disable gpu flag was added to `application_system_test_case.rb` application system test case
- Fixes slow system specs on MacOs

- Closes: https://github.com/Shopify/maintenance_tasks/issues/990


etiennebarrie:
Thanks a lot!

PR #1018, opened 2024-05-15
Expose Run instance on Task

olivier-thatch:
Expose Run instance on Task.

This is basically the same as #924, so all credit goes to @shalvah. I'm opening a new PR since the old one was closed and I'm hoping a new PR will be more visible to the maintainers (but also happy to close this PR in favor of the old one).

We've been using this in our codebase via the following monkey patch:
```ruby
require "maintenance_tasks"

module MaintenanceTasks
  class Run < ApplicationRecord
    module SetTaskRunPatch
      def task
        super.tap do |task|
          task.run ||= self
        end
      end
    end

    prepend SetTaskRunPatch
  end

  class Task
    attr_accessor :run
  end
end
```

This enables us to do some interesting stuff like exposing a logger to tasks and storing the log output on the Run instances, or adding the run ID to our error contexts, etc.

### DIFF
--- a/app/models/maintenance_tasks/run.rb
+++ b/app/models/maintenance_tasks/run.rb
@@ -443,6 +443,7 @@ module MaintenanceTasks
     def task
       @task ||= begin
         task = Task.named(task_name).new
+        task.run = self
         if task.attribute_names.any? && arguments.present?
           task.assign_attributes(arguments)
         end
--- a/app/models/maintenance_tasks/task.rb
+++ b/app/models/maintenance_tasks/task.rb
@@ -191,6 +191,9 @@ module MaintenanceTasks
       end
     end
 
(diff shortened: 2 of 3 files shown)


olivier-thatch:
@etiennebarrie Sorry for the ping but would love to get some feedback on this PR :)

etiennebarrie:
Our stance has been not to expose Run itself, so that it stays an internal concern, and solving the actual issues directly. For example:

> adding the run ID to our error contexts

Since we added #941, we should probably also add the id to the error task context: https://github.com/Shopify/maintenance_tasks/blob/93a50672c36a26e3a0954bb2deb50ab211b100f7/app/jobs/concerns/maintenance_tasks/task_job_concern.rb#L173-L177

PR #1031, opened 2024-05-28
Add direct upload capability

lavoiesl:
Depends on https://github.com/Shopify/maintenance_tasks/pull/1029

Adds the ability to enable [direct uploads](https://guides.rubyonrails.org/v7.1/active_storage_overview.html#direct-uploads), which can help circumvent upload limits on the server.

### DIFF
--- a/test/system/maintenance_tasks/runs_test.rb
+++ b/test/system/maintenance_tasks/runs_test.rb
@@ -60,16 +60,42 @@ module MaintenanceTasks
       MaintenanceTasks.metadata = nil
     end
 
-    test "run a CSV Task" do
-      visit maintenance_tasks_path
+    test "run a CSV Task without direct upload" do
+      with_direct_upload(false) do
+        visit maintenance_tasks_path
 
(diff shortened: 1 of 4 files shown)


test/system/maintenance_tasks/runs_test.rb
```diff
@@ -60,16 +60,42 @@ class RunsTest < ApplicationSystemTestCase
...
+
+    test "run a CSV Task with direct upload" do
+      with_direct_upload(true) do
+        visit maintenance_tasks_path
+
+        click_on("Maintenance::ImportPostsTask")
+
+        assert_selector("input[type=file]") do |input|
+          assert_equal(input["data-direct-upload-url"], rails_direct_uploads_url)
+        end
```
lavoiesl, line 90: Those tests are failing on Rails `<= 7.0`.

Not sure why, the usage is supposed to be the same: https://guides.rubyonrails.org/v6.1/active_storage_overview.html#usage

I could use some help 😅

PR #1032, opened 2024-05-29
Revert "Move to using vendored bulma and assets pipeline"

lavoiesl:
Reverts Shopify/maintenance_tasks#1029 while we investigate how not to break apps without `sprockets-rails`


rafaelfranca:
I'm ok with breaking support for apps not using Rails assets pipeline. We should use the most Rails defaults as possible. Of course if apps want to use something that isn't default they should, but that doesn't means they don't need to do anything. They will have to build their own solution to add support to maintenance task. Leave the easy path to those who follow Rails defaults.

rafaelfranca:
BTW, this gem is analog to [mission_control-jobs](https://github.com/rails/mission_control-jobs), and as you can see, it also only works in the Rails' assets pipeline (sprockets-rails or propshaft).

lavoiesl:
I don't have a strong opinion, I'll let you and @etiennebarrie decide.

PR #1033, opened 2024-05-29
Resume runs from saved position

segiddins:
Previously, they would be restarted with a nil cursor

### DIFF
--- a/app/models/maintenance_tasks/runner.rb
+++ b/app/models/maintenance_tasks/runner.rb
@@ -65,6 +65,7 @@ module MaintenanceTasks
     # @raise [EnqueuingError] if an error occurs while enqueuing the Run.
     def resume(run)
       job = instantiate_job(run)
+      job.cursor_position = run.cursor
       run.job_id = job.job_id
       run.enqueued!
       enqueue(run, job)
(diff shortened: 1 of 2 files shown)


etiennebarrie:
Interestingly, with ErrorTask, resuming works the first time, but then fails the second time.

While these change fixe the issue, I think the problem is more related to the job concern. It uses the run's cursor if the job doesn't have a cursor:

https://github.com/Shopify/maintenance_tasks/blob/3efda1894f7810837c74c555a3d722f58323733f/app/jobs/concerns/maintenance_tasks/task_job_concern.rb#L34

But at the end of the job, it saves the cursor from the job:
https://github.com/Shopify/maintenance_tasks/blob/3efda1894f7810837c74c555a3d722f58323733f/app/jobs/concerns/maintenance_tasks/task_job_concern.rb#L133

so if the job fails immediately, the cursor from the job is nil, and we lose the run's cursor.

This is where the problem should be fixed IMO.

segiddins:
Can we merge this for the time being, since it at least fixes one scenario and I don't think it will be a regression in the case you mention?

etiennebarrie:
I'm sorry I haven't been clear, I don't have another case, I'm talking about your scenario.

I'm just saying in terms of implementation we don't really want Runner to have to set the cursor on the job from the outside (since the cursor is already in the run, and we pass the run to the job). Instead, we want the run not to lose the value it has, by setting the cursor on the job in the job, leaving that responsibility to the job.

I have opened #1036 to fix this.

Issue #890, opened 2023-09-14
How do we set `limit`?


bray:
When `collection` returns an `ActiveRecord::Relation`, how do we set `limit` for each "batch"?  `job-iteration`'s [Best Practices](https://github.com/Shopify/job-iteration/blob/1ebc6974cceddd7e526630e85875059cf6c9da0f/guides/best-practices.md#batch-iteration) says it defaults to 100.  100 is too small for expensive queries and/or large datasets.

It looks like that's passed in [here](https://github.com/Shopify/job-iteration/blob/1ebc6974cceddd7e526630e85875059cf6c9da0f/lib/job-iteration/active_record_enumerator.rb#L10-L12).

But for `maintenance_tasks`, in the `collection` method, we're not explicitly building an enumerator. So how can we pass in or set the `@batch_size`?  I've tried some guesses like `@batch_size`, `@of`, `def batch_size`, `def of` inside the `MaintenanceTasks::Task`.

I also tried returning a few different classes in the collection method, but none have worked:

`#find_each(batch_size: 2000)` gets us what we want, but this returns an Enumerator and `#collection` can't be an Enumerator (it gives the error "#collection must be either an Active Record Relation, ActiveRecord::Batches::BatchEnumerator, Array, or CSV.").

`#in_batches(batch_size: 2000)` works (it returns an `ActiveRecord::Batches::BatchEnumerator`), but this is only if you want to [process a batch](https://github.com/Shopify/maintenance_tasks/tree/main#processing-batch-collections) in process, rather than individual items.

`#in_batches(batch_size: 2000).each_record` also returns an Enumerator.

nvasilevski:
I'd say this is purely a missing API as there is no reason not to allow tasks to redefine the batch size that is being used under the hood. The batch size will need to be passed to the `active_record_on_records` enumerator using the `batch_size:` option
https://github.com/Shopify/maintenance_tasks/blob/e1765da66ff0c870891c8608a7260f8a8292ac0e/app/jobs/concerns/maintenance_tasks/task_job_concern.rb#L42

So I'll try to put it on our roadmap or PRs are always welcomed! Unless someone else from maintainers remembers/has concerns on why this feature should not exist

The only requirement for this feature is that the tasks shouldn't be forced to always specify the batch size and it still should default to `100` (or to `nil` which means we'll use the default from `job-iteration`)

bray:
Ok, thanks - good to know I wasn't missing something. I'll try to get up a PR!

[...]

PR #1037, opened 2024-06-04
Allow Active Record collection tasks to specify a batch size.

adrianna-chang-shopify:
Closes: https://github.com/Shopify/maintenance_tasks/issues/890

This PR adds a task DSL for specifying the batch size on an Active Record Relation collection. Currently, job-iteration will default to fetching 100 records with `active_record_on_records`, which is rather small for a table with many thousands of rows. Task authors can now customize the batch size per task with a `collection_batch_size` method.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,31 @@ module Maintenance
 end
 ```
 
+#### Customizing the Batch Size
+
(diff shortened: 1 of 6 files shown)


nvasilevski (approved):
👍

PR #1040, opened 2024-06-07
Add support for bounded enumerables and stateless Procs in inclusion validation

lavoiesl:
Add support for a few variations of inclusion constraints.

The motivating factor behind this is to be able to resolve to a list of classes that can only be resolved at runtime.

For example, this is not stable, it depends in which order the classes are loaded:
```rb
validates :model, inclusion: { in: ActiveRecord::Base.descendants.map(&:name) }
```

It needs a lambda:
```rb
validates :model, inclusion: { in: ->() { ActiveRecord::Base.descendants.map(&:name) } }
```

However, stateful lambda are prohibited, because they depend on the task instance:
```rb
validates :model, inclusion: { in: ->() { |task| task.some_method } }
```

---

While at it, this PR also adds support for bounded enumerables:
```rb
validates :timeout, inclusion: { in: (0..30).step(5) }
```

---

See the update test and comments for a full explanation.

### DIFF
--- a/app/helpers/maintenance_tasks/tasks_helper.rb
+++ b/app/helpers/maintenance_tasks/tasks_helper.rb
@@ -7,6 +7,8 @@ module MaintenanceTasks
   #
   # @api private
   module TasksHelper
+    MAX_INCLUSION_ITEMS = 1000
+
     STATUS_COLOURS = {
       "new" => ["is-primary"],
       "enqueued" => ["is-primary is-light"],
@@ -101,10 +103,21 @@ module MaintenanceTasks
       )
     end
 
-    # Resolves values covered by the inclusion validator for a task attribute.
-    # Only Arrays are supported.
-    # Procs and lambdas are also supported, but only if they take no arguments and return an Array.
-    # Option types such as Symbols, and Range are not supported and return nil.
+    # Resolves values covered by the inclusion validator for a task attribute,
+    # but only if it can be resolved to an array in a stateless manner.
+    #
+    # Examples of supported constraints:
+    # - Arrays
+    # - Enumerable that can be converted to an Array
+    # - Procs and lambdas, but only if they take no arguments and return a value meeting the requirements above.
+    #
+    # Examples of unsupported constraints, which would return nil:
+    # - Symbols (which would normally resolve to an instance method)
+    # - Procs and lambdas that take an argument (which would be the instance)
+    # - Unbounded ranges (e.g. `1..`)
+    #
+    # If the resulting Array is larger than #MAX_INCLUSION_ITEMS, the method will return nil.
+    # This is meant to make sure the process does not misbehave if the inclusion range is giant or infinite.
     #
     # Returned values are used to populate a dropdown list of options.
     #
@@ -120,6 +133,12 @@ module MaintenanceTasks
 
       in_option = inclusion_validator.options[:in] || inclusion_validator.options[:within]
       in_option = in_option.call if in_option.is_a?(Proc) && in_option.arity.zero?
+
+      if in_option.is_a?(Enumerable)
+        items = in_option.take(MAX_INCLUSION_ITEMS + 1).to_a
+        in_option = items if items.size <= MAX_INCLUSION_ITEMS
+      end
+
       in_option if in_option.is_a?(Array)
     end
 
--- a/app/helpers/maintenance_tasks/tasks_helper.rb
+++ b/app/helpers/maintenance_tasks/tasks_helper.rb
@@ -102,8 +102,9 @@ module MaintenanceTasks
     end
 
     # Resolves values covered by the inclusion validator for a task attribute.
-    # Only Arrays are supported, option types such as:
-    # Procs, lambdas, symbols, and Range are not supported and return nil.
+    # Only Arrays are supported.
+    # Procs and lambdas are also supported, but only if they take no arguments and return an Array.
+    # Option types such as Symbols, and Range are not supported and return nil.
     #
     # Returned values are used to populate a dropdown list of options.
     #
@@ -118,6 +119,7 @@ module MaintenanceTasks
       return unless inclusion_validator
 
       in_option = inclusion_validator.options[:in] || inclusion_validator.options[:within]
+      in_option = in_option.call if in_option.is_a?(Proc) && in_option.arity.zero?
       in_option if in_option.is_a?(Array)
     end
 
--- a/test/dummy/app/tasks/maintenance/params_task.rb
+++ b/test/dummy/app/tasks/maintenance/params_task.rb
@@ -27,10 +27,12 @@ module Maintenance
 
     # Dropdown options with unsupported scenarios
     attribute :text_integer_attr_proc_no_arg, :integer
+    attribute :text_integer_attr_proc_arg, :integer
     attribute :text_integer_attr_undefined_symbol, :integer
     attribute :text_integer_attr_unbounded_range, :integer
 
     validates_inclusion_of :text_integer_attr_proc_no_arg, in: proc { [100, 200, 300] }, allow_nil: true
+    validates_inclusion_of :text_integer_attr_proc_arg, in: proc { |_task| [100, 200, 300] }, allow_nil: true
     validates_inclusion_of :text_integer_attr_undefined_symbol, in: :undefined_symbol, allow_nil: true
     validates_inclusion_of :text_integer_attr_unbounded_range, in: (100..), allow_nil: true
 
--- a/test/dummy/app/tasks/maintenance/params_task.rb
+++ b/test/dummy/app/tasks/maintenance/params_task.rb
@@ -30,11 +30,15 @@ module Maintenance
     attribute :text_integer_attr_proc_arg, :integer
     attribute :text_integer_attr_undefined_symbol, :integer
     attribute :text_integer_attr_unbounded_range, :integer
+    attribute :text_integer_attr_bounded_range, :integer
+    attribute :text_integer_attr_enumerable, :integer
 
     validates_inclusion_of :text_integer_attr_proc_no_arg, in: proc { [100, 200, 300] }, allow_nil: true
     validates_inclusion_of :text_integer_attr_proc_arg, in: proc { |_task| [100, 200, 300] }, allow_nil: true
     validates_inclusion_of :text_integer_attr_undefined_symbol, in: :undefined_symbol, allow_nil: true
     validates_inclusion_of :text_integer_attr_unbounded_range, in: (100..), allow_nil: true
+    validates_inclusion_of :text_integer_attr_bounded_range, in: (100..120), allow_nil: true
+    validates_inclusion_of :text_integer_attr_enumerable, in: (100..120).step(5), allow_nil: true
 
     class << self
       attr_accessor :fast_task
--- a/test/dummy/app/tasks/maintenance/params_task.rb
+++ b/test/dummy/app/tasks/maintenance/params_task.rb
@@ -26,13 +26,13 @@ module Maintenance
     validates_inclusion_of :boolean_dropdown_attr, within: [true, false], allow_nil: true
 
     # Dropdown options with unsupported scenarios
-    attribute :text_integer_attr, :integer
-    attribute :text_integer_attr2, :integer
-    attribute :text_integer_attr3, :integer
+    attribute :text_integer_attr_proc_no_arg, :integer
+    attribute :text_integer_attr_undefined_symbol, :integer
+    attribute :text_integer_attr_unbounded_range, :integer
 
-    validates_inclusion_of :text_integer_attr, in: proc { [100, 200, 300] }, allow_nil: true
-    validates_inclusion_of :text_integer_attr2, in: :undefined_symbol, allow_nil: true
-    validates_inclusion_of :text_integer_attr3, in: (100..), allow_nil: true
+    validates_inclusion_of :text_integer_attr_proc_no_arg, in: proc { [100, 200, 300] }, allow_nil: true
+    validates_inclusion_of :text_integer_attr_undefined_symbol, in: :undefined_symbol, allow_nil: true
+    validates_inclusion_of :text_integer_attr_unbounded_range, in: (100..), allow_nil: true
 
     class << self
       attr_accessor :fast_task
--- a/test/helpers/maintenance_tasks/tasks_helper_test.rb
+++ b/test/helpers/maintenance_tasks/tasks_helper_test.rb
@@ -92,12 +92,19 @@ module MaintenanceTasks
     end
 
     test "#resolve_inclusion_value resolves inclusion validator for task attributes" do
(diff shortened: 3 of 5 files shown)


app/helpers/maintenance_tasks/tasks_helper.rb
```diff
@@ -118,6 +132,13 @@ def resolve_inclusion_value(task_class, parameter_name)
       return unless inclusion_validator
 
       in_option = inclusion_validator.options[:in] || inclusion_validator.options[:within]
+      in_option = in_option.call if in_option.is_a?(Proc) && in_option.arity.zero?
+
+      if in_option.is_a?(Enumerable)
+        items = in_option.take(MAX_INCLUSION_ITEMS + 1).to_a
```
bony2023, line 138: nit, un-necessary as `.take` already returns an Array:

```suggestion
        items = in_option.take(MAX_INCLUSION_ITEMS + 1)
```

test/dummy/app/tasks/maintenance/params_task.rb
```diff
@@ -26,13 +26,19 @@ class ParamsTask < MaintenanceTasks::Task
...
+    attribute :text_integer_attr_unbounded_range, :integer
+    attribute :text_integer_attr_bounded_range, :integer
+    attribute :text_integer_attr_enumerable, :integer
+
+    validates_inclusion_of :text_integer_attr_proc_no_arg, in: proc { [100, 200, 300] }, allow_nil: true
+    validates_inclusion_of :text_integer_attr_proc_arg, in: proc { |_task| [100, 200, 300] }, allow_nil: true
+    validates_inclusion_of :text_integer_attr_undefined_symbol, in: :undefined_symbol, allow_nil: true
+    validates_inclusion_of :text_integer_attr_unbounded_range, in: (100..), allow_nil: true
+    validates_inclusion_of :text_integer_attr_bounded_range, in: (100..120), allow_nil: true
+    validates_inclusion_of :text_integer_attr_enumerable, in: (100..120).step(5), allow_nil: true
```
bony2023, line 41: non blocker but it would be good to also test the `MAX_INCLUSION_ITEMS` logic?

```suggestion
    validates_inclusion_of :text_integer_attr_enumerable, in: (100..120).step(5), allow_nil: true
    validates_inclusion_of :text_integer_attr_large_enumerable, in: (1..1001).step(1), allow_nil: true
```

bony2023 (approved):
Amazing, thank you! 🎉  
Few comments but not a blocker to merge

app/helpers/maintenance_tasks/tasks_helper.rb
lavoiesl: Take doesn't return an array if `in_option` is a lazy enumerator:
```rb
(1..).lazy.take(10)
=> #<Enumerator::Lazy: ...>
```

etiennebarrie (changes requested):
I don't think we want this. We're doing a best effort job of showing possible values when the inclusion validator is simple enough, but we shouldn't go and call random code, this is called in a view after all.

We have a way to extend the form by adding a `maintenance_tasks/tasks/custom` partial in the application. (It's not documented BTW sorry). I guess you could use this to conditionally add fields, hide the previous fields for an attribute, if you wanted to go down that route.

I think it's problematic that some validations can now break your views or make choices inaccessible. There should be some kind of opt-in at the very least.

Or maybe it's just not something that should be done in the engine. We do mention "Normally maintenance tasks are ephemeral, so they are used briefly and then deleted." in the README. The initial use case of wanting to list the application records feels like something that might be built in the application directly instead in some kind of admin interface, if it needs to access any and all models at any time.

https://github.com/Shopify/maintenance_tasks#should-i-use-maintenance-tasks

Otherwise one can always do `Rails.application.eager_load!; ActiveRecord::Base.descendants` in a console and stick the result in the validation of the task's param (since the task is temporary, it should be fine). Or just live with the fact that the form will not include a dropdown for the model, the target user for the engine being a developer, they should be able to deal with that situation.

app/helpers/maintenance_tasks/tasks_helper.rb
```diff
@@ -101,9 +103,21 @@ def csv_file_download_path(run)
...
+    # - Enumerable that can be converted to an Array
+    # - Procs and lambdas, but only if they take no arguments and return a value meeting the requirements above.
+    #
+    # Examples of unsupported constraints, which would return nil:
+    # - Symbols (which would normally resolve to an instance method)
+    # - Procs and lambdas that take an argument (which would be the instance)
+    # - Unbounded ranges (e.g. `1..`)
+    #
+    # If the resulting Array is larger than #MAX_INCLUSION_ITEMS, the method will return nil.
+    # This is meant to make sure the process does not misbehave if the inclusion range is giant or infinite.
```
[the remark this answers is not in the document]
lavoiesl: Correct.

"the process does not misbehave" refers to being slow (if giant) or hanging forever (if infinite).

However, using `in: 1..` is a perfectly acceptable use of the inclusion constraint and will properly validate any number above 0.

lavoiesl:
> So what are stateful constraints by the way? Lambdas that take the record? Also in: :some_method? We do have the object handy (although it's only initialized with its defaults obviously).
>
> I guess they raise more issues, but I wonder why we wouldn't support that as well at this point.

Yeah, validations are designed to run within the context of the object with the values at the right place, so adding that limitation seems fair to me.

If someone wants to call a method, they can do so in a proc:
```rb
validates :model, inclusion: { in: ->() { MyClass.allowed_model_values } }
```

---

> Do we need to deal with potential exceptions raised from the lambda? Before we were just getting a value from the task, now we're calling code, so I guess there's more potential for failure here.

Indeed, but if the code is throwing an exception, I think it's perfectly reasonable to let it raise unhandled

---

> We should also document the dropdown feature (and the scope of the support for views with this PR) in the README as well. I guess that could be an indicator that the feature is too complex if we need too many words to explain it. Before: "if you have an inclusion validator on your parameter, if the allowed values is a Array, it will be used to build a dropdown". Now we'll need to add some more words.

Yeah, I can do that once we settle 👍🏼 

---

> Also just realized that if there's an if: or unless: on a validation, we're not respecting that (and ignoring the validation altogether), and always forcing the values to be in the allowed list with the dropdown, when the validation might not apply.

That's a valid concern, but it's also already true today. I don't think this PR makes the status quo any different.

app/helpers/maintenance_tasks/tasks_helper.rb
```diff
@@ -118,6 +132,13 @@ def resolve_inclusion_value(task_class, parameter_name)
       return unless inclusion_validator
 
       in_option = inclusion_validator.options[:in] || inclusion_validator.options[:within]
+      in_option = in_option.call if in_option.is_a?(Proc) && in_option.arity.zero?
+
+      if in_option.is_a?(Enumerable)
+        items = in_option.take(MAX_INCLUSION_ITEMS + 1).to_a
+        in_option = items if items.size <= MAX_INCLUSION_ITEMS
```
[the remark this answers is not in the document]
lavoiesl: I think that should work, yeah. 
Not sure why, but I thought `.size` could return nil.
Looks like it returns Infinity:
```rb
> (1..).size
=> Infinity
```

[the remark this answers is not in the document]
lavoiesl: Ah, yes, `size` can indeed be `nil`:

```rb
e = Enumerator.new {|y| y << 1 }
=> #<Enumerator: ...>

> e.is_a?(Enumerable)
=> true

> e.size
=> nil
```

nvasilevski:
>Perhaps I picked up something?

It's reproducible on `main` so not related to the PR

app/helpers/maintenance_tasks/tasks_helper.rb
```diff
@@ -118,6 +132,16 @@ def resolve_inclusion_value(task_class, parameter_name)
...
+
+      if in_option.is_a?(Enumerable) &&
+          !(size = in_option.size).nil? &&
+          size > 0 &&
+          size <= MaintenanceTasks.max_task_enumerable_picker_size
+
+        in_option = in_option.to_a
+      end
+
       in_option if in_option.is_a?(Array)
```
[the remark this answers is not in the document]
lavoiesl: Yeah. I can change it to also include arrays, but that's a bigger departure from the way things work today. Considering it's 1000 items by default, I don't think it matters much either way. 

The max number of items is mostly to prevent infinite arrays.

lib/maintenance_tasks.rb
```diff
@@ -96,4 +96,16 @@ module MaintenanceTasks
   #
   #  @return [ActiveSupport::Duration] the threshold in seconds after which a task is considered stuck.
   mattr_accessor :stuck_task_duration, default: 5.minutes
+
+  # @!attribute max_task_enumerable_picker_size
```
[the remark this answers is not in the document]
lavoiesl: SGTM

app/helpers/maintenance_tasks/tasks_helper.rb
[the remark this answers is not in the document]
lavoiesl: Ok, changed it to apply to arrays too.

test/dummy/app/tasks/maintenance/params_task.rb
```diff
@@ -20,24 +20,36 @@ class ParamsTask < MaintenanceTasks::Task
...
+    attribute :text_integer_attr_proc_arg, :integer
+    attribute :text_integer_attr_unbounded_range, :integer
+    attribute :text_integer_attr_method_arg, :integer
 
-    validates_inclusion_of :text_integer_attr, in: proc { [100, 200, 300] }, allow_nil: true
-    validates_inclusion_of :text_integer_attr2, in: :undefined_symbol, allow_nil: true
-    validates_inclusion_of :text_integer_attr3, in: (100..), allow_nil: true
+    validates_inclusion_of :text_integer_attr_proc_arg, in: proc { |_task| [100, 200, 300] }, allow_nil: true
+    validates_inclusion_of :text_integer_attr_unbounded_range, in: (100..), allow_nil: true
+    validates_inclusion_of :text_integer_attr_method_arg, in: :method_with_arg, allow_nil: true
```
adrianna-chang-shopify, line 39: This will raise `ArgumentError` if a value is supplied in the UI and the task is ran, because the inclusion validator for Active Model Attributes does not support methods with arguments.

I'd still like to test that the UI renders properly -- that the arity check works and we render a text field here instead of a dropdown.

test/dummy/app/tasks/maintenance/params_task.rb
```diff
@@ -20,19 +20,32 @@ class ParamsTask < MaintenanceTasks::Task
...
+    attribute :text_integer_attr_proc_arg, :integer
+    attribute :text_integer_attr_undefined_symbol, :integer
+    attribute :text_integer_attr_unbounded_range, :integer
+    attribute :text_integer_attr_sizeless_enumerator, :integer
 
-    validates_inclusion_of :text_integer_attr, in: proc { [100, 200, 300] }, allow_nil: true
-    validates_inclusion_of :text_integer_attr2, in: :undefined_symbol, allow_nil: true
-    validates_inclusion_of :text_integer_attr3, in: (100..), allow_nil: true
+    validates_inclusion_of :text_integer_attr_proc_arg, in: proc { |_task| [100, 200, 300] }, allow_nil: true
+    validates_inclusion_of :text_integer_attr_undefined_symbol, in: :undefined_symbol, allow_nil: true
```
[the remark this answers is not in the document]
adrianna-chang-shopify: I opted to remove this attribute, although technically this helps us test that the UI still renders properly if something invalid is specified here. Yes, it will break when we actually instantiate the task, but we check at least that the view doesn't break while attempting to resolve the inclusion values. Similar to the comment I left here: https://github.com/Shopify/maintenance_tasks/pull/1040#discussion_r1842850410

test/dummy/app/tasks/maintenance/params_task.rb
etiennebarrie: We shouldn't keep it because it's not a valid Active Record validation.

app/helpers/maintenance_tasks/tasks_helper.rb
```diff
@@ -102,29 +102,40 @@ def csv_file_download_path(run)
...
         validator.kind == :inclusion
       end
       return unless inclusion_validator
 
       in_option = inclusion_validator.options[:in] || inclusion_validator.options[:within]
+      in_option = in_option.call if in_option.is_a?(Proc) && in_option.arity.zero?
+
+      if in_option.is_a?(Symbol) && task.respond_to?(in_option)
+        method = task.method(in_option)
+        in_option = method.call if method.arity.zero?
```
etiennebarrie, line 129: Let's just support all the normal resolution for this: https://github.com/rails/rails/blob/8aee12ec51728a6d8472335ec9d4ad68d0521773/activemodel/lib/active_model/validations/resolve_value.rb

Making a difference between a proc with a parameter or not makes little sense.

README.md
```diff
@@ -485,6 +485,24 @@ to run. Since arguments are specified in the user interface via text area
...
 
+#### Validating Task Parameters
+
+Task attributes can be validated using Active Model Validations. Attributes
+are validated when a Run is created, and validation errors are recorded on the
+Run.
+
+If an attribute uses an inclusion validator with a supported enumerable type,
+the set of values will be used to populate a dropdown in the user interface. The
+following types are supported:
```
etiennebarrie, line 496: "recorded" is a bit misleading, we don't persist the errors. Also I don't think we should mention the Run.
```suggestion
Task attributes can be validated using Active Model Validations. Attributes are
validated before a Task is enqueued.

If an attribute uses an inclusion validator with a supported `in:` option, the
set of values will be used to populate a dropdown in the user interface. The
following types are supported:
```

app/helpers/maintenance_tasks/tasks_helper.rb
adrianna-chang-shopify: Ah neat, I didn't realize resolution worked like this. Will fix 👍

adrianna-chang-shopify:
@etiennebarrie fixed to support arity-one procs / lambdas appropriately, and I also tweaked this to support callable objects, since that is supported in the `Validations::ResolveValue` module you linked.

README.md
```diff
@@ -485,6 +485,23 @@ to run. Since arguments are specified in the user interface via text area
...
+validated before a Task is enqueued.
+
+If an attribute uses an inclusion validator with a supported `in:` option, the
+set of values will be used to populate a dropdown in the user interface. The
+following types are supported:
+
+* Arrays
+* Procs and lambdas that optionally accept the instance instance, and return an Array.
+* Callable objects that receive one argument, the task instance, and return an Array.
+* Methods that return an Array, called on the task instance.
```
etiennebarrie, line 500:
```suggestion
* Procs and lambdas that optionally accept the Task instance, and return an Array.
* Callable objects that receive one argument, the Task instance, and return an Array.
* Methods that return an Array, called on the Task instance.
```

etiennebarrie (approved):

PR #1050, opened 2024-07-09
Replace build_enumerator with enumerator_builder in README

### LINKED ISSUE #1049: README mentions testing custom enumerators via #build_enumerator but should that be #enumerator_builder?
Hello, thanks for your work on this project. We're finding it very helpful.

I had to write a task that uses a custom enumerator, and I followed the steps in the README here: https://github.com/Shopify/maintenance_tasks?tab=readme-ov-file#tasks-with-custom-enumerators

> If you have a special use case requiring iteration over an unsupported collection type, such as external resources fetched from some API, you can implement the enumerator_builder(cursor:) method in your task.

Which I've done. [Later in the README](https://github.com/Shopify/maintenance_tasks?tab=readme-ov-file#writing-tests-for-a-task-that-uses-a-custom-enumerator), though, under "Writing tests for a Task that uses a custom enumerator":

> Tests for tasks that use custom enumerators need to instantiate the task class in order to call #build_enumerator. Once the task instance is set up, validate that #build_enumerator returns an enumerator yielding pairs of [item, cursor] as expected.

Is this correct? When I try to instantiate a task and call #build_enumerator, I get a NoMethodError. Looking through the source, it looks like [build_enumerator is defined on a different class](https://github.com/Shopify/maintenance_tasks/blob/b93a378e321c21dadbf0dcbb6e79648d33ed3182/app/jobs/concerns/maintenance_tasks/task_job_concern.rb#L33) (Maintenance::TaskJob), not on `Maintenance::Task`. Or am I missing something?

yjukaku:
Fixes #1049

`#enumerator_builder` is the right method to test, while `#build_enumerator` is an internal method on a different class.

### DIFF
--- a/README.md
+++ b/README.md
@@ -752,9 +752,9 @@ end
 ### Writing tests for a Task that uses a custom enumerator
 
 Tests for tasks that use custom enumerators need to instantiate the task class
-in order to call `#build_enumerator`. Once the task instance is set up, validate
-that `#build_enumerator` returns an enumerator yielding pairs of [item, cursor]
-as expected.
+in order to call `#enumerator_builder`. Once the task instance is set up,
+validate that `#enumerator_builder` returns an enumerator yielding pairs of
+`[item, cursor]` as expected.
 
 ```ruby
 # test/tasks/maintenance/custom_enumerating_task.rb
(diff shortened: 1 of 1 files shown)


etiennebarrie (approved):
Thanks!

PR #1059, opened 2024-07-29
Ensure task_context is never nil when error_handler is called

yjukaku:
If `@run.persist_error(error)` raises an error, `task_context` is never assigned and then is `nil` when `MaintenanceTasks.error_handler` is `call`ed.

### DIFF
--- a/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
+++ b/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
@@ -166,6 +166,7 @@ module MaintenanceTasks
     end
 
     def on_error(error)
+      task_context = {}
       @ticker.persist if defined?(@ticker)
 
       if defined?(@run)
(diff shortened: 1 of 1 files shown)


app/jobs/concerns/maintenance_tasks/task_job_concern.rb
```diff
@@ -182,7 +182,7 @@ def on_error(error)
       end
       errored_element = @errored_element if defined?(@errored_element)
     ensure
-      MaintenanceTasks.error_handler.call(error, task_context, errored_element)
+      MaintenanceTasks.error_handler.call(error, task_context || {}, errored_element)
```
adrianna-chang-shopify, line 185: What do you think about assigning `task_context = {}` at the start of this method instead? That would let us drop the else branch above as well.

```ruby
def on_error(error)
  task_context = {}

  @ticker.persist if defined?(@ticker)

  if defined?(@run)
    ...
    task_context = { ... }
  end
  ...
ensure
  MaintenanceTasks.error_handler.call(error, task_context, errored_element)
end

yjukaku: Good idea, that makes it much cleaner 👍

adrianna-chang-shopify (commented):
Can you squash your commits please @yjukaku and I'll merge ?

adrianna-chang-shopify (approved):

PR #1066, opened 2024-08-19
Make cursor to string migration reversible

### LINKED ISSUE #1065: String cursor migration is not reversible
I'm trying out this gem, and I got this error when trying to reverse the migrations so that I could work on a different branch.

```
== 20240817025553 ChangeCursorToString: reverting =============================
-- change_table(:maintenance_tasks_runs)
bin/rails aborted!
StandardError: An error has occurred, this and all later migrations canceled: (StandardError)

PG::DatatypeMismatch: ERROR:  column "cursor" cannot be cast automatically to type bigint
HINT:  You might need to specify "USING cursor::bigint".
/app/db/migrate/20240817025553_change_cursor_to_string.maintenance_tasks.rb:16:in `block in down'
/app/db/migrate/20240817025553_change_cursor_to_string.maintenance_tasks.rb:15:in `down'

Caused by:
ActiveRecord::StatementInvalid: PG::DatatypeMismatch: ERROR:  column "cursor" cannot be cast automatically to type bigint (ActiveRecord::StatementInvalid)
HINT:  You might need to specify "USING cursor::bigint".
/app/db/migrate/20240817025553_change_cursor_to_string.maintenance_tasks.rb:16:in `block in down'
/app/db/migrate/20240817025553_change_cursor_to_string.maintenance_tasks.rb:15:in `down'
```

I think maybe the way to fix this would be to change it to this:

```ruby
  def down
      t.change(:cursor, "bigint USING cursor::bigint")
    end
  end
```

I was a bit confused about how to apply the hint but then I found this https://stackoverflow.com/a/44132110 but at least in this case I don't think it's necessary to use the CAST function explicitly. I expect that it's largely moot anyway since people probably won't be reversing this in production, only in development environments where they're testing it out.

Thanks!

ibrahima:
I think this fixes #1065.

### DIFF
--- a/db/migrate/20210219212931_change_cursor_to_string.rb
+++ b/db/migrate/20210219212931_change_cursor_to_string.rb
@@ -12,7 +12,7 @@ class ChangeCursorToString < ActiveRecord::Migration[6.0]
 
   def down
     change_table(:maintenance_tasks_runs) do |t|
-      t.change(:cursor, "bigint USING cursor::bigint")
+      t.change(:cursor, "bigint USING CAST(cursor as bigint)")
     end
   end
 end
--- a/db/migrate/20210219212931_change_cursor_to_string.rb
+++ b/db/migrate/20210219212931_change_cursor_to_string.rb
@@ -12,7 +12,7 @@ class ChangeCursorToString < ActiveRecord::Migration[6.0]
 
   def down
     change_table(:maintenance_tasks_runs) do |t|
-      t.change(:cursor, :bigint)
+      t.change(:cursor, "bigint USING cursor::bigint")
     end
   end
 end


db/migrate/20210219212931_change_cursor_to_string.rb
```diff
@@ -12,7 +12,7 @@ def up
 
   def down
     change_table(:maintenance_tasks_runs) do |t|
-      t.change(:cursor, :bigint)
+      t.change(:cursor, "bigint USING cursor::bigint")
```
nvasilevski, line 15: This is postgresql specific, right? Which makes it broken for any other database. We should look for a solution that works for all adapters or branch in the worst case

ibrahima: Ohhh, you're right, I forgot about that. I think the CAST function should be cross-platform.

ibrahima:
"I have signed the CLA!"

PR #1080, opened 2024-09-09
Add generator option to specify path to generate tasks in

guillermoap:
This PR introduces a new tasks_path option to the Task Generator, allowing users to specify a custom path for generating task files. Currently the only namespacing available is inside the `app/tasks/` folder which I think lacks some flexibility when organizing an app with different directories.


guillermoap:
I have signed the CLA!

PR #1113, opened 2024-11-10
Reference query params to set task attribute values

derek-meulmeester:
It can be useful to link to a maintenance task with task values pre-populated. For example if in an internal app we want to be able to run a maintenance task on a customer, linking to `/maintenance_tasks/tasks/Maintenance::CustomerTask?customer_id=123` from the `/internal/customers/123` route and having the attribute filled in makes it easier and less error prone.

This PR sets the attribute field value if there's a corresponding query param with the same name.
If no query param is provided then we'll continue to respect the default values set on the task attribute.

Here's an example with some prefilled values via query param:
Screenshot 2024-11-09 at 9 48 40 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -485,6 +485,12 @@ to run. Since arguments are specified in the user interface via text area
 inputs, it’s important to check that they conform to the format your Task
 expects, and to sanitize any inputs if necessary.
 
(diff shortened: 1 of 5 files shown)


derek-meulmeester:
I have signed the CLA!

Issue #1116, opened 2024-11-13
Don't lose form data when there is a validation error


etiennebarrie:
Currently if there's a validation on a task and it fails, all of the form data is lost. It's a pretty bad user experience.

We should do like a well-behaved Rails application does and have each field filled with its previously-submitted value.

PR #1117, opened 2024-11-15
Use an existing method for validating params in ParamsTask

etiennebarrie:
While I was debugging some changes for #1116, I noticed that if anything is entered in `text_integer_attr2`, we get a 500 error.


etiennebarrie:
Yes I just noticed it wasn't related to the other PR, so I thought I'd fix it separately, but if we're changing it there then there's no need for this.

PR #1121, opened 2024-11-19
Add direct upload

ehasrouni:
Adds the ability to enable [direct uploads](https://guides.rubyonrails.org/v7.1/active_storage_overview.html#direct-uploads), which can help circumvent upload limits on the server. This is helpful when attempting to upload larger CSV files without having to play with the [max size of 20mb](https://github.com/ShopifyRestricted/gcp-chef/blob/872b0394b27f44825f1f65722892d882f88582e8/cookbooks/nginx/templates/default/nginx.conf.erb#L56)

re-attempts https://github.com/Shopify/maintenance_tasks/pull/1031

### DIFF
--- a/test/system/maintenance_tasks/runs_test.rb
+++ b/test/system/maintenance_tasks/runs_test.rb
@@ -60,10 +60,34 @@ module MaintenanceTasks
       MaintenanceTasks.metadata = nil
     end
 
-    test "run a CSV Task" do
+    test "run a CSV Task without direct upload" do
+      Rails.application.config.stubs(:respond_to?).with(:assets).returns(false)
       visit maintenance_tasks_path
 
       click_on("Maintenance::ImportPostsTask")
+
+      assert_selector("input[type=file]") do |input|
+        assert_nil(input["data-direct-upload-url"])
+      end
+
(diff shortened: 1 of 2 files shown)


test/system/maintenance_tasks/runs_test.rb
```diff
@@ -72,6 +77,26 @@ class RunsTest < ApplicationSystemTestCase
...
+        visit maintenance_tasks_path
+
+        click_on("Maintenance::ImportPostsTask")
+
+        attach_file("csv_file", "test/fixtures/files/sample.csv")
+        click_on "Run"
+        upload_url = page.server_url + "/rails/active_storage/direct_uploads"
+
+        assert_selector("input[type=file]") do |input|
+          assert_equal(input["data-direct-upload-url"], upload_url)
```
ehasrouni, line 91: `input["data-direct-upload-url"]` is returning nil for 'Rails <=7.0'

More specifically, it is failing here from the looks of it: https://github.com/rails/rails/blob/c11312a918555aaad99b334d374a791cd6268152/actionview/lib/action_view/helpers/form_tag_helper.rb#L1078

PR #1046, opened 2024-07-06
Add parameters from web

yamitake:
I love this gem and use it frequently in my projects.

This update allows parameters passed from the browser to be reflected in the form on the Task screen. 
For example, when accessing the URL 
`http://localhost:3000/maintenance_tasks/tasks/Maintenance::ParamsTask?post_ids=739&content=custom_content`, 

the form will automatically set the values of post_ids and content.

スクリーンショット 2024-07-06 13 43 51

PR #1123, opened 2024-11-20
Prefill Task parameters with query params

etiennebarrie:
There have been multiple attempts at adding filling the form with query params (#1046, #1051, #1113). This one uses the Run and Task objects, and solves #1116 at the same time.

This is based on #1122 and just adds what's needed to support query params filling the form.

### DIFF
--- a/app/controllers/maintenance_tasks/tasks_controller.rb
+++ b/app/controllers/maintenance_tasks/tasks_controller.rb
@@ -18,7 +18,11 @@ module MaintenanceTasks
     # Renders the page responsible for providing Task actions to users.
     # Shows running and completed instances of the Task.
     def show
-      @task = TaskDataShow.prepare(params.fetch(:id), runs_cursor: params[:cursor])
+      @task = TaskDataShow.prepare(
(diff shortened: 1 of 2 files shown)


app/controllers/maintenance_tasks/tasks_controller.rb
```diff
@@ -18,7 +18,11 @@ def index
     # Renders the page responsible for providing Task actions to users.
     # Shows running and completed instances of the Task.
     def show
-      @task = TaskDataShow.prepare(params.fetch(:id), runs_cursor: params[:cursor])
+      @task = TaskDataShow.prepare(
+        params.fetch(:id),
+        runs_cursor: params[:cursor],
+        arguments: params.except(:id, :cursor, :controller, :action).permit!,
```
etiennebarrie, line 24: I just realized this prevents Tasks from having a `cursor` attribute, so I'll remove `:cursor` from this list and the cursor will be used both for the runs and the Task inputs. Which means it may end up showing an empty list of runs below the form to start a task, but I think that's ok.

PR #1125, opened 2024-11-22
Add Ruby dependency in Gemfile

etiennebarrie:
This adds a Ruby requirement to the main Gemfile in order to make sure Dependabot/Bundler uses at least Ruby 3.2.0 to resolve dependencies. I guess we could also use `.ruby-version` directly.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,8 +11,9 @@ gem "mocha"
 gem "puma"
 if !@rails_gem_requirement
   gem "rails"
+  ruby ">= 3.2.0"
 else
-  # causes Dependabot to ignore the next line and update the next gem "rails"
+  # causes Dependabot to ignore the next line and update the previous gem "rails"
   rails = "rails"
   gem rails, @rails_gem_requirement
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -293,5 +293,8 @@ DEPENDENCIES
   sqlite3
   yard
 
+RUBY VERSION
+   ruby 3.3.0p0
+
 BUNDLED WITH
    2.4.6


adrianna-chang-shopify (approved):

etiennebarrie:
I tried this on main, because Dependabot was unable to create a security update for `rails-html-sanitizer`. Copying from https://github.com/Shopify/maintenance_tasks/actions/runs/12137201207/job/33839915704 because that won't be kept forever:

```
Could not find compatible versions

Because rails >= 7.2.1.2, < 7.2.2 depends on actionmailer = 7.2.1.2
  and rails >= 7.2.1.1, < 7.2.1.2 depends on actionmailer = 7.2.1.1,
  rails >= 7.2.1.1, < 7.2.2 requires actionmailer = 7.2.1.1 OR = 7.2.1.2.
And because rails >= 7.2.1, < 7.2.1.1 depends on actionmailer = 7.2.1,
... 500 lines
And because the current Bundler version (2.5.16) does not satisfy bundler >= 1.3.0, < 2.0
  and rails >= 8.0.0.beta1 depends on Ruby >= 3.2.0,
  every version of maintenance_tasks requires Ruby >= 3.2.0.
So, because Gemfile depends on maintenance_tasks >= 0
  and current Ruby version is = 3.1.4,
  version solving has failed.

Your bundle requires a different version of Bundler than the one you're running.
Install the necessary version with `gem install bundler:2.0.0.pre.3` and rerun bundler using `run.rb _2.0.0.pre.3_`
```

But it didn't work either. I'm just going to upgrade manually for now.

PR #1131, opened 2024-12-11
Support UUID primary keys

gregpardo:
The current listing of completed tasks assumes the primary key is an integer. 

See relevant issue
https://github.com/Shopify/maintenance_tasks/issues/462

This change is intended to support any primary key and adapter. 

Instead of getting the latest completed by MAX(id), we instead:
1. Group the completed tasks by task name and select the latest ended_at (should always be set on completed tasks)
2. Join back on task_name and latest_ended_at date for the full result

There could be a slight performance impact over the MAX(id) version I believe.

The tests appear to be covered by the tasks_tests.rb

### DIFF
--- a/app/models/maintenance_tasks/run.rb
+++ b/app/models/maintenance_tasks/run.rb
@@ -68,6 +68,13 @@ module MaintenanceTasks
       with_attached_csv_file if ActiveStorage::Attachment.table_exists?
     end
 
+    scope :last_completed_per_task, -> do
+      subquery = completed
+                   .select('task_name, MAX(ended_at) AS latest_ended_at')
+                   .group(:task_name)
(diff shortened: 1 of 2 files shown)


etiennebarrie:
Could we keep the existing behavior when not using uuids? There's no index on `ended_at`, that worries me a bit.

Also just making it clear here, looking at `ended_at` changes the behavior (probably for the better) if you have a long run, then a short run that finishes first. Currently we'll show the long run while it's still running, then the short run once they're all completed (since it's the last run that's completed). With these changes, we'll show the long run while it's running, and it will still be shown once it's completed.

gregpardo:
What about using `created_at`. It looks like there does appear to be an index on that. Since we are already scoping by completed that would give the latest created that is completed. Seems it would work the same as the existing solution?

PR #1150, opened 2025-01-21
Introduce MaintenanceTasks::Task.report_on

andrewn617:
By default any errors encountered while processing an iteration will be raised and the task with error. In our app we noticed a lot developers writing very similar code to rescue some exceptions, report them and moving on to the next iteration. So, let's introduce a helper to make this easier.

We introduced a `report_on` method to `MaintenanceTask::Tasks`. Exceptions passed to this method will be rescued during iteration and reported the rails error reporter, then iteration will continue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1017,6 +1017,30 @@ Note that `context` may be empty if the Task produced an error before any
 context could be gathered (for example, if deserializing the job to process
 your Task failed).
 
+#### Reporting errors during iteration
+
+By default, errors raised during task iteration will be raised to the application
+and iteration will stop. However, you may want to handle some errors and continue
+iteration. `MaintenanceTasks::Task.report_on` can be used to rescue certain
+exceptions and report them to the Rails error reporter.
+
+```ruby
+class MyTask < MaintenanceTasks::Task
+  report_on(MyException)
+end
+```
+
+`MaintenanceTasks::Task` also includes `ActiveSupport::Rescuable` which you can use
+to implement custom error handling.
+
+```ruby
+class MyTask < MaintenanceTasks::Task
+  rescue_from(MyException) do |exception|
+    handle(exception)
+  end
+end
+```
+
 #### Customizing the maintenance tasks module
 
 `MaintenanceTasks.tasks_module` can be configured to define the module in which
--- a/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
+++ b/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
@@ -112,7 +112,7 @@ module MaintenanceTasks
       end
     rescue => error
       @errored_element = input
-      raise error
+      raise error unless @task.rescue_with_handler(error)
     end
 
     def before_perform
--- a/app/models/maintenance_tasks/task.rb
+++ b/app/models/maintenance_tasks/task.rb
@@ -8,6 +8,7 @@ module MaintenanceTasks
     include ActiveModel::Attributes
     include ActiveModel::AttributeAssignment
     include ActiveModel::Validations
+    include ActiveSupport::Rescuable
 
     class NotFoundError < NameError; end
 
@@ -198,6 +199,20 @@ module MaintenanceTasks
       #   (see https://api.rubyonrails.org/classes/ActiveSupport/Callbacks/ClassMethods.html#method-i-set_callback)
       def after_error(*filter_list, &block)
         set_callback(:error, :after, *filter_list, &block)
+      end
+
+      # Rescue listed exceptions during an iteration and report them to the error reporter, then
+      # continue iteration.
+      #
+      # @param exceptions list of exceptions to rescue and report
+      def report_on(*exceptions)
+        rescue_from(*exceptions) do |exception|
+          if Rails.gem_version >= Gem::Version.new("7.1")
+            Rails.error.report(exception, source: "maintenance_tasks")
+          else
+            Rails.error.report(exception, handled: true)
+          end
+        end
       end
 
       private
--- a/test/jobs/maintenance_tasks/task_job_test.rb
+++ b/test/jobs/maintenance_tasks/task_job_test.rb
@@ -719,6 +719,23 @@ module MaintenanceTasks
       TaskJob.perform_now(run)
     end
 
+    test ".report_on reports the error" do
+      run = Run.create!(task_name: "Maintenance::UpdatePostsTask")
+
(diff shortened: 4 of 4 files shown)


app/jobs/concerns/maintenance_tasks/task_job_concern.rb
```diff
@@ -112,7 +112,11 @@ def task_iteration(input)
       end
     rescue => error
       @errored_element = input
-      raise error
+      if @task.exception_handlers.keys.include?(error.class)
```
nvasilevski, line 115: Do we have an option to split it into two `rescue`, like:
```ruby
    rescue *@task.exception_handlers.keys => error
      @errored_element = input
      @task.exception_handlers[error.class].call(error)
    rescue => error
      @errored_element = input
      raise error
```

`@errored_element = input` has to be duplicated so not sure if it's much cleaner. So totally up to you

app/models/maintenance_tasks/task.rb
```diff
@@ -200,6 +202,20 @@ def after_error(*filter_list, &block)
...
       end
 
+      # Rescue listed exceptions during an iteration. By default the rescued exception will be logged with
+      # Rails.error.report. A block can be supplied for custom handling behaviour.
+      #
+      # @param exceptions [Class] list of exceptions to rescue
+      # @yield [error] Block to be invoked with the exception instance
+      # @yieldparam error [StandardError] Exception instance
+      def rescue_from(*exceptions, &block)
+        handler = block_given? ? block : ->(e) { Rails.error.report(e) }
```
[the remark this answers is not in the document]
andrewn617: Yeah, I think that's a good idea, and then I can put the rails version check in there and kill two birds with one stone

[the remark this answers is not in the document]
andrewn617: I like the idea of letting users configure the default error handler, but I am not sure how to implement. We already have `MaintenanceTask.error_handler` which called when the job itself errors out. So ideally I'd like to re-use that and not introduce two error handlers.

However, since this error is being handled within the task itself, it does not have access to the job and the run to provide the right params to the error handler ([see here](https://github.com/Shopify/maintenance_tasks/blob/8cdc2db43a68013beb4d2504f964413f1260c983/app/jobs/concerns/maintenance_tasks/task_job_concern.rb#L192)). So it won't seem to work.

@adrianna-chang-shopify thoughts?

andrewn617:
@kmcphillips I added support for 6.1 (and tested it locally) but ci just seems broken for 6.1 🤔 So maybe we _should_ just remove it.

README.md
```diff
@@ -407,6 +407,34 @@ module Maintenance
 end
 ```
 
+### Handling errors during iteration
+
+By default any errors encountered while processing an iteration will be raised and
+the task with error. Errors can be rescued using the `rescue_from` handler. Errors
+rescued this way will be reported by the handler and iteration will continue.
```
adrianna-chang-shopify, line 414: Couple small wording tweaks here
```suggestion
By default, any errors encountered while processing an iteration will raise and
the task will error. Errors can be rescued using the `rescue_from` method, and configuring
a handler. The handler will receive the rescued error and iteration will continue.
```

README.md
```diff
@@ -407,6 +407,34 @@ module Maintenance
...
+
+The default handler just calls `Rails.error.report` with the raised error.
```
adrianna-chang-shopify, line 416:
```suggestion
The default handler calls `Rails.error.report` with the raised error:
```

README.md
```diff
@@ -407,6 +407,34 @@ module Maintenance
...
+
+```ruby
+module Maintenance
+  class UpdatePostRescueErrorTask < MaintenanceTasks::Task
+    rescue_from ActiveRecord::RecordNotUnique
+  end
+end
+```
+
+Or you can pass a custom handler with a block.
```
adrianna-chang-shopify, line 426:
```suggestion
Or you can pass a custom handler with a block:
```

README.md
```diff
@@ -407,6 +407,34 @@ module Maintenance
...
+```ruby
+module Maintenance
+  class UpdatePostRescueErrorTask < MaintenanceTasks::Task
+    rescue_from ActiveRecord::RecordNotUnique do |error|
+      CustomErrorHandler.report(error)
+    end
+  end
+end
+```
+
```
adrianna-chang-shopify, line 437: Maybe we should add a note indicating that if the handler raises, the whole task _will_ error?

test/jobs/maintenance_tasks/task_job_test.rb
```diff
@@ -707,5 +707,45 @@ class << self
 
       TaskJob.perform_now(run)
     end
+
+    test ".rescue_on" do
```
adrianna-chang-shopify, line 711:
```suggestion
    test ".rescue_from" do
```

app/models/maintenance_tasks/task.rb
```diff
@@ -208,6 +226,14 @@ def load_constants
 
         Rails.autoloaders.main.eager_load_namespace(namespace)
       end
+
+      def default_handler(error)
+        if Rails::VERSION::MAJOR >= 7
```
kmcphillips, line 231: I still think we should just drop support below 7, but this won't affect many people so 🤷

andrewn617:
Turns out `assert_error_reported` is not available in 7.0 Instead of trying to have two versions of the test, I just wrote my own assertion to use and we can delete it when we drop support for 7.0

app/models/maintenance_tasks/task.rb
```diff
@@ -200,6 +201,20 @@ def after_error(*filter_list, &block)
...
       end
 
+      # Rescue listed exceptions during an iteration and report them to the error reporter, then
+      # continue iteration.
+      #
+      # @param exceptions list of exceptions to rescue and report
+      def report_on(*exceptions)
+        rescue_from(*exceptions) do |exception|
+          if Rails.gem_version >= Gem::Version.new("7.1")
+            Rails.error.report(exception, source: "maintenance-tasks")
```
etiennebarrie, line 211:
```suggestion
            Rails.error.report(exception, source: "maintenance_tasks")
```
Let's use the gem name I guess? There's not much guidance on the source parameter but I think it makes more sense to use the exact name of the gem.

etiennebarrie (approved):
Sounds really useful, thanks!

BTW it feels like `report_on` would have its place in `ActiveSupport::Rescuable`.

adrianna-chang-shopify (approved):
Nice! 👏

PR #1151, opened 2025-01-22
Remove support for Rails 6.1

andrewn617:
Drop support for Rails 6.1, it's EOL now.

PR #1152, opened 2025-01-22
Deprecate MaintenanceTasks.error_handler. Instead use the Rails.error…

andrewn617:
Rails 7.0 introduced the [rails error reporter](https://guides.rubyonrails.org/error_reporting.html). Now that we are [dropping support for 6.1](https://github.com/Shopify/maintenance_tasks/pull/1151) we can modernize this gem and use the error reporter instead of the error handler.

This PR deprecates the error handler, and reports exceptions to the rails error reporter instead - unless the app is still setting a handler, in which case we preserve that behaviour but fire off a deprecation warning.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,9 +74,9 @@ The generator creates and runs a migration to add the necessary table to your
 database. It also mounts Maintenance Tasks in your `config/routes.rb`. By
 default the web UI can be accessed in the new `/maintenance_tasks` path.
 
-In case you use an exception reporting service (e.g. Bugsnag) you might want to
-define an error handler. See [Customizing the error
-handler](#customizing-the-error-handler) for more information.
+This gem uses the [Rails Error Reporter](https://guides.rubyonrails.org/error_reporting.html) to report errors. If you are using a bug
+tracking service you may want to subscribe to the reporter. See [Reporting Errors](#reporting-errors)
+for more information.
 
 ### Active Job Dependency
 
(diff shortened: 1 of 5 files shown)


lib/maintenance_tasks.rb
```diff
@@ -94,4 +84,29 @@ module MaintenanceTasks
...
+    # no-doc
+    def error_handler=(proc)
+      deprecator.warn(DEPRECATION_MESSAGE)
+
+      @error_handler = proc
+    end
+
+    # no-doc
+    def deprecator
+      @deprecator ||= ActiveSupport::Deprecation.new
```
adrianna-chang-shopify, line 109: Should we add the target deprecation horizon + library name? We'd probably need to target the next major version here (3.0)

lib/maintenance_tasks.rb
```diff
@@ -94,4 +84,29 @@ module MaintenanceTasks
   #
   #  @return [ActiveSupport::Duration] the threshold in seconds after which a task is considered stuck.
   mattr_accessor :stuck_task_duration, default: 5.minutes
+
+  class << self
+    DEPRECATION_MESSAGE = "MaintenanceTasks.error_handler is deprecated and will be removed in the next release. " \
+      "Instead, reports will be sent to the Rails error reporter. Do not set a handler and subscribe" \
+      "to the error reporter instead."
+
+    # no-doc
```
adrianna-chang-shopify, line 93: Should we use the [`@deprecated` directive](https://rubydoc.info/gems/yard/file/docs/Tags.md#deprecated) for these instead?

adrianna-chang-shopify (commented):
Looking good, a few small comments!

lib/maintenance_tasks.rb
etiennebarrie: Yes also we want to add it to the application's deprecators when the engine is initialized.

README.md
```diff
@@ -988,44 +988,32 @@ If you are stuck in `pausing` and wish to preserve your tasks's position
...
-  end
-end
-```
-
-The error handler should be a lambda that accepts three arguments:
+Reports to the error reporter will contain the following data:
 
 * `error`: The exception that was raised.
-* `task_context`: A hash with additional information about the Task and the
+* `context`: A hash with additional information about the Task and the
```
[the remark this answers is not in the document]
andrewn617: It doesn't need to be serialized, so I think I'll just pass the whole run instance?

test/jobs/maintenance_tasks/task_job_test.rb
```diff
@@ -707,5 +718,35 @@ class << self
 
       TaskJob.perform_now(run)
     end
+
+    private
+
+    # TODO: Use rails built in version of this assertion once we drop support for 7.0
+    def assert_error_reported(error = nil, &block)
```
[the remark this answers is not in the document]
andrewn617: Good idea

adrianna-chang-shopify (approved):

PR #1154, opened 2025-01-23
Make task run `created_at` timestamps consistent

larouxn:
👋 Was recently looking at the on-hover content for run `created_at` values and noticed it differs between the `task` and `run` view partials. Personally I find the `UTC` appended format more intuitive than `iso8061` hence the change. Hence, I'm proposing we use the same format for both views. That said we could move to all `iso8061` if that's preferable. The underlying goal is consistency.

This is effectively a no-op change for the task view and a small visual change for the run view. For reference here are the four timestamp variants at the code level.

```ruby
irb(main):010> MaintenanceTasks::Run.all.first.created_at
  MaintenanceTasks::Run Load (0.9ms)  SELECT `maintenance_tasks_runs`.* FROM `maintenance_tasks_runs` ORDER BY `maintenance_tasks_runs`.`id` ASC LIMIT 1
=> 2025-01-20 01:29:40.451906000 UTC +00:00
irb(main):011> MaintenanceTasks::Run.all.first.created_at.utc
  MaintenanceTasks::Run Load (0.6ms)  SELECT `maintenance_tasks_runs`.* FROM `maintenance_tasks_runs` ORDER BY `maintenance_tasks_runs`.`id` ASC LIMIT 1
=> 2025-01-20 01:29:40.451906 UTC
irb(main):012> MaintenanceTasks::Run.all.first.created_at.utc.iso8601
  MaintenanceTasks::Run Load (0.6ms)  SELECT `maintenance_tasks_runs`.* FROM `maintenance_tasks_runs` ORDER BY `maintenance_tasks_runs`.`id` ASC LIMIT 1
=> "2025-01-20T01:29:40Z"
```

🎩 🎩 🎩 🎩 🎩 

**Task**

Before
tasks_before

After
tasks_after

**Run**

Before
run_before

After
run_after


etiennebarrie (approved):
Thanks!

PR #1169, opened 2025-02-17
[FEATURE] Mask sensitive run arguments

3assy2018:
In some cases, maintenance tasks require the use of sensitive data in their arguments; however, we noticed that there should be a protection layer to hide these sensitive data from the UI once the task is submitted to run.
This PR introduces a new class method in the maintenance tasks to mask any sensitive attributes and hide them in the arguments list in the UI.

[TBD] The PR also assumes that the Rails global parameter filters contain sensitive keys, which should be automatically taken into account, and redacted globally across maintenance tasks.

The final result should look like the following screenshot:
Screenshot 2025-02-17 at 17 14 45


app/helpers/maintenance_tasks/tasks_helper.rb
```diff
@@ -182,5 +182,18 @@ def datetime_field_help_text
         class: "content is-small",
       )
     end
+
+    # Mask sensitive information from the arguments of a task.
+    #
+    # @param run [Run] the Run that contains the arguments to be masked.
+    def mask_arguments(run)
```
adrianna-chang-shopify, line 189: What do you think about having the `Run` handle this instead? That would allow us to memoize the parameter filter.

```ruby
module MaintenanceTasks
  class Run < ApplicationRecord
    ...
    def masked_arguments
      argument_filter.filter(arguments)
    end

    private

    def argument_filter
      @argument_filter ||= ParameterFilter.new(Rails.application.config.filter_parameters + task.masked_attributes)
    end
```

app/models/maintenance_tasks/task.rb
```diff
@@ -28,6 +28,11 @@ class NotFoundError < NameError; end
     # @api private
     class_attribute :collection_builder_strategy, default: NullCollectionBuilder.new
 
+    # the sensitive attributes that will be filtered when fetching a run.
```
adrianna-chang-shopify, line 31:
```suggestion
    # The sensitive attributes that will be filtered when fetching a run.
```

README.md
```diff
@@ -509,6 +509,31 @@ following types are supported:
...
+module Maintenance
+  class SensitiveParamsTask < MaintenanceTasks::Task
+    attribute :sensitive_content, :string
+    
+    mask_attribute :sensitive_content
+  end
+end
+```
+
+If you have any filtered parameters in the global rails parameter filter, they will be
```
adrianna-chang-shopify, line 529: What do you think about linking to the [docs](https://guides.rubyonrails.org/configuring.html#config-filter-parameters) here?

3assy2018: Great idea, thanks for flagging it! I added it in the latest commit

adrianna-chang-shopify (commented):
Looks good to me! Can you squash your commits and then I'll merge?

3assy2018:
Hey Shopifolk!
Thanks a lot for your collaboration, and happy to hear that the feature looks good.

@adrianna-chang-shopify I have tried to squash the commits as you requested, but I ended up having a strange commit history, but it it squashed anyways.
Let me know if you have any further requests 🙇‍♂️

PR #1170, opened 2025-02-18
correct error reporting source

buckley-w-david:
After seeing the deprecation message introduced in #1152 I began investigating how to migrate from `error_handler` to subscribing to error notifications. In the process I noticed that `README.md` disagrees with `task_job_concern.rb` in what `source` the errors will be reported with - `maintenance_tasks` vs `maintenance-tasks` https://github.com/Shopify/maintenance_tasks/blob/7505b9c64942cab305b5bec91ec2a5bb73546915/app/jobs/concerns/maintenance_tasks/task_job_concern.rb#L194


adrianna-chang-shopify (approved):
Thank you!

PR #1178, opened 2025-03-06
Apply ID to each run to take advantage of anchor links directly to a run

hllewelyn:
We have some tasks that are ran multiple times, and occasionally we need to reference a specific run. 

This is currently only possible by quoting the run ID which the recipient then needs to search for in the list, so I thought it would be helpful to simply apply that ID to the run wrapping box to be able to use anchor links to link directly to that run.

### DIFF
--- a/app/views/maintenance_tasks/runs/_run.html.erb
+++ b/app/views/maintenance_tasks/runs/_run.html.erb
@@ -2,7 +2,7 @@
   <h5 class="title is-5">
     <%= time_tag run.created_at, title: run.created_at.utc %>
     <%= status_tag run.status %>
-    <span class="is-pulled-right" title="Run ID">#<%= run.id %></span>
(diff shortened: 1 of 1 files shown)


hllewelyn:
> It's a good idea, thanks! Is there a link somewhere where we could right-click and copy link or do we have to manually write the id in the URL?

Currently @etiennebarrie we would have to manually write the ID in the URL. I'm definitely open to turning the existing Job ID  into a link though, or perhaps even the title without the status? 

I think my preference would be the former - what do you think?

hllewelyn:
I have signed the CLA!

I agree, I'll make the run ID on the right the permalink.

hllewelyn:
I've added the permalink @etiennebarrie 👍🏻 

CleanShot 2025-03-06 at 16 44 18@2x

PR #1184, opened 2025-03-14
feat: added required field indicator

candidosales:
- [x] Added `required` label in the required fields

CleanShot 2025-03-17 at 12 28 56@2x


candidosales:
@nvasilevski, could you review it?

app/views/maintenance_tasks/runs/_serializable.html.erb
```diff
@@ -9,7 +9,7 @@
             <% if !value.empty? %>
               <% if value.include?("\n") %>
                 <pre><%= value %></pre>
-        	  <% else %>
+              <% else %>
```
candidosales, line 12: After changing how to render the space in the VSCode, I found the issue 🙏🏼 .

candidosales:
@etiennebarrie , could you review it?

app/controllers/maintenance_tasks/application_controller.rb
```diff
@@ -11,7 +11,7 @@ class ApplicationController < MaintenanceTasks.parent_controller.constantize
       policy.style_src(
         BULMA_CDN,
         # ruby syntax highlighting
-        "'sha256-2AM66zjeDBmWDHyVQs45fGjYfGmjoZBwkyy5tNwIWG0='",
+        "'sha256-WHHDQLdkleXnAN5zs0GDXC5ls41CHUaVsJtVpaNx+EM='",
```
nvasilevski, line 14: Was this intentional? Let's add an explanation to why we are changing this

nvasilevski (approved):
I think this makes sense, pretty straightforward improvement. Let's have one more review

app/helpers/maintenance_tasks/tasks_helper.rb
```diff
@@ -190,5 +190,17 @@ def datetime_field_help_text
...
     end
+
+    # Checks if an attribute is required for a given Task.
+    #
+    # @param task_data_show [MaintenanceTasks::TaskDataShow] The TaskDataShow instance.
+    # @param attribute_name [Symbol] The name of the attribute to check.
+    # @return [Boolean] Whether the attribute is required.
+    def attribute_required?(task_data_show, attribute_name)
+      model_class = task_data_show.name.constantize
+      model_class.validators_on(attribute_name).any? do |validator|
```
adrianna-chang-shopify, line 201: Can we pass in the form object (task) here instead? Would be good to keep the naming consistent with `#resolve_inclusion_value` as well:
```suggestion
    def attribute_required?(task, parameter_name)
      task_class = task.class
      task_class.validators_on(parameter_name).any? do |validator|
```

app/controllers/maintenance_tasks/application_controller.rb
adrianna-chang-shopify: This is changing because we added a new CSS rule to `app/views/layouts/maintenance_tasks/application.html.erb`

app/helpers/maintenance_tasks/tasks_helper.rb
adrianna-chang-shopify: This method will also need a test please.

adrianna-chang-shopify (commented):
Changes make sense, a couple small comments. Thanks!

candidosales:
@adrianna-chang-shopify , could you review it? 🙏🏼

app/helpers/maintenance_tasks/tasks_helper.rb
```diff
@@ -190,5 +190,17 @@ def datetime_field_help_text
...
+        validator.kind == :presence
```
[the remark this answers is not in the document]
candidosales: I have tried use but not works as expected:

```ruby
def attribute_required?(task, parameter_name)
    task.class.validators_on(parameter_name).any? do |validator|
    puts "Validator: #{validator.options.inspect}"
    validator.options[:presence] == true
    end
end
```

I have inspected the options and it isn't showing `presence`

```bash
Validator: {}
Validator: {with: /\A(\s?\d+(,\s?\d+\s?)*)\z/, allow_blank: true}
Validator: {in: [100, 200, 300], allow_nil: true}
Validator: {in: #<Proc:0x0000000145b92760 /Users/candidogomes/Documents/Projects/maintenance_tasks/test/dummy/app/tasks/maintenance/params_task.rb:41>, allow_nil: true}
Validator: {in: #<Proc:0x0000000145b92328 /Users/candidogomes/Documents/Projects/maintenance_tasks/test/dummy/app/tasks/maintenance/params_task.rb:42>, allow_nil: true}
Validator: {in: :dropdown_attr_options, allow_nil: true}
Validator: {in: Maintenance::DropdownOptions, allow_nil: true}
Validator: {within: [true, false], allow_nil: true}
Validator: {in: 100.., allow_nil: true}
```

app/views/layouts/maintenance_tasks/application.html.erb
```diff
@@ -54,6 +54,11 @@
         box-shadow: 0 4px 6px -1px #0000001a,
                     0 2px 4px -2px #0000001a;
       }
+      .label.is-required:after {
+        content: " (required)";
+        color: #ff6685;
+        font-size: 12px;
+      }
```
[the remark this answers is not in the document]
candidosales: Yeah, I merged it, but I haven't updated it yet. I'll do it soon

test/helpers/maintenance_tasks/tasks_helper_test.rb
```diff
@@ -102,8 +102,6 @@ class TasksHelperTest < ActionView::TestCase
       ].each do |attribute|
         assert_match "Select a value", markup(attribute).squish
       end
-
-      refute_match "Select a value", markup("text_integer_attr_unbounded_range").squish
```
etiennebarrie, line 106: No reason to remove this.

etiennebarrie (commented):
Just one last change and it looks good to me.

adrianna-chang-shopify:
Oh actually seems like we have some test failures, can you investigate?

candidosales:
@adrianna-chang-shopify, I'm using the same command `bundle exec rake test TESTOPTS="--seed 35438"` and no issue 

CleanShot 2025-04-15 at 12 03 57@2x

candidosales:
I can replicate with `bundle exec rake test TESTOPTS="--seed 17954"` ... investigating

adrianna-chang-shopify:
This is still going to cause issues depending on the order in which the tests run (the constant may or may not be loaded by the time one of "list all tasks" tests runs).

Perhaps we can reuse one of the existing test tasks (in `test/dummy/app/tasks`) instead?

etiennebarrie (approved):

PR #1189, opened 2025-03-20
System test robustness

etiennebarrie:
We paired with @adrianna-chang-shopify on improving the stability of system tests.

We found a few issues related the code, but a lot of it is related to changes to Chrome 134 it seems:

https://github.com/SeleniumHQ/selenium/issues/15401
https://github.com/teamcapybara/capybara/issues/2800

Last commit is a hack to make sure Capybara retries when the element is no longer on the page because it's something it's always done, and now it just considers them unknown errors. This will need to be reverted once things have settled.

### DIFF
--- a/test/application_system_test_case.rb
+++ b/test/application_system_test_case.rb
@@ -35,6 +35,9 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
   setup do
     travel_to Time.zone.local(2020, 1, 9, 9, 41, 44)
     page.driver.browser.download_path = "test/dummy/tmp/downloads"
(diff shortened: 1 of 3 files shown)


etiennebarrie:
We've also tried to only apply the hack, but the other changes were necessary because the tests were flaky in the first place, see https://github.com/Shopify/maintenance_tasks/issues/1176.

etiennebarrie:
Actually the commit that disables checking for the visibility of elements might not be a good idea. We don't use invisible elements at all so I'll let that split, but for reference for the future, if this cause an issue, feel free to revert 21053665ec17e08be66f176b0322c970971071ab.

PR #1163, opened 2025-02-07
feat: new UI, upgrade bulma 1.0.3

candidosales:
**Why**

When I first came into contact with the tool, I found the style a bit intimidating, and there were opportunities for improvement, so I'm creating this PR to improve the developer's life.

**What**

- [x] Upgrade [Bulma 1.0.3](https://github.com/jgthms/bulma/releases/tag/1.0.3)

## New home

Reduce the weight of the typography and improve the colours

CleanShot 2025-02-08 at 07 29 36@2x

## New task UI

Splitting the inputs and arguments sections into three columns will prevent the user from scrolling through the page.
 
CleanShot 2025-02-08 at 08 22 06@2x

## Saving spacing

I added accordion sections to the Source code and Runs. The first run section will always be open.

https://github.com/user-attachments/assets/ed45b2f4-5978-41c1-829b-e3b4b8d34c5a

Please feel free to share any feedback - I'd love to hear your perspective! 😊

PR #1195, opened 2025-03-25
Use more specific CSP directives

tdeo:
We wanted to try out this gem in our repo but we're using the `script-src-elem` directive in our CSP for `<script>` inline tags, which fallbacks to `script-src` if unset.

But since `script-src-elem` is not defined in maintenance-tasks, we're seeing the following error at this moment:

image

Opening this as draft for now since there are no tests yet

### DIFF
--- a/app/controllers/maintenance_tasks/application_controller.rb
+++ b/app/controllers/maintenance_tasks/application_controller.rb
@@ -7,13 +7,18 @@ module MaintenanceTasks
   class ApplicationController < MaintenanceTasks.parent_controller.constantize
     BULMA_CDN = "https://cdn.jsdelivr.net"
 
-    RUBY_SYNTAX_HIGHLIGHTING = "'sha256-2AM66zjeDBmWDHyVQs45fGjYfGmjoZBwkyy5tNwIWG0='"
-    PAGE_REFRESH_SCRIPT = "'sha256-NiHKryHWudRC2IteTqmY9v1VkaDUA/5jhgXkMTkgo2w='"
-
     content_security_policy do |policy|
-      policy.style_src_elem(BULMA_CDN, RUBY_SYNTAX_HIGHLIGHTING)
-      policy.script_src_elem(PAGE_REFRESH_SCRIPT)
+      policy.style_src_elem(
+        BULMA_CDN,
+        # <style> tag in app/views/layouts/maintenance_tasks/application.html.erb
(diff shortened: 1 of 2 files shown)


tdeo:
> What you should do is set the strictest policy possible in the dummy app, then use the smallest possible fix in the engine. In that case, if the dummy app disallows everything with `script-src-elem` instead of `script-src`, the engine can just allow our script element with `script-src-elem`.

Hi @etiennebarrie, thanks for the pointers! To be as conservative as possible, I did set `default-src(:none)` on the dummy app so that all missing directives fallback to that one and be as exhaustive as possible. 

I had to keep both `frame-src` and `style-src-elem` set for the dummy app iframe to load and for tests to remain green. I'm not sure I fully undestand why `MaintenanceTasks::ApplicationController` does override `policy.frame_ancestors(:self)`, inherently allowing iframing into the same app but nowhere else

tdeo:
Indeed, I did update the dummy app to list all possible CSP directives (all those available in ActionDispatch except the ones unsupported in Selenium), to ensure that indeed the directives are set as specifically and not thanks to fallbacks

app/controllers/maintenance_tasks/application_controller.rb
```diff
@@ -7,17 +7,16 @@ module MaintenanceTasks
   class ApplicationController < MaintenanceTasks.parent_controller.constantize
     BULMA_CDN = "https://cdn.jsdelivr.net"
 
+    RUBY_SYNTAX_HIGHLIGHTING = "'sha256-2AM66zjeDBmWDHyVQs45fGjYfGmjoZBwkyy5tNwIWG0='"
+    PAGE_REFRESH_SCRIPT = "'sha256-NiHKryHWudRC2IteTqmY9v1VkaDUA/5jhgXkMTkgo2w='"
+
     content_security_policy do |policy|
-      policy.style_src(
-        BULMA_CDN,
-        # ruby syntax highlighting
```
etiennebarrie, line 13: This needs to be updated though, it's doing more now with #1163. Feel free to leave it as is, otherwise it should say
```ruby
# <style> tag in app/views/layouts/maintenance_tasks/application.html.erb
```
and the script should say
```ruby
# <script> tag in app/views/layouts/maintenance_tasks/application.html.erb
```

etiennebarrie:
Just realized this broke Safari. Somehow it uses style-src even when style-src-elem is set.

tdeo:
Hello @etiennebarrie which safari are you using ? I do get working styles with the latest version 2.12.0 locally

PR #1202, opened 2025-04-07
Add `Task#metadata`

### LINKED ISSUE #1197: Expose `Run#metadata` to `Task` subclasses
Similar to https://github.com/Shopify/maintenance_tasks/pull/1112, would be interesting to supply the `run.metadata` hash to the `Task` instance when it's executed. 

A use case we have for this is to identify the user that enqueued the run for audit/logging purposes. We already have this on `MaintenanceTasks.metadata`:

```ruby

MaintenanceTasks.parent_controller = "Dev::ApplicationController"

MaintenanceTasks.metadata = ->() do
  { user_id: current_user.id }
end
```

But this object isn't available inside the task

```ruby
# app/tasks/maintenance/hello_task.rb

module Maintenance
  class HelloTask < MaintenanceTasks::Task
    no_collection

    def process
      user = User.find(metadata.fetch("user_id")) # <-- undefined local variable or method 'metadata' for an instance of Maintenance::HelloTask
      puts "User #{user.email} is invoking the task"
    end
  end
end
```

I can try to submit a patch for this if it's welcome

lucasmazza:
Closes #1197

### DIFF
--- a/app/models/maintenance_tasks/run.rb
+++ b/app/models/maintenance_tasks/run.rb
@@ -422,6 +422,8 @@ module MaintenanceTasks
         if task.attribute_names.any? && arguments.present?
           task.assign_attributes(arguments)
         end
+
+        task.metadata = metadata
         task
       rescue ActiveModel::UnknownAttributeError
         task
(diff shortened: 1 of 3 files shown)


lucasmazza:
:heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #1203, opened 2025-04-10
Support UUID primary keys

stephannv:
This solve 2 issues in apps using UUID as primary keys:

When using `maintenance_tasks_runs` table with integer ids we get an error when uploading a CSV to task because it will try to save an integer ID into active storage attachment table (polymorphic association with records).

We could change manually the migration to use UUID as primary key but we face another problem when listing completed runs because the code uses `completed_runs.select("MAX(id) as id")...` and we get the error: `ERROR: function max(uuid) does not exist`. So I changed to use created_at instead id.


rafaelfranca (approved):

adrianna-chang-shopify (approved):
Thanks!

PR #1216, opened 2025-05-11
feat: Add attribute descriptions to MaintenanceTasks

candidosales:
## Context

Add a new attribute to include a description for the field, enhancing context and accessibility for developers to understand how to use each input appropriately.

CleanShot 2025-05-10 at 21 12 23@2x

- Introduced the `attribute_description` method in `TasksHelper` to retrieve descriptions of task attributes.
- Updated `Task` model to support attribute descriptions during attribute definition.
- Enhanced the `show` view to display attribute descriptions when available.
- A test for the new `attribute_description` method was added to ensure correct functionality.

### DIFF
--- a/app/helpers/maintenance_tasks/tasks_helper.rb
+++ b/app/helpers/maintenance_tasks/tasks_helper.rb
@@ -199,5 +199,14 @@ module MaintenanceTasks
         validator.kind == :presence
       end
     end
+
+    # Returns the description for a given attribute
+    #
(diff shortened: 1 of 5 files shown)


candidosales:
@etiennebarrie and @adrianna-chang-shopify , could you review it?

adrianna-chang-shopify (commented):
I'm not sure about this, why not just add comments above each `attribute` in the task itself since we render the source code in the UI?

PR #1218, opened 2025-05-14
Configuration attribute to overwrite UI default refresh interval

kirkath:
By default when the maintenance_tasks UI is open in a browser, it refreshes every 3 seconds.
This is great, but sometimes a bit too ambitious. We wanted to set this to 30 seconds or even a minute in our application, but did not find a configuration parameter to do so.

Therefore I am creating this PR to add such a configuration parameter to the gem.

With the changes in this PR our main base applicaation can just specify 
```
MaintenanceTasks.default_refresh_rate = 30.seconds
```
in `initializers/maintenance_tasks.rb` and it will propagate.

### DIFF
--- a/app/views/layouts/maintenance_tasks/application.html.erb
+++ b/app/views/layouts/maintenance_tasks/application.html.erb
@@ -65,6 +65,7 @@
       function refresh() {
         const target = document.querySelector("[data-refresh]")
         if (!target || !target.dataset.refresh) return
+        const interval = parseInt(target.dataset.interval)
         window.setTimeout(() => {
           document.body.style.cursor = "wait"
           fetch(document.location, { headers: { "X-Requested-With": "XMLHttpRequest" } }).then(
@@ -80,7 +81,7 @@
             },
             error => location.reload()
           )
-        }, <%= MaintenanceTasks.default_refresh_rate.in_seconds * 1000 %>)
+        }, interval)
       }
       document.addEventListener('DOMContentLoaded', refresh)
     </script>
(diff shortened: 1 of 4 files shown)


app/views/layouts/maintenance_tasks/application.html.erb
```diff
@@ -80,7 +80,7 @@
             },
             error => location.reload()
           )
-        }, 3000)
+        }, <%= MaintenanceTasks.default_refresh_rate.in_seconds * 1000 %>)
```
etiennebarrie, line 83: We should probably set the value in `[data-refresh]`. This will fail the subresource integrity check for the `<script>` tag if a different value is used.

etiennebarrie: Yes, except I was thinking we could reuse the `data-refresh` attribute on the HTML side. And on the Ruby side we don't need two instance variables.
There's also the case of the `tasks#show` view, which is set in this spot: https://github.com/Shopify/maintenance_tasks/blob/03a38d78659888fa3b3a94402c506777db84878a/app/views/maintenance_tasks/tasks/show.html.erb#L45

With your current changes, it refreshes as fast as it can, which is a lot 😂 I guess the parseInt returns NaN and setTimeout interprets this as 0, as soon as possible.
I think we could do without the parsing anyway given [non-number delay values are silently coerced into numbers](https://developer.mozilla.org/en-US/docs/Web/API/Window/setTimeout#non-number_delay_values_are_silently_coerced_into_numbers).

etiennebarrie:
You will also need to update the SHA in ApplicationController: https://github.com/Shopify/maintenance_tasks/blob/03a38d78659888fa3b3a94402c506777db84878a/app/controllers/maintenance_tasks/application_controller.rb#L17-L18

The easiest way to do this is to run the code in a browser and use the value it gives you that would have been accepted.

PR #1227, opened 2025-06-02
Add back error reporting code example to README

joshuay03:
Ref: https://github.com/Shopify/maintenance_tasks/pull/1152

One 'gotcha' of Rails' error reporting subscribers is that they're 'global' i.e. all reported errors are reported to all subscribers, and it's up to the user to filter them as needed for custom per-source handling. This wasn't something users needed to think about when using `MaintenanceTasks.error_handler` - only errors reported by the gem would make their way through to your handling logic.

For this reason, I think providing an example on how to filter by the source would be beneficial to users migrating from the deprecated API.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1047,6 +1047,25 @@ Note that `context` may be empty if the Task produced an error before any
 context could be gathered (for example, if deserializing the job to process your
 Task failed).
 
+Here's an example custom subscriber to the Rails error reporter for integrating
+with an exception monitoring service (Bugsnag):
+
+```ruby
+# config/initializers/maintenance_tasks.rb
+
+class MaintenanceTasksErrorSubscriber
+  def report(error, handled:, severity:, context:, source: nil)
+    return unless source == "maintenance-tasks"
+
+    Bugsnag.notify(error) do |notification|
+      notification.add_metadata(:task, context)
+    end
+  end
+end
+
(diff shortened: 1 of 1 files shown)


etiennebarrie (approved):
> only errors reported by the gem would make their way through to your handling logic

Yes, but on the other hand application monitoring tools should eventually be changed to subscribe to the Rails error handler, and then you won't need a specific subscriber for the gem, errors from the gem will flow to the tool like all errors.

joshuay03:
> > only errors reported by the gem would make their way through to your handling logic
> 
> Yes, but on the other hand application monitoring tools should eventually be changed to subscribe to the Rails error handler, and then you won't need a specific subscriber for the gem, errors from the gem will flow to the tool like all errors.

I agree, and this is the direction we're going in as well. However, it's possible that users have custom logic in these handlers (e.g., adding additional task-specific metadata). Or, if users are migrating from the deprecated API, their first step might be to introduce a single subscriber for this particular scenario before they update the rest of their application, which might be a bit more involved (like in our case).

With this in mind, I thought the example might be useful. However, I don't have a strong opinion about it. I could mention something to indicate that, ideally, you'd have a single subscriber for your entire app, and that this example is specifically for filtering to just task-related errors. Let me know what you think.

etiennebarrie:
I think it's worth having in the README, it's clear enough that it does things custom for the engine. Thanks for your contribution!

PR #1241, opened 2025-06-26
Avoid incrementing in-memory lock version on tick count updates

adrianna-chang-shopify:
~This can lead to increased contention on writes to the `maintenance_tasks_runs` table. We mainly care about race conditions for run status updates; incrementing the lock version on every tick count update is unnecessary.~

As @etiennebarrie reminded me in the comments, this PR actually doesn't prevent increment the DB lock version, it just removes the code to keep the in-memory one in sync. This is still worth shipping since there's no real point in updating the in-memory attribute, since we immediately call `reload_status` after to refresh the lock version from the DB. We may want to re-evaluate relying on `#update_counters`.

### DIFF
--- a/app/models/maintenance_tasks/run.rb
+++ b/app/models/maintenance_tasks/run.rb
@@ -126,11 +126,6 @@ module MaintenanceTasks
         time_running: duration,
         touch: true,
       )
-      if locking_enabled?
-        locking_column = self.class.locking_column
-        self[locking_column] += 1
-        clear_attribute_change(locking_column)
-      end
     end
 
     # Marks the run as errored and persists the error data.
--- a/test/models/maintenance_tasks/run_test.rb
+++ b/test/models/maintenance_tasks/run_test.rb
@@ -169,17 +169,6 @@ module MaintenanceTasks
       assert_equal 12.2, run.time_running
     end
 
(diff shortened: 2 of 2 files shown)


nvasilevski (approved):

etiennebarrie (commented):
This doesn't change the fact that Active Record does increment it in `update_counters`, though.

It might just cause an object to think it's stale (because it didn't have the in-memory increment). Although `reload_status` being called in `JobConcern#each_iteration` means we do update the lock version at each iteration, and I'm not seeing stale object retries, so it also means this in-memory update is not necessary.

adrianna-chang-shopify:
> This doesn't change the fact that Active Record does increment it in update_counters, though.

Right, I forgot it did that, and this code was about updating the in-memory lock version to keep the two in sync :/ 
Agree that it's worth shipping this regardless since it doesn't appear to  be necessary, but this is insufficient to actually avoid incrementing the lock version on ticks. Do you think it's worth using `update_all` directly over `update_counters` to try to bypass the locking for ticks?

adrianna-chang-shopify:
Okay let's revert. I think I'll need to modify it slightly to work with the changes around status reload frequency being configurable, but I'll take a look.

PR #1243, opened 2025-06-30
Mark tasks as being abstract to hide them from the list

sunny:
Hi there! Thanks for maintaining maintenance tasks ✨ 

This patch lets you inherit from tasks that only serve as parents without including them in the list. We can mark them as being abstract like so:

```rb
class ApplicationTask < MaintenanceTasks::Task
  self.abstract_class = true
end
```


sunny:
I have signed the CLA!

adrianna-chang-shopify:
There was some previous discussion on whether the gem should support abstract tasks (and a listed workaround) here: https://github.com/Shopify/maintenance_tasks/pull/788#issuecomment-1469797858

PR #1248, opened 2025-07-13
Drop support for no-longer-supported Rails 7.0 and clean up ✨

larouxn:
Proposing we drop support for Rails 7.0 (not 7.1 nor 7.2, _yet_) as it's no longer supported as of over 3 months ago. Doing so additionally allows us to clean up quite a few `if >= 7.1` type conditionals. ([Rails EOL info source](https://endoflife.date/rails))

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,6 @@ jobs:
       fail-fast: false
       matrix:
         gemfile:
-        - gemfiles/rails_7_0.gemfile
         - gemfiles/rails_7_1.gemfile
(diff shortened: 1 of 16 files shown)


etiennebarrie:
> Github is stalling on the checks for `rails_7_0_gemfile`, which was removed from the CI matrix, so I'm just gonna ignore that and merge 😅

FYI the fix is to go into the [repository "Branches" settings](https://github.com/Shopify/maintenance_tasks/settings/branches) then edit the branch protection rules for main, then scroll down, remove the statuses we don't want, and not forget to save the form. I've done it, just sharing for next time.

PR #1251, opened 2025-07-18
Feature: Add optional output to maintenance task to display output

scotttam:


app/models/maintenance_tasks/run.rb
```diff
@@ -443,6 +443,20 @@ def masked_arguments
...
     end
 
+    # Appends output to the existing output in the database if the column exists.
+    #
+    # @param new_output [String] the output to append.
+    def append_output(new_output)
+      return unless self.class.column_names.include?("output")
+
+      # Reload output from database to get the latest value
+      current_output = self.class.where(id: id).pluck(:output).first || ""
```
nvasilevski, line 453: Unfortunately I don't think we can afford adding a feature like that to the framework. It may work okay for smaller applications but for large applications with a lot of rows to iterate over the approach of transferring large chunks for data back-and-forth doesn't scale well

nvasilevski (commented):
I'm going to agree with Adrianna, in our experience this solution will not sustain in tasks that iterate over millions of records. And looking at the examples, from my personal experience I wouldn't even recommend issuing logs such as `Marking user for cleanup` or `Successfully processed`, those things should be implied by the "green path" of the task, meaning that if no error happened and the iteration counter went up it should be safe to assume that the task did what it was supposed to do. Otherwise a log filled with entries like "business as usual" becomes extremely hard to utilize for debugging purposes or just gets truncated depending on the infra setup

scotttam:
Thanks everyone for the feedback! 

`Rails.logger` isn't a great IMO as it requires running the task and then watching the logs, which in production is kinda of a pain and the format is terrible. Ideally I'd like to see the output immediately from the run.

I understand your concern about larger applications with multiple rows touching the database. Another idea I had was rather than writing to the database was writing the output to `Rails.cache`. That way, the output could be short lived (TTL 1 hour?), and we wouldn't have to have a migration or touch the database. How would folks feel about that?

paulodiovani:
This is actually something I was looking for. A way to check the output from a task without the need to dig through the logs.

I don't think that calling `update` on each step is ideal, though.

Some ideas (and/or):

- Collect the logs in memory (or cache) during processing by default and make them available in the task instance
- Write logs somewhere using the `after_complete` callback by default. The destination could be a custom logger interface.
- Include an optional method to get the logs to show in the ui (e.g. `Task#collect_logs`)

I believe all of these can be done with callbacks or customizing parent controller or job, but it would be nice to support the feature by default.

example:

```ruby
module Maintenance
  class ImportPostsTask < MaintenanceTasks::Task
    # set a custom logger interface
    with_logger MyCustomLogger

    def collection
      # ...
    end

    def process
      # do something ...
      write_log_output "then log something", append: true
    end

    def log_output
       read_log_output # default implementation to read logs from interface
    end
  end
end
```

PR #1276, opened 2025-09-09
Fix progress display for completed tasks with zero tick_total

iberianpig:
### Problem

When a maintenance task completes with an empty collection (`tick_total = 0`), the progress bar shows infinite loading instead of completion. This
happens because the `<progress>` HTML element without a `value` attribute continuously animates, even when the task has finished successfully.

This is especially problematic for tasks with mutable collections. For example, when running a task multiple times where the first execution processes
all matching articles, subsequent runs will have `tick_total = 0` but the progress bar still shows infinite loading.

**Example scenario:**
```ruby
module Maintenance
  class UpdateArticleStatusTask < MaintenanceTasks::Task
    def collection
      Article.where(status: 'draft', published_at: nil)
    end

    def process(article)
      ActiveRecord::Base.transaction do
        article.update!(status: 'archived')
        some_other_operation(article)
      end
    end

    def count
      collection.count
    end
  end
end
```

When this task runs again after all matching articles have been processed, tick_total becomes 0, but the progress bar shows infinite loading instead of
completion.

### Solution

Modified MaintenanceTasks::Progress#value to return 0 when:
- The run is completed (succeeded, errored, or cancelled)
- The tick_total is 0

This makes the HTML `<progress>` element receive value="0", showing a completed state instead of the infinite animation.

HTML output comparison:
#### Before: `<progress class="progress mt-4 is-primary is-light"></progress>` (infinite progress animation)

Screenshot from 2025-09-10 00-00-28

#### After: `<progress value="0" class="progress mt-4 is-success"></progress>` (with value="0", shows completed state)
image

Also updated Progress#text to handle cases where count is nil and task is completed, showing "Processed 0 items." for better user feedback.

### DIFF
--- a/app/models/maintenance_tasks/progress.rb
+++ b/app/models/maintenance_tasks/progress.rb
@@ -29,6 +29,7 @@ module MaintenanceTasks
     def value
       return 0 if @run.completed? && @run.tick_total.to_i.zero?
       return @run.tick_count if estimatable? || @run.stopped?
+
       nil
     end
 


app/models/maintenance_tasks/progress.rb
```diff
@@ -51,7 +53,9 @@ def text
       count = @run.tick_count
```
adrianna-chang-shopify, line 53: Perhaps we could do this instead?
```suggestion
      count = @run.tick_count || 0
```

And leave the if statement intact?

adrianna-chang-shopify (commented):
Thanks for contributing! Can you sign the CLA?

iberianpig:
To reproduce the issue, please use the provided `issue_1726.rb` file in the repository root:

## Reproduction Steps

```bash
ruby issue_1726.rb
```

1. Once the Rails app starts, navigate to http://localhost:3000/maintenance_tasks
2. **First execution**: Run the `MigratePostsToOldPosts` task
   - This processes 1 old post (created in 2023)
   - Task completes successfully with proper progress bar display
3. **Second execution**: Run the same task again immediately
   - Now there are 0 posts to process (already deleted in first run)
   - Task completes with `tick_total = 0`, `tick_count = 0`
   - **Issue occurs**: Progress bar shows infinite loading animation

## Expected Behavior (with fix)
Second execution should show: `<progress value="0" class="progress mt-4 is-success"></progress>`

## Actual Behavior (without fix)  
Second execution shows: `<progress class="progress mt-4 is-primary is-light"></progress>` (no value attribute = infinite loading)

This demonstrates the real-world scenario where tasks with mutable collections complete with empty results on subsequent runs. The issue is specific to completed tasks with `tick_total = 0`, which is different from testing with arbitrary empty IDs in ParamsTask.

etiennebarrie:
Your `tick_count` is:
```ruby
    t.bigint "tick_count"
```

When it should be
https://github.com/Shopify/maintenance_tasks/blob/76a40cebd5ea6839e1d238d3dce0971daa1c352b/test/dummy/db/schema.rb#L47

The issue is that the migration doesn't create the column like that:
https://github.com/Shopify/maintenance_tasks/blob/76a40cebd5ea6839e1d238d3dce0971daa1c352b/db/migrate/20220706101937_change_runs_tick_columns_to_bigints.rb#L6

To repro on the repo easily:
```console
$ rm test/dummy/db/schema.rb 
$ bin/rails db:drop db:create db:migrate
```

Then run ParamsTasks with `post_ids` set to `42`.

---

The issue does not impact MySQL, because the migration runs like this:
```
Migrating to ChangeRunsTickColumnsToBigints (20220706101937)
== 20220706101937 ChangeRunsTickColumnsToBigints: migrating ===================
-- change_table(:maintenance_tasks_runs, {bulk: true})
   (10.2ms)  ALTER TABLE `maintenance_tasks_runs` CHANGE `tick_count` `tick_count` bigint DEFAULT 0 NOT NULL, CHANGE `tick_total` `tick_total` bigint DEFAULT NULL
   -> 0.0229s
== 20220706101937 ChangeRunsTickColumnsToBigints: migrated (0.0230s) ==========
```

Whereas on SQLite we get:
```
Migrating to ChangeRunsTickColumnsToBigints (20220706101937)
== 20220706101937 ChangeRunsTickColumnsToBigints: migrating ===================
-- change_table(:maintenance_tasks_runs, {bulk: true})
  TRANSACTION (0.0ms)  BEGIN immediate TRANSACTION
  SQL (0.2ms)  PRAGMA foreign_keys
  SQL (0.0ms)  PRAGMA defer_foreign_keys
   (0.0ms)  PRAGMA defer_foreign_keys = ON
   (0.0ms)  PRAGMA foreign_keys = OFF
   (0.3ms)  CREATE TEMPORARY TABLE "amaintenance_tasks_runs" ("id" integer PRIMARY KEY AUTOINCREMENT NOT NULL, "task_name" varchar NOT NULL, "started_at" datetime(6), "ended_at" datetime(6), "time_running" float DEFAULT 0.0 NOT NULL, "tick_count" integer DEFAULT 0 NOT NULL, "tick_total" integer, "job_id" varchar, "cursor" varchar, "status" varchar DEFAULT 'enqueued' NOT NULL, "error_class" varchar, "error_message" varchar, "backtrace" text, "created_at" datetime(6) NOT NULL, "updated_at" datetime(6) NOT NULL, "arguments" text, "lock_version" integer DEFAULT 0 NOT NULL)
   (0.0ms)  CREATE INDEX "tindex_amaintenance_tasks_runs_on_task_name_and_created_at" ON "amaintenance_tasks_runs" ("task_name", "created_at" DESC)
  SQL (0.0ms)  INSERT INTO "amaintenance_tasks_runs" ("id","task_name","started_at","ended_at","time_running","tick_count","tick_total","job_id","cursor","status","error_class","error_message","backtrace","created_at","updated_at","arguments","lock_version")
                     SELECT "id","task_name","started_at","ended_at","time_running","tick_count","tick_total","job_id","cursor","status","error_class","error_message","backtrace","created_at","updated_at","arguments","lock_version" FROM "maintenance_tasks_runs"
   (0.1ms)  DROP TABLE "maintenance_tasks_runs"
   (0.0ms)  CREATE TABLE "maintenance_tasks_runs" ("id" integer PRIMARY KEY AUTOINCREMENT NOT NULL, "task_name" varchar NOT NULL, "started_at" datetime(6), "ended_at" datetime(6), "time_running" float DEFAULT 0.0 NOT NULL, "tick_count" bigint, "tick_total" integer, "job_id" varchar, "cursor" varchar, "status" varchar DEFAULT 'enqueued' NOT NULL, "error_class" varchar, "error_message" varchar, "backtrace" text, "created_at" datetime(6) NOT NULL, "updated_at" datetime(6) NOT NULL, "arguments" text, "lock_version" integer DEFAULT 0 NOT NULL)
   (0.0ms)  CREATE INDEX "index_maintenance_tasks_runs_on_task_name_and_created_at" ON "maintenance_tasks_runs" ("task_name", "created_at" DESC)
  SQL (0.0ms)  INSERT INTO "maintenance_tasks_runs" ("id","task_name","started_at","ended_at","time_running","tick_count","tick_total","job_id","cursor","status","error_class","error_message","backtrace","created_at","updated_at","arguments","lock_version")
                     SELECT "id","task_name","started_at","ended_at","time_running","tick_count","tick_total","job_id","cursor","status","error_class","error_message","backtrace","created_at","updated_at","arguments","lock_version" FROM "amaintenance_tasks_runs"
   (0.0ms)  DROP TABLE "amaintenance_tasks_runs"
   (0.0ms)  PRAGMA defer_foreign_keys = 0
   (0.0ms)  PRAGMA foreign_keys = 1
  SQL (0.0ms)  PRAGMA foreign_keys
  SQL (0.0ms)  PRAGMA defer_foreign_keys
   (0.0ms)  PRAGMA defer_foreign_keys = ON
   (0.0ms)  PRAGMA foreign_keys = OFF
   (0.0ms)  CREATE TEMPORARY TABLE "amaintenance_tasks_runs" ("id" integer PRIMARY KEY AUTOINCREMENT NOT NULL, "task_name" varchar NOT NULL, "started_at" datetime(6), "ended_at" datetime(6), "time_running" float DEFAULT 0.0 NOT NULL, "tick_count" bigint, "tick_total" integer, "job_id" varchar, "cursor" varchar, "status" varchar DEFAULT 'enqueued' NOT NULL, "error_class" varchar, "error_message" varchar, "backtrace" text, "created_at" datetime(6) NOT NULL, "updated_at" datetime(6) NOT NULL, "arguments" text, "lock_version" integer DEFAULT 0 NOT NULL)
   (0.0ms)  CREATE INDEX "tindex_amaintenance_tasks_runs_on_task_name_and_created_at" ON "amaintenance_tasks_runs" ("task_name", "created_at" DESC)
  SQL (0.0ms)  INSERT INTO "amaintenance_tasks_runs" ("id","task_name","started_at","ended_at","time_running","tick_count","tick_total","job_id","cursor","status","error_class","error_message","backtrace","created_at","updated_at","arguments","lock_version")
                     SELECT "id","task_name","started_at","ended_at","time_running","tick_count","tick_total","job_id","cursor","status","error_class","error_message","backtrace","created_at","updated_at","arguments","lock_version" FROM "maintenance_tasks_runs"
   (0.0ms)  DROP TABLE "maintenance_tasks_runs"
   (0.0ms)  CREATE TABLE "maintenance_tasks_runs" ("id" integer PRIMARY KEY AUTOINCREMENT NOT NULL, "task_name" varchar NOT NULL, "started_at" datetime(6), "ended_at" datetime(6), "time_running" float DEFAULT 0.0 NOT NULL, "tick_count" bigint, "tick_total" bigint, "job_id" varchar, "cursor" varchar, "status" varchar DEFAULT 'enqueued' NOT NULL, "error_class" varchar, "error_message" varchar, "backtrace" text, "created_at" datetime(6) NOT NULL, "updated_at" datetime(6) NOT NULL, "arguments" text, "lock_version" integer DEFAULT 0 NOT NULL)
   (0.0ms)  CREATE INDEX "index_maintenance_tasks_runs_on_task_name_and_created_at" ON "maintenance_tasks_runs" ("task_name", "created_at" DESC)
  SQL (0.0ms)  INSERT INTO "maintenance_tasks_runs" ("id","task_name","started_at","ended_at","time_running","tick_count","tick_total","job_id","cursor","status","error_class","error_message","backtrace","created_at","updated_at","arguments","lock_version")
                     SELECT "id","task_name","started_at","ended_at","time_running","tick_count","tick_total","job_id","cursor","status","error_class","error_message","backtrace","created_at","updated_at","arguments","lock_version" FROM "amaintenance_tasks_runs"
   (0.0ms)  DROP TABLE "amaintenance_tasks_runs"
   (0.0ms)  PRAGMA defer_foreign_keys = 0
   (0.0ms)  PRAGMA foreign_keys = 1
   -> 0.0248s
```

It seems the MySQL adapter preserves the `NOT NULL` and `DEFAULT` whereas the SQLite adapter doesn't.

---

You can fix this in your application by adding a migration that sets the default for the column to 0.
We'll probably need to add a new migration to fix this for everyone out there using SQLite, and we need to check PostgreSQL too.
Maybe we'll want to add some code to the gem to explicitly default to 0 too, just to avoid the glitch.

Thanks again for your contribution and the detailed repro script!

iberianpig:
@etiennebarrie Thank you for the very detailed analysis!

I looked more into this and found it's a SQLite migration issue that began in Rails 7.x.
Currently, migration with SQLite drops NOT NULL and DEFAULT constraints that aren't clearly specified, so we must pass all column options during migrations.

For anyone with this problem:

Here's a workaround to fix the constraints:

```ruby
# db/migrate/[timestamp]_restore_tick_count_constraints.rb
class RestoreTickCountConstraints < ActiveRecord::Migration[7.0]
  def up
    if ActiveRecord::Base.connection.adapter_name == 'SQLite'
      change_column_default :maintenance_tasks_runs, :tick_count, from: nil, to: 0
      change_column_null :maintenance_tasks_runs, :tick_count, false, 0
    end
  end

  def down
    if ActiveRecord::Base.connection.adapter_name == 'SQLite'
      change_column_null :maintenance_tasks_runs, :tick_count, true
      change_column_default :maintenance_tasks_runs, :tick_count, from: 0, to: nil
    end
  end
end
```
This is actually a bug in Rails itself, so it should be fixed in Rails.

PR #1286, opened 2025-09-23
Add a nonce to inline script

brandoncc:
Adding a nonce so that the code is compatible with strict Content Security Policies.


fresh-eggs:
@brandoncc had another look at the report-only violation we discussed, Étienne is right in that the enforcing policy applied by the maintenance-tasks gem already [accounts for those inline scripts with a hash](https://content-security-policy.com/hash/).

You could either safely ignore report-only violations from controllers that belong to the engine or add those hashes to the report-only policy to stop reporting them.

PR #1287, opened 2025-09-23
(Not) handled exceptions

eval:
This PR:
- ...fixes an inconsistency in the `source` that is reported for rescued exceptions (ie via report_on)
- ...ensures exceptions are by default considered not handled. Whereas 'report_on exceptions' are considered handled.

This makes it possible to distinguish between rescued exceptions and other exceptions in an error-subscriber:
```ruby
class MaintenanceTasksErrorSubscriber
  def report(error, handled:, severity:, context:, source: nil)
    return unless source == "maintenance-tasks"

    ExceptionService.capture(ex) unless handled
    ...e.g. logging based on severity
  end
end
```

### DIFF
--- a/app/models/maintenance_tasks/task.rb
+++ b/app/models/maintenance_tasks/task.rb
@@ -235,7 +235,7 @@ module MaintenanceTasks
       #
       # @param exceptions list of exceptions to rescue and report
       # @param report_options [Hash] optionally, supply additional options for `Rails.error.report`.
-      #   By default: <code>{ source: "maintenance-tasks" }</code>.
+      #   By default: <code>{ handled: true, source: "maintenance-tasks" }</code>.
       def report_on(*exceptions, **report_options)
         rescue_from(*exceptions) do |exception|
           Rails.error.report(exception, source: "maintenance-tasks", **report_options)
--- a/README.md
+++ b/README.md
@@ -1027,7 +1027,7 @@ be placed in a `maintenance_tasks.rb` initializer.
 Exceptions raised while a Task is performing are rescued and information about
 the error is persisted and visible in the UI.
 
-Errors are also sent to the `Rails.error.reporter`, which can be configured by
+Errors are also sent to `Rails.error.report`, which can be configured by
 your application. See the [Error Reporting in Rails
 Applications][rails-error-reporting] guide for more details.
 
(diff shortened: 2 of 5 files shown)


adrianna-chang-shopify (commented):
Hey @eval , thank you for your contribution!

Most of the changes in this PR look good to me, the only one I'm on the fence about is whether we should treat calls to the Error Reporter from within `#on_error` as `handled: false`. I acknowledge that this is actually the current behaviour, so explicitly adding `handled: false` actually doesn't do anything... And I understand the motivation to distinguish "expected" errors (which we configure via `report_on`) from "unexpected" errors (which get caught by the general `on_error` mechanism). The only thing I'm getting stuck on is that "unhandled" generally means that the error wasn't handled gracefully by the application code and potentially caused some sort of crash, but here we _are_ technically handling the error within the `maintenance_tasks` infra and not re-raising it. It does make sense, on the subscriber side, to want to only "do something" if we're hitting an error that was not "expected" -- but "handled" vs "expected" aren't quite the same thing.

Could I get a second opinion on this? cc @gmcgibbon or @etiennebarrie do you have thoughts on this?

app/models/maintenance_tasks/task.rb
```diff
@@ -235,10 +235,10 @@ def after_error(*filter_list, &block)
       #
       # @param exceptions list of exceptions to rescue and report
       # @param report_options [Hash] optionally, supply additional options for `Rails.error.report`.
-      #   By default: <code>{ source: "maintenance_tasks" }</code> or (Rails <v7.1) <code>{ handled: true }</code>.
+      #   By default: <code>{ handled: true, source: "maintenance-tasks" }</code>.
       def report_on(*exceptions, **report_options)
         rescue_from(*exceptions) do |exception|
-          Rails.error.report(exception, source: "maintenance_tasks", **report_options)
+          Rails.error.report(exception, handled: true, source: "maintenance-tasks", **report_options)
```
etiennebarrie: This is the default, let's not pass it, only test it.
```suggestion
          Rails.error.report(exception, source: "maintenance-tasks", **report_options)
```

etiennebarrie (commented):
Sorry I had a pending review comment that didn't go in yesterday.

PR #1440, opened 2026-04-06
Handle Task::NotFoundError with a 404 page instead of raising an error

kaibadash:
deleted task

- Add `rescue_from Task::NotFoundError` in `ApplicationController` to render a friendly 404 page when users access a task that no longer exists
- Create a dedicated not found view (`not_found.html.erb`) styled with Bulma to match the existing UI, with a link back to the tasks list
- Remove the `rescue_responses` mapping from `Engine` since the error is now handled directly by the controller

## Motivation

Maintenance tasks are short-lived by nature — they are created for one-time data migrations or cleanup operations and are expected to be deleted once they are no longer needed. However, users may still have bookmarked URLs or links in Slack/docs pointing to deleted tasks.

Previously, accessing a deleted task with no associated runs raised `MaintenanceTasks::Task::NotFoundError`, which:

1. Displayed an unfriendly error page to users
2. Was reported to error monitoring systems (e.g., Sentry) as an unhandled exception, creating noise

By using `rescue_from` in the controller, the exception is caught before it propagates to error monitoring middleware, and users see a clean 404 page with navigation back to the task list.

### DIFF
--- a/app/controllers/maintenance_tasks/application_controller.rb
+++ b/app/controllers/maintenance_tasks/application_controller.rb
@@ -31,5 +31,9 @@ module MaintenanceTasks
     end
 
     protect_from_forgery with: :exception
+
+    rescue_from Task::NotFoundError do |error|
+      render("maintenance_tasks/not_found", status: :not_found)
+    end
   end
 end
--- a/app/controllers/maintenance_tasks/application_controller.rb
+++ b/app/controllers/maintenance_tasks/application_controller.rb
@@ -31,5 +31,9 @@ module MaintenanceTasks
     end
 
     protect_from_forgery with: :exception
+
+    rescue_from Task::NotFoundError do
+      render("maintenance_tasks/not_found", status: :not_found)
+    end
   end
 end
--- a/app/views/maintenance_tasks/not_found.html.erb
+++ b/app/views/maintenance_tasks/not_found.html.erb
@@ -1,0 +1,9 @@
+<% content_for :page_title, "Not Found" %>
+
+<div class="container has-text-centered">
+  <h1 class="title is-3 has-text-weight-bold">Task Not Found</h1>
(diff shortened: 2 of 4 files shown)


etiennebarrie (commented):
I think this could be nice.

At first I thought that shouldn't be necessary (easy enough to tweak the URL), and the Sentry issue is a problem with that gem or the configuration (although it shouldn't be necessary to add this exception class to list of excluded exceptions, doing so would solve that issue).

But the change is small, the page looks nicer (and also the main app's 404 page might be super custom) and moving away from `action_dispatch.rescue_responses` means better support for exception monitoring tools which don't respect [`action_dispatch.report_exception`](https://github.com/rails/rails/blob/v8.1.3/actionpack/lib/action_dispatch/middleware/show_exceptions.rb#L38).

kaibadash:
Thank you for your review! I'm sorry for the late reply. I'll fix that asap 👍

kaibadash:
I'm sorry I haven't been able to work on this because I've been so busy. I'll get to it as soon as possible.

etiennebarrie:
No worries. Don't feel rushed.

app/controllers/maintenance_tasks/application_controller.rb
```diff
@@ -31,5 +31,9 @@ class ApplicationController < MaintenanceTasks.parent_controller.constantize
     end
 
     protect_from_forgery with: :exception
+
+    rescue_from Task::NotFoundError do |error|
```
etiennebarrie, line 35:
```suggestion
    rescue_from Task::NotFoundError do
```

This was causing a lint error, I removed the block parameter.

etiennebarrie (approved):

etiennebarrie:
Thank you!

kaibadash:
I'm sorry I couldn't help you. Thank you.

etiennebarrie:
Don't be sorry, you did great!

PR #1463, opened 2026-05-01
Bound the count query with a server-side statement timeout

selesse:
### Why

`Task#count` populates the progress bar before iteration begins and is called from [`TaskJobConcern#on_start`](https://github.com/Shopify/maintenance_tasks/blob/cc20956e71aa3798b0c2ea116cbdb29018b80e72/app/jobs/concerns/maintenance_tasks/task_job_concern.rb#L147-L151). On large or poorly-indexed collections this query can run for tens of seconds and stall the task — even though the count is only a UI hint and a failure here should not fail the run.

We've seen this hit production tasks repeatedly on my team, with count queries running long enough to time out the entire run.

### Why not `Timeout.timeout`?

Active Record has no cross-platform query timeout API, so the obvious fix is `Timeout.timeout(5) { @task.count }`. That's unsafe to use around database calls — it raises across thread boundaries at arbitrary bytecode points, which can leave the connection mid-protocol with the server-side query still running on the database. That's the path #1388 currently uses for its count-preview endpoint, and it's the path I'd like to avoid for `on_start`.

### What this does instead

Wrap the count call in a small adapter-aware helper that issues the database's *native* cancellation:

| Adapter | Mechanism |
|---|---|
| PostgreSQL / PostGIS | `SET LOCAL statement_timeout` inside `transaction(requires_new: true)`, auto-reset on commit/rollback |
| MySQL / Trilogy | `SET SESSION max_execution_time`, prior value restored in `ensure` |
| SQLite / unknown | no-op (the helper yields unmodified) |

When the database cancels the query it raises `ActiveRecord::QueryCanceled`. `safe_count` rescues it; the run starts with a `nil` count (the progress bar shows progress without a percentage) and the cancellation is reported to `Rails.error` as a *handled* warning so it remains visible without paging anyone.

Default is 5 seconds, configurable via `MaintenanceTasks.count_timeout_ms` (`nil` or `0` to disable).

### Tests

Unit tests use mocked connections to assert the right SQL is issued for each adapter, that prior values are restored, that errors propagate, and that `nil`/`0`/negative timeouts short-circuit before touching the connection.

Integration tests against a real PostgreSQL connection (the existing [`gemfiles/postgresql.gemfile`](https://github.com/Shopify/maintenance_tasks/blob/cc20956e71aa3798b0c2ea116cbdb29018b80e72/.github/workflows/ci.yml#L23-L27) CI matrix) verify that a slow query is actually cancelled, that fast queries pass through, and that `statement_timeout` is reset after both successful and failed blocks. They skip on SQLite.

### Questions for reviewers

1. **Default value.** I picked 5s as a "behave better by default" floor — long enough that almost every reasonable count finishes, short enough to prevent stalls. Open to making it `nil` (opt-in) if you'd prefer to avoid behaviour changes for existing users on first upgrade.
2. **Multi-DB.** The timeout is applied to `ActiveRecord::Base.connection`. If a task's `#count` queries a model on a different connection, the timeout won't apply there. Documented as a caveat; could be extended to introspect the relation's connection if useful.

### DIFF
--- a/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
+++ b/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
@@ -145,9 +145,21 @@ module MaintenanceTasks
     end
 
     def on_start
-      count = @task.count
+      @run.start(safe_count)
+    end
+
+    def safe_count
+      count = CountTimeout.with_timeout(MaintenanceTasks.count_timeout_ms) { @task.count }
       count = @collection_enum.size if count == NO_COUNT_DEFINED
-      @run.start(count)
+      count
+    rescue ActiveRecord::QueryCanceled => error
+      Rails.error.report(
+        error,
+        handled: true,
+        severity: :warning,
+        context: { task: @task.class.name, timeout_ms: MaintenanceTasks.count_timeout_ms },
+      )
+      nil
     end
 
     def on_shutdown
--- a/lib/maintenance_tasks.rb
+++ b/lib/maintenance_tasks.rb
@@ -7,6 +7,7 @@ require "active_record"
 
 require "job-iteration"
 require "maintenance_tasks/engine"
+require "maintenance_tasks/count_timeout"
 
 # The engine's namespace module. It provides isolation between the host
 # application's code and the engine-specific code. Top-level engine constants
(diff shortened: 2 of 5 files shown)


ryanquanz (approved):

etiennebarrie:
> The `count == :no_count` → `@collection_enum.size` fallback is preserved as-is. That fallback is the bigger footgun — operators returning `:no_count` to opt out of expensive counts get the count run for them anyway — but changing it is a separate behaviour change that needs a deprecation cycle, and I wanted to keep this PR scoped to the timeout safety net so it can land on its own.

Can you explain the footgun exactly? If I change `Maintenance::UpdatePostsTask` to return `nil` as described in the README, there's no count emitted. If you do return `:no_count`, the count is calculated but that's user error, we could change `:no_count` to be something else, it's not documented or anything. It doesn't mean "don't count" it means "there's no count method defined in the task".

selesse:
> > The count == :no_count → @collection_enum.size fallback is preserved as-is. That fallback is the bigger footgun — operators returning :no_count to opt out of expensive counts get the count run for them anyway — but changing it is a separate behaviour change that needs a deprecation cycle, and I wanted to keep this PR scoped to the timeout safety net so it can land on its own.
>
> Can you explain the footgun exactly? If I change Maintenance::UpdatePostsTask to return nil as described in the README, there's no count emitted. If you do return :no_count, the count is calculated but that's user error, we could change :no_count to be something else, it's not documented or anything. It doesn't mean "don't count" it means "there's no count method defined in the task".

Fair. The footgun documented is based on my team's experience in the past few weeks. AI tools are confidently recommending `:no_count` instead of nil as a way to skip counting.

etiennebarrie:
I don't think we should be implementing query timeouts for multiple database adapters in this gem, that should be in Active Record. Then once it is, we could use it here.

selesse:
> I don't think we should be implementing query timeouts for multiple database adapters in this gem, that should be in Active Record. Then once it is, we could use it here.

I had a similar thought, but given that timeouts aren't supported uniformly across different databases, I think that's a much bigger lift.

catwomey:
General thought after using this gem for the last few years.  For active record collections it might be more intuitive for the count to be an opt in rather than an opt out.  Or have tighter safety around the count so that it doesn't fail a run.

I've run many maintenance tasks that are performant enough with active record collections  in 100 record batches, but completely crash when you try to count the entire collection.

etiennebarrie:
Right, that would be good to at least not error out the task completely, of course. But the task template does generate an empty `count` method, so counting is opt-in in that sense. (Note that the template can be overridden from the application too, in `lib/generators/maintenance_tasks/templates/task.rb.tt`)

But I still maintain that we shouldn't introduce concepts here that are better suited to the framework or the application.

PR #1475, opened 2026-05-14
Drop support for no-longer-supported Rails 7.1 and clean up ✨

larouxn:
Inspired / based on https://github.com/Shopify/maintenance_tasks/pull/1248

Proposing we drop support for Rails 7.1 (not 7.2, _yet_) as it's no longer supported as of over 3 months ago. Doing so additionally allows us to clean up quite a few `if >= 7.1` type conditionals. ([Rails EOL info source](https://endoflife.date/rails))

Screenshot From 2026-05-14 21-34-05

Regarding the likely `rails_7_1_gemfile` CI failure we can follow the guidance here: https://github.com/Shopify/maintenance_tasks/pull/1248#issuecomment-3127483925.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,6 @@ jobs:
       fail-fast: false
       matrix:
         gemfile:
-        - gemfiles/rails_7_1.gemfile
         - gemfiles/rails_7_2.gemfile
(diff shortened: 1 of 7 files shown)


adrianna-chang-shopify (approved):
Thanks @larouxn ! 🙇‍♀️

PR #1485, opened 2026-05-26
Add tag macro for grouping tasks in the web UI

ahmdrefat:
## Summary

Introduces a `tag` class macro on `MaintenanceTasks::Task` that records one or more symbol labels on a Task class. Tags are inherited from the superclass and additive across `tag` calls — calling `tag :foo, :bar` on a subclass extends (does not replace) the parent's tags.

```ruby
class Maintenance::CleanupOldPostsTask < MaintenanceTasks::Task
  tag :data, :cleanup
  # ...
end
```

`TaskDataIndex#tags` exposes the value to the index view, returning an empty frozen array when the underlying class has been deleted so the deleted-task render path stays safe.

This PR is **data-layer only**. The tag-filter chips on the Tasks index are part of the follow-up UI PR (so this one can be reviewed and shipped in isolation).

## Stack

This is **PR 1 of 4** in a stack that progressively lands a refreshed Tasks UI:

1. **(this PR)** Tagging API
2. UI overhaul (search, tag filter chips, dark theme, navbar refresh indicator)
3. Auto-refresh: morph DOM in place instead of replacing subtrees
4. Run-card spacing & style refinement

The follow-up PRs are open as drafts; their diffs will shrink as the earlier ones land.

## Test plan

- [x] `bundle exec rubocop` clean on touched files
- [x] New `task_test.rb` coverage: default empty-and-frozen tags, additive `tag`, string coercion, dedup, sub/superclass isolation

### DIFF
--- a/app/models/maintenance_tasks/task.rb
+++ b/app/models/maintenance_tasks/task.rb
@@ -39,6 +39,11 @@ module MaintenanceTasks
     # @api private
     class_attribute :status_reload_frequency, default: MaintenanceTasks.status_reload_frequency
 
+    # Tags applied to this Task, used to group and filter Tasks in the
+    # web UI. Subclasses inherit parent tags and may add their own via
+    # the `tag` class macro.
+    class_attribute :tags, default: [].freeze
+
     define_callbacks :start, :complete, :error, :cancel, :pause, :interrupt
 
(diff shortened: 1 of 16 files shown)


ahmdrefat:

PR #1486, opened 2026-05-26
Overhaul the web UI: dark theme, search, tag filters, keyboard shortcuts

ahmdrefat:
## Summary

Refreshes the Tasks index and Run views with quality-of-life improvements that scale to admin UIs with hundreds of tasks:

- **Light/dark theme toggle** that respects `prefers-color-scheme`, persists in `localStorage`, and is painted synchronously on load to avoid FOUC.
- **Client-side search & tag filters** on the Tasks index. `/` focuses the search, `Escape` clears search and unpresses tag chips, `t` toggles theme.
- **Reworked task table** with separate "with previous runs" and "Never run" sections, accessible status pill, monospaced name with hover-friendly hit area, and a row pulse when status changes during a refresh.
- **Navbar refresh indicator** that visualizes when the page is polling.
- **Run cards** wrapped in `.run` so the run-id permalink becomes a sticky absolute anchor with a 2.75rem touch target; custom-content `id` scoped per-run to avoid duplicate ids.

Inline CSS/JS stay inline — CSP hashes in `ApplicationController` are updated to match.

No public API changes beyond what landed in PR 1 (the `tag` macro). All existing tests pass.

## Stack

**PR 2 of 4.** Depends on #1485.

Until #1485 merges, this PR's diff includes the tagging commit. Once it merges, this diff shrinks to just the UI changes.

Follow-ups (already drafted):
- PR 3: Auto-refresh — morph DOM in place
- PR 4: Run-card spacing & style refinement

## Test plan

- [x] `bundle exec rubocop` clean on all touched Ruby files
- [x] Manual dogfood in browser:
  - light + dark theme renders cleanly, persists across reloads
  - `/` focus / `Escape` clear / `t` toggle keyboard shortcuts
  - search narrows rows by name; tag chip narrows by tag; both compose
  - auto-refresh on `[data-refresh]` containers still updates without breaking filters

### DIFF
--- a/app/controllers/maintenance_tasks/application_controller.rb
+++ b/app/controllers/maintenance_tasks/application_controller.rb
@@ -11,7 +11,7 @@ module MaintenanceTasks
       policy.style_src_elem(
         BULMA_CDN,
         # <style> tag in app/views/layouts/maintenance_tasks/application.html.erb
-        "'sha256-b9tTK1UaF0U8792/A1vIUkeZwjPgECIOeKJdhYED06A='",
+        "'sha256-2+QYdn+KxogttA55Mat4WTVkM02itlZXLQ5aolcv3qk='",
       )
       capybara_lockstep_scripts = [
         "'sha256-1AoN3ZtJC5OvqkMgrYvhZjp4kI8QjJjO7TAyKYiDw+U='",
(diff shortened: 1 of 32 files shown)


ahmdrefat:

PR #1487, opened 2026-05-26
Auto-refresh: morph the DOM in place instead of replacing subtrees

ahmdrefat:
## Summary

Before, every 3-second poll on a page with `[data-refresh]` wholesale replaced the matching subtree, then patched things back up: re-applying `<details open>` state, re-focusing the active element, restoring `window.scrollY`, and adding a pulse class to rows whose status had transitioned. With nothing changed server-side, this still produced a complete subtree replacement every tick.

This PR adds a small `morph(oldNode, newNode)` walker (~70 lines inline JS):

- **Fast-path on `isEqualNode`** — identical subtrees produce zero DOM mutations. Verified with `MutationObserver`: a no-op poll records **0** mutation records (was: one large `childList` replacement covering the whole subtree).
- **Per-element attribute sync.** `<details open>` is explicitly skipped — server always renders open, but a user-closed detail must stay closed.
- **Children matched by id when present** (run cards have `id=\"run_N\"`), otherwise by position. Server ordering is deterministic so position matching is stable.

Because morph preserves node identity, the explicit snapshot/restore of `<details>` state, focus, and scrollY is no longer needed — it's gone. The row-pulse on status transitions stays and keys off the live (preserved) row.

### Constraint

Form input `value`/`checked` properties are not synced — no inputs currently live inside `[data-refresh]`. This is called out in an inline comment so the next person to add an input to a polled region sees it.

CSP hash for the inline `<script>` updated accordingly.

## Stack

**PR 3 of 4.** Depends on #1486.

Until #1485 and #1486 merge, this PR's diff includes their commits. Once they merge, this diff shrinks to just the morph rewrite + CSP hash bump.

## Test plan

- [x] `bundle exec rubocop` clean
- [x] In-browser via `MutationObserver`:
  - **No-op poll** (no server-side change) → 0 DOM mutations across 2 polling cycles
  - **Divergent state** (mutate row's `data-status` client-side) → next morph reverts only that attribute, **2** attribute mutations, row node identity preserved
  - Stale client-only attribute removed by morph (e.g. a marker `data-*`)
  - `<details>` that user has closed stays closed across polls

### DIFF
--- a/app/controllers/maintenance_tasks/application_controller.rb
+++ b/app/controllers/maintenance_tasks/application_controller.rb
@@ -19,7 +19,7 @@ module MaintenanceTasks
       ] if defined?(Capybara::Lockstep)
       policy.script_src_elem(
         # <script> tag in app/views/layouts/maintenance_tasks/application.html.erb
-        "'sha256-EgJg72VJBHkkzLEDQu1lwiwXRk+MTKJ2i9w3W3CWlBc='",
+        "'sha256-ZkcBB93dJLHCMozzRI0qwnn9xOL1apTbXgbZ1u33rvI='",
         # <script> tag for capybara-lockstep
         *capybara_lockstep_scripts,
       )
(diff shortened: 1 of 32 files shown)


ahmdrefat:

PR #1489, opened 2026-05-29
Add toggle to disable task list auto-refresh

### LINKED ISSUE #1490: Add the ability to disable auto refresh
# Add the ability to disable auto refresh

## Problem

The Maintenance Tasks web UI auto-refreshes the task list page every 3 seconds to keep run statuses up to date. This behavior is driven by JavaScript in the application layout: when an element with `data-refresh` is present and truthy, the page periodically fetches the current URL and replaces the refreshed content in place.

While this is useful when monitoring active runs, it can be disruptive in other situations:

- **Debugging or inspecting the page** — the DOM is replaced every few seconds, making it hard to read logs, copy values, or use browser dev tools.
- **Stable UI for screenshots or demos** — auto-refresh causes the page to flicker and the cursor to change to a wait state.
- **Reduced background activity** — unnecessary polling when no runs are in progress or when the operator does not need live updates.

Today there is no way for users to turn this off on the task list page. Auto-refresh is always enabled (`@refresh = true` in `TasksController#set_refresh`).

## Proposed solution

Add a user-facing toggle on the task list page to disable and re-enable auto-refresh.

### Behavior

| State | URL | `data-refresh` attribute | UI control |
|-------|-----|--------------------------|------------|
| Default (auto-refresh on) | `/maintenance_tasks/` | `true` | "Disable auto-refresh" link |
| Auto-refresh off | `/maintenance_tasks/?refresh=false` | absent / empty | "Enable auto-refresh" link |

- Auto-refresh remains **on by default** so existing behavior is unchanged.
- Disabling auto-refresh sets `?refresh=false` on the URL, which the controller reads to set `@refresh = false`.
- Re-enabling auto-refresh navigates back to the root path without the query parameter.
- The toggle is a small button aligned to the top-right of the task list.

### Implementation outline

1. **Controller** — Update `MaintenanceTasks::TasksController#set_refresh` to respect a `refresh` query parameter:

   ```ruby
   @refresh = params[:refresh] != "false"
   ```

2. **View** — Add toggle links above the task list in `app/views/maintenance_tasks/tasks/index.html.erb`, and continue to set `data-refresh` from `@refresh` so the existing JavaScript respects the preference.

3. **Tests** — Add integration tests covering:
   - Task list auto-refreshes by default (`[data-refresh=true]` present).
   - `refresh=false` disables auto-refresh (`[data-refresh=true]` absent).
   - Correct toggle link is shown in each state.

4. **Documentation** — Document the toggle and `?refresh=false` URL parameter in the README.

## Scope

This change applies to the **task list** (`TasksController#index`) only. Individual task show pages already conditionally set `data-refresh` based on whether there are active runs (`@task.refresh?`), and that behavior is unchanged.

## Acceptance criteria

- [ ] Task list auto-refreshes by default (no change to current default behavior).
- [ ] Users can disable auto-refresh via a visible control on the task list page.
- [ ] Disabling auto-refresh persists for the session via the `?refresh=false` URL parameter.
- [ ] Users can re-enable auto-refresh from the same page.
- [ ] Integration tests cover both refresh states and the toggle links.
- [ ] README documents the feature.

knightq:
Closes https://github.com/Shopify/maintenance_tasks/issues/1490

## Problem

The Maintenance Tasks web UI auto-refreshes pages every 3 seconds to keep run statuses up to date. This behavior is driven by JavaScript in the application layout: when an element with `data-refresh` is present and truthy, the page periodically fetches the current URL and replaces the refreshed content in place.

While this is useful when monitoring active runs, it can be disruptive in other situations:

- **Debugging or inspecting the page** — the DOM is replaced every few seconds, making it hard to read logs, copy values, or use browser dev tools.
- **Stable UI for screenshots or demos** — auto-refresh causes the page to flicker and the cursor to change to a wait state.
- **Reduced background activity** — unnecessary polling when the operator does not need live updates.

Two pages are affected:

| Page | When it auto-refreshes | Prior behavior |
|------|------------------------|----------------|
| **Task list** (`TasksController#index`) | Always | `@refresh` was hard-coded to `true` |
| **Task show** (`TasksController#show`) | While the task has active runs (`@task.refresh?`) | No way to override `@task.refresh?` |

There was no user-facing way to turn auto-refresh off on either page.

## Proposed solution

Add a user-facing toggle on the **task show page** (when the task has active runs) to disable and re-enable auto-refresh. Respect a shared `?refresh=false` URL parameter in the controller so the preference applies consistently across both pages.

#### Task show page (primary UI)

The toggle is shown only when the task has active runs (`@task.refresh?`). When there are no active runs, the page does not auto-refresh and no toggle is displayed.

| State | URL | `data-refresh` attribute | UI control |
|-------|-----|--------------------------|------------|
| Default (auto-refresh on) | `/maintenance_tasks/tasks/:id` | `true` (when active runs exist) | "Disable auto-refresh" button |
| Auto-refresh off | `/maintenance_tasks/tasks/:id?refresh=false` | absent / empty | "Enable auto-refresh" button |

- Auto-refresh remains **on by default** when active runs are present.
- Disabling auto-refresh navigates to the same task with `?refresh=false`.
- Re-enabling auto-refresh navigates back to the task path without the query parameter.
- The toggle is a small button aligned to the top-right, above the runs section.

#### Task list page

The task list continues to auto-refresh by default and respects `?refresh=false` when present, but **does not include a toggle control**. Users who disable auto-refresh on a task show page can keep `refresh=false` in the URL when navigating back to the list, or append it manually.

| State | URL | `data-refresh` attribute | UI control |
|-------|-----|--------------------------|------------|
| Default (auto-refresh on) | `/maintenance_tasks/` | `true` | *(none)* |
| Auto-refresh off | `/maintenance_tasks/?refresh=false` | absent / empty | *(none)* |

### Implementation outline

1. **Controller** — Run `set_refresh` on both `index` and `show`, and read the `refresh` query parameter:

   ```ruby
   before_action :set_refresh, only: [:index, :show]

   def set_refresh
     @refresh = params[:refresh] != "false"
   end
   ```

2. **Task show view** — In `app/views/maintenance_tasks/tasks/show.html.erb`:
   - Render Disable/Enable toggle links when `@task.refresh?` is true.
   - Set `data-refresh` from `@task.refresh? && @refresh` so both active-run status and the user preference are respected.

3. **Task list view** — In `app/views/maintenance_tasks/tasks/index.html.erb`, continue setting `data-refresh` from `@refresh` (no toggle UI).

4. **Tests** — Integration tests in `test/integration/maintenance_tasks/tasks_controller_test.rb` covering:
   - Task show page auto-refreshes by default when there are active runs (`[data-refresh=true]` present).
   - `refresh=false` disables auto-refresh on the task show page (`[data-refresh=true]` absent).
   - Correct toggle link is shown in each state on the task show page.
   - No toggle links when the task has no active runs.

5. **Documentation** — Document the toggle and `?refresh=false` URL parameter in the README (under "Running a Task from the Web UI").

## Mocks

- Screenshot 2026-05-29 alle 21 45 41
- Screenshot 2026-05-29 alle 21 45 14

## Scope

- **In scope:** User toggle on the task show page when active runs exist; controller-level `?refresh=false` support on both index and show actions.
- **Out of scope:** Toggle UI on the task list page; persisting the preference beyond the current URL (e.g. cookies or local storage).

## Acceptance criteria

- [x] Task show page auto-refreshes by default when there are active runs (no change to current default behavior).
- [x] Users can disable auto-refresh via a visible control on the task show page (when active runs exist).
- [x] Disabling auto-refresh persists via the `?refresh=false` URL parameter.
- [x] Users can re-enable auto-refresh from the same page.
- [x] No toggle is shown when the task has no active runs.
- [x] Integration tests cover both refresh states, toggle links, and the no-active-runs case.
- [x] README documents the feature.

### DIFF
--- a/app/controllers/maintenance_tasks/tasks_controller.rb
+++ b/app/controllers/maintenance_tasks/tasks_controller.rb
@@ -28,7 +28,7 @@ module MaintenanceTasks
     private
 
     def set_refresh
-      @refresh = true
+      @refresh = params[:refresh] != "false"
     end
   end
 end
--- a/app/controllers/maintenance_tasks/tasks_controller.rb
+++ b/app/controllers/maintenance_tasks/tasks_controller.rb
@@ -7,7 +7,7 @@ module MaintenanceTasks
   #
   # @api private
   class TasksController < ApplicationController
-    before_action :set_refresh, only: [:index]
+    before_action :set_refresh, only: [:index, :show]
 
     # Renders the maintenance_tasks/tasks page, displaying
     # available tasks to users, grouped by category.
--- a/app/views/maintenance_tasks/tasks/show.html.erb
+++ b/app/views/maintenance_tasks/tasks/show.html.erb
@@ -42,7 +42,16 @@
 </details>
 <% end %>
 
-<%= tag.div(data: { refresh: @task.refresh? || "" }) do %>
+<% if @task.refresh? %>
+  <div class="is-flex is-justify-content-flex-end mb-2">
+    <% if @refresh %>
+      <%= link_to "Disable auto-refresh", task_path(@task, refresh: false), class: "button is-small" %>
+    <% else %>
+      <%= link_to "Enable auto-refresh", task_path(@task), class: "button is-small" %>
+    <% end %>
+  </div>
+<% end %>
+<%= tag.div(data: { refresh: (@task.refresh? && @refresh) || "" }) do %>
   <% if @task.active_runs.any? %>
     <hr>
 
--- a/test/integration/maintenance_tasks/tasks_controller_test.rb
+++ b/test/integration/maintenance_tasks/tasks_controller_test.rb
@@ -4,32 +4,44 @@ require "test_helper"
 
 module MaintenanceTasks
   class TasksControllerTest < ActionDispatch::IntegrationTest
-    test "task list auto-refreshes by default" do
-      get maintenance_tasks_path
+    # Maintenance::UpdatePostsTask has a paused (active) run, so its show page
+    # refreshes by default.
+    ACTIVE_TASK = "Maintenance::UpdatePostsTask"
+
+    test "task page auto-refreshes by default when there are active runs" do
+      get maintenance_tasks.task_path(ACTIVE_TASK)
 
       assert_response :success
       assert_select "[data-refresh=true]"
     end
 
-    test "task list does not auto-refresh when refresh=false is passed" do
-      get maintenance_tasks_path, params: { refresh: "false" }
+    test "task page does not auto-refresh when refresh=false is passed" do
+      get maintenance_tasks.task_path(ACTIVE_TASK), params: { refresh: "false" }
 
       assert_response :success
       assert_select "[data-refresh=true]", false
     end
 
-    test "task list shows a link to disable auto-refresh when refreshing" do
-      get maintenance_tasks_path
+    test "task page shows a link to disable auto-refresh when refreshing" do
+      get maintenance_tasks.task_path(ACTIVE_TASK)
 
       assert_response :success
       assert_select "a[href*='refresh=false']", text: "Disable auto-refresh"
     end
 
-    test "task list shows a link to enable auto-refresh when not refreshing" do
-      get maintenance_tasks_path, params: { refresh: "false" }
+    test "task page shows a link to enable auto-refresh when not refreshing" do
+      get maintenance_tasks.task_path(ACTIVE_TASK), params: { refresh: "false" }
 
       assert_response :success
-      assert_select "a[href='/maintenance_tasks/']", text: "Enable auto-refresh"
+      assert_select "a", text: "Enable auto-refresh"
+    end
+
+    test "task page shows no auto-refresh toggle when there are no active runs" do
+      get maintenance_tasks.task_path("Maintenance::ImportPostsTask")
+
+      assert_response :success
+      assert_select "a", text: "Disable auto-refresh", count: 0
+      assert_select "a", text: "Enable auto-refresh", count: 0
     end
   end
 end
--- a/test/integration/maintenance_tasks/tasks_controller_test.rb
+++ b/test/integration/maintenance_tasks/tasks_controller_test.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+module MaintenanceTasks
+  class TasksControllerTest < ActionDispatch::IntegrationTest
+    test "task list auto-refreshes by default" do
+      get maintenance_tasks_path
+
+      assert_response :success
+      assert_select "[data-refresh=true]"
+    end
+
+    test "task list does not auto-refresh when refresh=false is passed" do
+      get maintenance_tasks_path, params: { refresh: "false" }
+
+      assert_response :success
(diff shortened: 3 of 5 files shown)


app/views/maintenance_tasks/tasks/show.html.erb
```diff
@@ -42,7 +42,16 @@
 </details>
 <% end %>
 
-<%= tag.div(data: { refresh: @task.refresh? || "" }) do %>
+<% if @task.refresh? %>
```
adrianna-chang-shopify, line 45: Should we move this whole block into the refreshed wrapper (`<%= tag.div(data: { refresh: (@task.refresh? && @refresh) || "" }) do %>`)? Otherwise, when a task finishes, we'll still show the disable / enable button even though `@task.refresh? ` is `false`, because we _won't_ have refreshed this part of the page.

test/integration/maintenance_tasks/tasks_controller_test.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+module MaintenanceTasks
+  class TasksControllerTest < ActionDispatch::IntegrationTest
```
adrianna-chang-shopify, line 6: I'd prefer we follow the convention of the existing system tests and test the toggle behaviour in `test/system/maintenance_tasks/tasks_test.rb`.

app/controllers/maintenance_tasks/tasks_controller.rb
adrianna-chang-shopify, line 24: We should filter out `refresh` as a param here.

app/controllers/maintenance_tasks/tasks_controller.rb
```diff
@@ -28,7 +28,7 @@ def show
     private
 
     def set_refresh
-      @refresh = true
+      @refresh = params[:refresh] != "false"
```
adrianna-chang-shopify, line 31: I think we could name this something more intention revealing now.
```suggestion
      @auto_refresh_enabled = params[:refresh] != "false"
```

app/views/maintenance_tasks/tasks/show.html.erb
```diff
@@ -42,7 +42,16 @@
...
+  <div class="is-flex is-justify-content-flex-end mb-2">
+    <% if @refresh %>
+      <%= link_to "Disable auto-refresh", task_path(@task, refresh: false), class: "button is-small" %>
```
adrianna-chang-shopify, line 48:
```suggestion
      <%= link_to "Disable auto-refresh", task_path(@task, refresh: false), class: "button is-small is-light" %>
```

Nit, to differentiate it from the bg a bit more.

app/views/maintenance_tasks/tasks/show.html.erb
```diff
@@ -42,7 +42,16 @@
...
+    <% else %>
+      <%= link_to "Enable auto-refresh", task_path(@task), class: "button is-small" %>
```
adrianna-chang-shopify, line 50:
```suggestion
      <%= link_to "Enable auto-refresh", task_path(@task), class: "button is-small is-light" %>
```

adrianna-chang-shopify (commented):
Hey @knightq , thanks for contributing to the gem! This change makes sense to me, I left a few comments.

PR #1525, opened 2026-07-27
Use a minimum Ruby 3.3 for Dependabot

etiennebarrie:
Otherwise Dependabot can't upgrade rubocop-shopify or selenium-webdriver, both of which require Ruby 3.3.

Also commit changes to Gemfile.lock which happen by just running `bundle install`, the `BUNDLED WITH` section declaring Bundler 4.0.11 but the `CHECKSUMS` section using 4.0.13.

Then that broke because we test using Gemfile and Ruby 3.2, which was breaking because of the changed requirement in Gemfile. So moving to test Rails 8.1 with a separate Gemfile, which goes into the other branch and doesn't add the 3.3 Ruby requirement.

That means that CI will no longer have a single case testing precisely with the dependencies declared in Gemfile.lock, but that is mitigated by the weekly CI helping us catch incompatibilities early. We could also add a Ruby 4.0 / Gemfile / sqlite special case to the matrix, which will ensure at least one job passes in the situation where a gem update breaks something in the engine/tests. Not sure how helpful it is, but let me know if you'd want that.


that-jill (approved):

adrianna-chang-shopify (approved):

etiennebarrie:
Actually rewrote the last commit to only use the Rails 8.1-specific Gemfile for Ruby 3.2. That lets us still test 3.3, 3.4 and 4.0 with the exact versions from Gemfile.lock.

etiennebarrie:
That didn't fix the issue 😢 
image

PR #1527, opened 2026-07-29
Expose the Run id to the Task instance

lucasmazza:
A Task often wants to know which execution it belongs to. Exposing the whole Run on the Task has been turned down before, so this exposes just the id: `Task#run_id` gives a stable identifier for a single execution — useful for correlating log lines and reports, or as an idempotency key — without coupling the Task to the Run object.

`Task#initialize` now accepts an optional `run_id:` keyword that `Run#task` fills in. `Task.new` still works with no arguments so `Maintenance::SomeTask.collection` and friends keep working in tests, where `run_id` is simply nil.

This picks up #1112, which went stale, with the review feedback there applied.

### DIFF
--- a/README.md
+++ b/README.md
@@ -731,6 +731,29 @@ values will be available for subsequent calls to `process` within the same job.
 Still, memoization can be used for throttling or reporting, and you can use
 [Task callbacks](#using-task-callbacks) to persist or log a report for example.
 
+#### Identifying the Run
+
+A Task instance exposes the id of the Run it is being processed by as `run_id`,
+which gives you a stable identifier for the execution of the Task across
+interruptions and potential pauses:
+
+```ruby
+class Maintenance::UpdatePostsTask < MaintenanceTasks::Task
+  def collection
+    Post.all
+  end
+
+  def process(post)
+    post.update!(content: "Updated by run #{run_id}")
+  end
+end
+```
+
+It is useful to correlate log lines or reports with the Run they came from, or
+as an idempotency key for work performed while processing the collection. It is
+`nil` when a Task is instantiated outside of a typical workflow, such as when
+you call `Maintenance::UpdatePostsTask.new` in your tests.
+
 ### Writing tests for a Task
 
 The task generator will also create a test file for your task in the folder
--- a/README.md
+++ b/README.md
@@ -731,6 +731,28 @@ values will be available for subsequent calls to `process` within the same job.
 Still, memoization can be used for throttling or reporting, and you can use
 [Task callbacks](#using-task-callbacks) to persist or log a report for example.
 
+#### Identifying the Run
+
+A Task instance exposes the id of the Run it is being processed by as `run_id`,
+which gives you a stable identifier for a single execution of the Task:
+
+```ruby
+class Maintenance::UpdatePostsTask < MaintenanceTasks::Task
+  def collection
+    Post.all
+  end
+
+  def process(post)
+    post.update!(content: "Updated by run #{run_id}")
+  end
+end
+```
+
+It is useful to correlate log lines or reports with the Run they came from, or
+as an idempotency key for work performed while processing the collection. It is
+`nil` when the Task is instantiated outside of a Run, which is the case when you
+call `Maintenance::UpdatePostsTask.new` in your tests.
+
 ### Writing tests for a Task
 
 The task generator will also create a test file for your task in the folder
--- a/README.md
+++ b/README.md
@@ -734,7 +734,7 @@ Still, memoization can be used for throttling or reporting, and you can use
 #### Identifying the Run
 
 A Task instance exposes the id of the Run it is being processed by as `run_id`,
-which gives you a stable identifier for a single execution of the Task:
+which gives you a stable identifier for the execution of the Task across interruptions and potential pauses:
 
 ```ruby
 class Maintenance::UpdatePostsTask < MaintenanceTasks::Task
--- a/test/dummy/app/tasks/maintenance/run_id_task.rb
+++ b/test/dummy/app/tasks/maintenance/run_id_task.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+module Maintenance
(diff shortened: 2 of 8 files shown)


test/models/maintenance_tasks/run_test.rb
```diff
@@ -636,6 +636,16 @@ class RunTest < ActiveSupport::TestCase
       assert_equal({ "foo" => "bar" }, run.task.metadata)
     end
 
+    test "#task sets the run id on the Task" do
+      run = Run.create!(task_name: "Maintenance::TestTask")
+      assert_equal run.id, Run.find(run.id).task.run_id
```
etiennebarrie, line 641: Any reason not to do this instead?
```suggestion
      assert_equal run.id, run.task.run_id
```

README.md
```diff
@@ -731,6 +731,28 @@ values will be available for subsequent calls to `process` within the same job.
 Still, memoization can be used for throttling or reporting, and you can use
 [Task callbacks](#using-task-callbacks) to persist or log a report for example.
 
+#### Identifying the Run
+
+A Task instance exposes the id of the Run it is being processed by as `run_id`,
+which gives you a stable identifier for a single execution of the Task:
```
etiennebarrie: Maybe add a few words to disambiguate with pauses/resumes/interruptions. I feel like the `single` here makes it ambiguous.
```suggestion
which gives you a stable identifier for the execution of the Task across interruptions and potential pauses:
```

etiennebarrie (commented):
Yeah I think that can be useful.

test/models/maintenance_tasks/run_test.rb
lucasmazza: `run.task.run_id` here is `nil` because `run.task` is memoized before during validation, before the model is persisted.

etiennebarrie: Right, and that can't be an issue because none of the callbacks on Task run when we enqueue, so we can't see a `nil` `run_id`.

etiennebarrie (approved):
Looks good on my end, I'll ping the rest of the team to see if anyone else agrees!

README.md
```diff
@@ -731,6 +731,28 @@ values will be available for subsequent calls to `process` within the same job.
...
+  def process(post)
+    post.update!(content: "Updated by run #{run_id}")
+  end
+end
+```
+
+It is useful to correlate log lines or reports with the Run they came from, or
+as an idempotency key for work performed while processing the collection. It is
+`nil` when the Task is instantiated outside of a Run, which is the case when you
+call `Maintenance::UpdatePostsTask.new` in your tests.
```
adrianna-chang-shopify: Nit, I'd generalize the wording here a bit, since "outside of a Run" involves understanding how we've implemented Run logic, which the typical user doesn't need to understand.
```suggestion
`nil` when a Task is instantiated outside of a typical workflow, such as when you call `Maintenance::UpdatePostsTask.new` in your tests.
```

README.md
```diff
@@ -731,6 +731,28 @@ values will be available for subsequent calls to `process` within the same job.
 Still, memoization can be used for throttling or reporting, and you can use
 [Task callbacks](#using-task-callbacks) to persist or log a report for example.
 
+#### Identifying the Run
+
+A Task instance exposes the id of the Run it is being processed by as `run_id`,
+which gives you a stable identifier for the execution of the Task across interruptions and potential pauses:
```
adrianna-chang-shopify:
```suggestion
which gives you a stable identifier for the execution of the Task across
interruptions and potential pauses:
```

test/dummy/app/tasks/maintenance/run_id_task.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Maintenance
+  class RunIdTask < MaintenanceTasks::Task
```
adrianna-chang-shopify: I would personally remove this fixture, I think it's a bit overkill just to validate that we instantiate the task with the `run_id` and can access it from within the task instance. IMO the combination of the unit tests from run_test.rb and task_test.rb probide sufficient coverage that `Task.new(run_id:)` receives and stores the `run_id`, and that `Run#task` sets it when we are returning the task instance.

adrianna-chang-shopify (commented):
I have a few small comments, but exposing `run_id` seems reasonable to me.

README.md
```diff
@@ -731,6 +731,28 @@ values will be available for subsequent calls to `process` within the same job.
...
+
+  def process(post)
+    post.update!(content: "Updated by run #{run_id}")
+  end
+end
+```
+
+It is useful to correlate log lines or reports with the Run they came from, or
+as an idempotency key for work performed while processing the collection. It is
+`nil` when a Task is instantiated outside of a typical workflow, such as when you call `Maintenance::UpdatePostsTask.new` in your tests.
```
adrianna-chang-shopify: Nit, let's keept he line wrapping consistent
```suggestion
`nil` when a Task is instantiated outside of a typical workflow, such as when
you call `Maintenance::UpdatePostsTask.new` in your tests.
```

test/jobs/maintenance_tasks/task_job_test.rb
```diff
@@ -31,6 +31,17 @@ class TaskJobTest < ActiveJob::TestCase
       assert_no_enqueued_jobs
     end
 
+    test ".perform exposes the Run id to the Task while it is processing" do
```
adrianna-chang-shopify: Sorry, to clarify, I think we can remove the job test here and the task fixture entirely -- we haven't changed anything about the job's behaviour, we are ultimately re-testing that the task instance exposes the run.

lucasmazza: done at d3c94f3

adrianna-chang-shopify (approved):

adrianna-chang-shopify:
@lucasmazza can you squash your commits please? I'll merge after. Thanks!

lucasmazza:
@adrianna-chang-shopify squashed and pushed

PR #1533, opened 2026-08-15
Add task auto-refresh toggle

### LINKED ISSUE #1490: Add the ability to disable auto refresh

salasebas:
Closes #1490.

Continues #1489 with its review feedback applied. Adds a task-page toggle for disabling auto-refresh without passing `refresh` through as a task argument.

Original implementation by @knightq, preserved as co-author.

## Testing

- `bin/setup`
- `bundle exec rails test`
- `bundle exec rails test:system`
- `bundle exec rubocop`
- `bundle exec herb analyze`

### DIFF
--- a/app/views/maintenance_tasks/tasks/index.html.erb
+++ b/app/views/maintenance_tasks/tasks/index.html.erb
@@ -1,4 +1,4 @@
-<%= tag.div(data: { refresh: (defined?(@refresh) && @refresh) || "" }) do %>
+<%= tag.div(data: { refresh: (defined?(@auto_refresh_enabled) && @auto_refresh_enabled) || "" }) do %>
   <% if @available_tasks.empty? %>
     <div class="content is-large">
       <h3 class="title is-3"> The MaintenanceTasks gem has been successfully installed! </h3>
--- a/app/views/maintenance_tasks/tasks/show.html.erb
+++ b/app/views/maintenance_tasks/tasks/show.html.erb
@@ -42,7 +42,17 @@
 </details>
 <% end %>
 
-<%= tag.div(data: { refresh: @task.refresh? || "" }) do %>
+<%= tag.div(data: { refresh: (@task.refresh? && @auto_refresh_enabled) || "" }) do %>
+  <% if @task.refresh? %>
+    <div class="is-flex is-justify-content-flex-end mb-2">
+      <% if @auto_refresh_enabled %>
+        <%= link_to "Disable auto-refresh", task_path(@task, refresh: false), class: "button is-small is-light" %>
+      <% else %>
+        <%= link_to "Enable auto-refresh", task_path(@task), class: "button is-small is-light" %>
+      <% end %>
+    </div>
+  <% end %>
+
   <% if @task.active_runs.any? %>
     <hr>
 
--- a/README.md
+++ b/README.md
@@ -879,6 +879,11 @@ end
 
 You can run your new Task by accessing the Web UI and clicking on "Run".
 
+A Task's page auto-refreshes in the background while it has active runs, so the
+progress and run statuses stay up to date. You can turn this off with the
+"Disable auto-refresh" toggle on the Task page, which adds a `?refresh=false`
+parameter to the URL. Use "Enable auto-refresh" to turn it back on.
+
 #### Running a Task from the command line
 
 Alternatively, you can run your Task in the command line:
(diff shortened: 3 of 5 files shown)


salasebas:
I have signed the CLA!

knightq (approved):

app/views/maintenance_tasks/tasks/show.html.erb
```diff
@@ -42,7 +42,17 @@
 </details>
 <% end %>
 
-<%= tag.div(data: { refresh: @task.refresh? || "" }) do %>
+<%= tag.div(data: { refresh: (@task.refresh? && @auto_refresh_enabled) || "" }) do %>
```
gmcgibbon, line 45:
```suggestion
<%= tag.div(data: { refresh: (@task.refresh? && defined?(@auto_refresh_enabled) && @auto_refresh_enabled) || "" }) do %>
```

Might be helpful to define a view helper to shorthand `defined?(@auto_refresh_enabled) && @auto_refresh_enabled)` for the index view and for `<% if @auto_refresh_enabled %>` below.

gmcgibbon (approved):
LGTM once we add the missing `defined?` checks.

app/views/maintenance_tasks/tasks/index.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= tag.div(data: { refresh: (defined?(@refresh) && @refresh) || "" }) do %>
+<%= tag.div(data: { refresh: (defined?(@auto_refresh_enabled) && @auto_refresh_enabled) || "" }) do %>
```
gmcgibbon, line 1:
```suggestion
<%= tag.div(data: { refresh: (@available_tasks.any? && defined?(@auto_refresh_enabled) && @auto_refresh_enabled) || "" }) do %>
```
This too. No need to refresh the index if there's no tasks.

gmcgibbon:
On second thought, I don't think the `defined?` is adding value. I'll remove it and add my change myself. Thank you!

PR #1534, opened 2026-08-15
Fix cursor pagination ordering for task runs

salasebas:
## What changed

Paginate runs using the deterministic `(created_at, id)` order while preserving the existing ID cursor.

This prevents duplicated or omitted runs when IDs do not follow creation order or multiple runs share a timestamp.

## Release notes

Fixed task run history pagination for UUID primary keys, backfilled records, and identical timestamps. No database migration is required.


app/models/maintenance_tasks/runs_page.rb
```diff
@@ -61,5 +62,14 @@ def last?
       records
       @extra_run.nil?
     end
+
+    private
+
+    def cursor_condition(cursor_run)
+      run = @runs.klass.arel_table
```
gmcgibbon: This is private API in Active Record. Let's instead use this:
```rb
@runs.where(created_at: ...cursor_run.created_at)
    .or(@runs.where(created_at: cursor_run.created_at, id: ...cursor_run.id))
```

salasebas: Updated in 712722c: the cursor now uses range predicates and `Relation#or` instead of Arel nodes.

adrianna-chang-shopify (approved):
Can you squash your commits, please?

PR #1535, opened 2026-08-15
Mark interrupted Runs as errored when re-enqueuing fails

salasebas:
### Why

An interrupted Task is expected to be re-enqueued so another job can resume it. Active Job can reject an enqueue without raising when a `before_enqueue` callback aborts or a queue adapter reports an enqueue error.

The deferred re-enqueue currently ignores that result, leaving the Run as `interrupted` even though no job was enqueued to continue it.

### What changed

Check `successfully_enqueued?` after the re-enqueue attempt. When it fails, use the adapter's `enqueue_error` if available, or raise an `ActiveJob::EnqueueError` for a callback rejection.

The existing error handling then persists the Run as `errored`, where it can be resumed. This preserves the callback's ability to reject the enqueue while preventing the Run from being stranded without a job.

### DIFF
--- a/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
+++ b/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
@@ -30,6 +30,21 @@ module MaintenanceTasks
       end
     end
 
+    if ActiveJob.gem_version < Gem::Version.new("8.1")
+      # Active Job 7.2 and 8.0 retry by enqueuing a duplicate, so
+      # +successfully_enqueued?+ is set on the copy rather than this job.
+      # Enqueue self instead, keeping the +enqueue_retry+ instrumentation.
+      def retry_job(options = {})
+        return if defined?(@retried) && @retried
+
+        result = instrument(:enqueue_retry, options.slice(:error, :wait)) do
+          enqueue(options)
+        end
+        @retried = true
+        result
+      end
+    end
+
     private
 
     def serialized_cursor_position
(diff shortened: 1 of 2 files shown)


gmcgibbon:
Looks like some tests are failing. Can you please take a look?

app/jobs/concerns/maintenance_tasks/task_job_concern.rb
```diff
@@ -30,6 +30,15 @@ def retry_on(*, **)
       end
     end
 
+    # Rails 7.2/8.0 retry a dup, so enqueue self for successfully_enqueued?.
```
adrianna-chang-shopify: Can we apply this only to Rails <= 8.0 ?
```ruby
  if ActiveJob.gem_version < Gem::Version.new("8.1")
    def retry_job(options = {})
      return if defined?(@retried) && @retried
  ...
```

I think we may also need to preserve the `enqueue_retry` instrumentation that AJ uses?
```ruby
  instrument :enqueue_retry, options.slice(:error, :wait) do
    # enqueue operation
  end
```

salasebas: ✅

test/jobs/maintenance_tasks/task_job_test.rb
```diff
@@ -173,6 +173,38 @@ class TaskJobTest < ActiveJob::TestCase
       assert_enqueued_with(job: TaskJob) { TaskJob.perform_now(@run) }
     end
 
+    test ".perform_now instruments enqueue_retry when the job is interrupted" do
```
adrianna-chang-shopify: Let's wrap this test in a version check as well:
```ruby
  if ActiveJob.gem_version < Gem::Version.new("8.1")
    test "..." do
      # ...
    end
  end
```

Ensures we only test for the patched behaviour (and will make it easier for us to remember to remove the test once the min version reaches 8.1)

adrianna-chang-shopify (approved):
One small piece of feedback, I'll merge after! Thanks for the contribution.

test/jobs/maintenance_tasks/task_job_test.rb
salasebas: ✅

PR #1536, opened 2026-08-15
Skip task setup for non-runnable Runs

salasebas:
A job can remain queued after its Run is cancelled. When eventually worked, it currently downloads CSV data, builds the collection, sets `started_at`, and runs `after_start` before stopping.

Claim the Run before task setup and abort when it is no longer runnable. Preserve pausing and cancelling transitions, and prevent the no-locking fallback from moving paused or completed Runs back to running.

Adds regression coverage for cancelled jobs, lifecycle callbacks, races, and both locking paths.

Tests: `bundle exec rake test` and `bundle exec rubocop --no-parallel`.

### DIFF
--- a/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
+++ b/app/jobs/concerns/maintenance_tasks/task_job_concern.rb
@@ -139,18 +139,32 @@ module MaintenanceTasks
 
     def before_perform
       @run = arguments.first
-      @task = @run.task
-      if @task.has_csv_content?
-        @task.csv_content = @run.csv_file.download
+      @run.running
+
+      if @run.running?
+        @task = @run.task
+        if @task.has_csv_content?
+          @task.csv_content = @run.csv_file.download
+        end
+        @run.reload_status
       end
 
-      @run.running
+      abort_unless_run_is_running
+      @last_status_reload = Time.now
 
       @ticker = Ticker.new(MaintenanceTasks.ticker_delay) do |ticks, duration|
         @run.persist_progress(ticks, duration)
       end
+    end
 
-      @last_status_reload = nil
+    def abort_unless_run_is_running
+      return if @run.running?
+
+      if @run.cancelling? || @run.pausing?
+        @run.job_shutdown
+        @run.persist_transition
+      end
+      throw(:abort)
     end
 
     def on_start
--- a/app/models/concerns/maintenance_tasks/run_concern.rb
+++ b/app/models/concerns/maintenance_tasks/run_concern.rb
@@ -29,6 +29,8 @@ module MaintenanceTasks
       :interrupted,
     ].freeze
 
+    RUNNABLE_STATUSES = [:enqueued, :running, :interrupted].freeze
+
     STOPPING_STATUSES = [
       :pausing,
       :cancelling,
@@ -233,6 +235,14 @@ module MaintenanceTasks
       ACTIVE_STATUSES.include?(status.to_sym)
     end
 
+    # Returns whether the Run is runnable, which is defined as
+    # having a status of enqueued, running, or interrupted.
+    #
+    # @return [Boolean] whether the Run is runnable.
+    def runnable?
+      RUNNABLE_STATUSES.include?(status.to_sym)
+    end
+
     # Returns the duration left for the Run to finish based on the number of
     # ticks left and the average time needed to process a tick. Returns nil if
     # the Run is completed, or if tick_count or tick_total is zero.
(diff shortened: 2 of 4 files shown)


app/models/concerns/maintenance_tasks/run_concern.rb
```diff
@@ -250,20 +252,21 @@ def time_to_completion
...
-    # If the run is stopping already, it will not transition to running.
+    # If the Run is not runnable, it will not transition to running.
     # Rescues and retries status transition if an ActiveRecord::StaleObjectError
     # is encountered.
     def running
       if locking_enabled?
         with_stale_object_retry do
-          running! unless stopping?
+          reload_status if running?
+          running! if RUNNABLE_STATUSES.include?(status.to_sym)
```
gmcgibbon:
```suggestion
          running! if runnable?
```
Let's define a predicate for this. It is easier to read.

app/jobs/concerns/maintenance_tasks/task_job_concern.rb
```diff
@@ -139,18 +139,30 @@ def task_iteration(input)
 
     def before_perform
       @run = arguments.first
+      @run.running
+      abort_unless_run_is_running
```
gmcgibbon: Do we need to call this multiple times? Perhaps we just need the second one.

app/models/concerns/maintenance_tasks/run_concern.rb
salasebas: Added Run#runnable?

app/jobs/concerns/maintenance_tasks/task_job_concern.rb
salasebas: fixed

app/jobs/concerns/maintenance_tasks/task_job_concern.rb
```diff
@@ -139,18 +139,32 @@ def task_iteration(input)
...
+      if @run.running?
+        @task = @run.task
+        if @task.has_csv_content?
+          @task.csv_content = @run.csv_file.download
+        end
+        @run.reload_status
       end
 
-      @run.running
+      abort_unless_run_is_running
```
adrianna-chang-shopify: Nit, could we move the check outside of the method and call it `abort_perform` instead? e.g.

```ruby
  abort_perform unless @run.running?

# ...

  def abort_perform
    if @run.cancelling? || @run.pausing?
      @run.job_shutdown
      @run.persist_transition
    end

    throw(:abort)
  end
```

test/jobs/maintenance_tasks/task_job_test.rb
```diff
@@ -33,16 +33,64 @@ class TaskJobTest < ActiveJob::TestCase
...
       travel MaintenanceTasks.stuck_task_duration
-      @run.cancel # force cancel the Run
-      assert_predicate @run, :cancelled?
-      Maintenance::TestTask.any_instance.expects(:process).never
+      run.cancel # force cancel the Run
+      assert_predicate run, :cancelled?
+      Maintenance::CallbackTestTask.any_instance.expects(:collection).never
+      Maintenance::CallbackTestTask.any_instance.expects(:process).never
+      Maintenance::CallbackTestTask.any_instance
+        .expects(:after_start_callback).never
```
adrianna-chang-shopify, line 50: Nit, can we break up this test a bit?
```suggestion
      run = Run.create!(task_name: "Maintenance::CallbackTestTask")
      TaskJob.perform_later(run)
      
      Maintenance::CallbackTestTask.any_instance
        .expects(:after_cancel_callback).once
      run.cancel
      travel MaintenanceTasks.stuck_task_duration
      run.cancel # force cancel the Run
      
      assert_predicate run, :cancelled?
      
      Maintenance::CallbackTestTask.any_instance.expects(:collection).never
      Maintenance::CallbackTestTask.any_instance.expects(:process).never
      Maintenance::CallbackTestTask.any_instance
        .expects(:after_start_callback).never
```

test/jobs/maintenance_tasks/task_job_test.rb
```diff
@@ -33,16 +33,64 @@ class TaskJobTest < ActiveJob::TestCase
...
+      Run.stubs(:locking_enabled?).returns(false)
+      run = Run.new(task_name: "Maintenance::ImportPostsTask")
+      run.csv_file.attach(
+        io: file_fixture("sample.csv").open,
+        filename: "sample.csv",
+      )
+      run.save!
+      csv_content = file_fixture("sample.csv").binread
+      download_key = run.csv_file.blob.key
+      ActiveStorage::Blob.service.expects(:download).with do |key|
```
adrianna-chang-shopify, line 87: Nit, would also just separate the operational part from the expectations here:
```suggestion
      run.csv_file.attach(
        io: file_fixture("sample.csv").open,
        filename: "sample.csv",
      )
      run.save!
      
      csv_content = file_fixture("sample.csv").binread
      download_key = run.csv_file.blob.key
      ActiveStorage::Blob.service.expects(:download).with do |key|
```

adrianna-chang-shopify (approved):
Thanks for working on this! A few nits, I'll merge once addressed.

CI should be fixed after a rebase as well :)

test/jobs/maintenance_tasks/task_job_test.rb
salasebas: fixed

test/jobs/maintenance_tasks/task_job_test.rb
salasebas: fixed

app/jobs/concerns/maintenance_tasks/task_job_concern.rb
salasebas: fixed

PR #1542, opened 2026-08-21
Refresh feature refactor

gmcgibbon:
- Remove redundant `defined?` check
- Don't refresh the index page when it is empty. It makes copy pasting the task generating command annoying.
- Add a refresh disabling button to the task index


gmcgibbon:
The CI issue will be fixed by https://github.com/rails/rails/pull/58519

PR #1544, opened 2026-08-31
Use and enforce strict locals

etiennebarrie:
This makes sure all the engine templates use strict locals.

Strict-local templates are better because they allow only compiling the template once (see https://guides.rubyonrails.org/action_view_overview.html#strict-locals) because template compilation needs a `render` call to know how to define the compiled templates for non-strict-local templates. They also theoretically allow eager-loading of templates at boot time (as it stands templates are only eager-loaded in `Rails::Application#ractorize!` but that could/should change). They also have better error messages and better optional values (no need for `defined?`).

I didn't go with Herb linters because they don't really match the reality of the feature in Action View, ignoring non-partials (see https://github.com/marcoroth/herb/issues/2501).


nvasilevski (approved):

test/strict_locals_test.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class StrictLocalsTest < ActiveSupport::TestCase
+  test "all engine templates declare strict locals" do
```
nvasilevski, line 6: Any opinions on this being a linter? Since we may want to enforce this across multiple repos it would be much more awesome to be able to just set `EnforceViewStrictLocals: true`

etiennebarrie: I looked into it but herb linters:
1. are in JS and would require to add a JS runtime here
2. think non-partial templates don't need strict locals (see the issue linked to above)

PR #1549, opened 2026-09-14
Pin json gem to < 3 for released Rails versions

### LINKED ISSUE #1546: Weekly CI run failed
https://github.com/Shopify/maintenance_tasks/actions/runs/34790586181

adrianna-chang-shopify:
Compat with json 3 was only added to Rails recently: https://github.com/rails/rails/pull/58601

Pin the version for everything except Rails main.

Closes: https://github.com/Shopify/maintenance_tasks/issues/1546


etiennebarrie (approved):

Gemfile
```diff
@@ -6,6 +6,9 @@ gemspec
 
 gem "debug"
 gem "herb"
+# Released Rails versions are not compatible with JSON 3 yet.
+# https://github.com/rails/rails/pull/58601
+gem "json", @json_gem_requirement || "< 3"
```
etiennebarrie, line 13: Not sure how Dependabot is going to react to that and if it will try to upgrade JSON anyway, but Rails 8.1 will soon have a release that support json 3, so we'll need another tweak to this to handle that once it's out.

adrianna-chang-shopify: AFAICT the Dependabot rewriter parses `||` as an `or` node and skips it: https://github.com/dependabot/dependabot-core/blob/main/bundler/lib/dependabot/bundler/file_updater/requirement_replacer.rb#L120

I added an extra note to the Gemfile about upgrading to a release with proper JSON 3 support as soon as we can.

